PR #90, opened 2020-05-13
Requirements for listing action in the actions marketplace

zkoppert:


action.yml
```diff
@@ -0,0 +1,9 @@
+name: super-linter
+author: github
```
admiralAwkbar, line 2: Can this be GitHub, or does it need to be lower case?

admiralAwkbar:
@zkoppert Can we make that update?

PR #93, opened 2020-05-14
even better

admiralAwkbar:
- Better local commands


admiralAwkbar:
- Needed to make the local commands more readable and the table seems to really help

PR #94, opened 2020-05-14
Switch to versioned docker container example

zkoppert:
Switch to versioned pointer from pointing people to the latest so they don't get unexpected or unstable behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ jobs:
       # Run Linter against code base #
       ################################
       - name: Lint Code Base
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@zkoppert I agree we need to cut some releases and make this more stable for long term usage and more predictable upgrade patterns... count me in...

zkoppert:
@admiralAwkbar I cut a release v1.0.0 and tested the example workflow file with the changes made in this PR on a personal repo and its working well. I'll update the release instructions in the `README.md` and after that I think we are good to go.

PR #100, opened 2020-06-12
Restructure linter flags

### LINKED ISSUE #98: Restructure linter flags
As discussed in #7, we'd like to be able to run each individual linter in a separate process.
To do this, we need an easy way to enable only one specific linter per process.

Rather than defaulting to linting all supported languages, and taking an environment variable to disable a specific language. I think we should:

Default to linting all languages (A reasonable default for the end user to expect, if they do not mess with any environment variables), but use variables to *enable* specific languages.
Meaning if no variables are present, lint all. If one variable is present, lint only that language, two variables is two languages, etc.

We could leave the variables to work the way they do now. However this means that if support is added for new languages, the variables passed to any one instance of the application, would have to be modified to disable the new language.

**Additional context**
Please read #7.

Hanse00:
As discussed in #98, this PR implements the logic required to enable all linters by default, but only enable the ones explicitly enabled, if any are explicitly enabled using environment variables.

The code can be tested by running:
`docker build . -t admiralawkbar/super-linter:latest` to build a new version of the container.

Then run:
`docker run -e RUN_LOCAL=true -e ACTIONS_RUNNER_DEBUG=true admiralawkbar/super-linter`, with some combination of the `VALIDATE_[LANGUAGE]` variables set to `true`, `false`, or left out.

What you should see is:
If none of the variables are added, all languages are set to true. Once either of the variables are added, that variable takes whichever value given, any variables not set are left to false.

Fixes #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,14 @@ jobs:
 ```
 
 ## Environment variables
-The super-linter allows you to pass the following `ENV` variables to be able to trigger different functionality:
+The super-linter allows you to pass the following `ENV` variables to be able to trigger different functionality.
+
+*Note:* All the `VALIDATE_[LANGAUGE]` variables behave in a specific way.  
+If none of them are passed, then they all default to true.  
+However if any one of the variables are set, we default to leaving any unset variable to false.  
+This means that if you run the linter "out of the box", all langauges will be checked.
+But if you wish to select specific linters, we give you full control to choose which linters are run,
+and won't run anything unexpected.
 
 | **ENV VAR** | **Default Value** | **Notes** |
 | --- | --- | --- |
@@ -120,10 +127,10 @@ The super-linter allows you to pass the following `ENV` variables to be able to 
 | **VALIDATE_JAVASCRIPT_STANDARD** | `true` | Flag to enable or disable the linting process of the language. (Utilizing: standard) |
 | **VALIDATE_TYPESCRIPT_ES** | `true` | Flag to enable or disable the linting process of the language. (Utilizing: eslint) |
 | **VALIDATE_TYPESCRIPT_STANDARD** | `true` | Flag to enable or disable the linting process of the language. (Utilizing: standard) |
-| **ANSIBLE_DIRECTORY** | `/ansible` | Flag to set the root directory for Ansible file location(s). |
 | **VALIDATE_DOCKER** | `true` | Flag to enable or disable the linting process of the language. |
 | **VALIDATE_GO** | `true` | Flag to enable or disable the linting process of the language. |
 | **VALIDATE_TERRAFORM** | `true` | Flag to enable or disable the linting process of the language. |
+| **ANSIBLE_DIRECTORY** | `/ansible` | Flag to set the root directory for Ansible file location(s). |
 | **ACTIONS_RUNNER_DEBUG** | `false` | Flag to enable additional information about the linter, versions, and additional output. |
 
 ### Template rules files
--- a/docs/run-linter-locally.md
+++ b/docs/run-linter-locally.md
@@ -23,29 +23,7 @@ Once the container has been downloaded to your local environment, you can then b
   - **NOTE:** The flag:`RUN_LOCAL` will set: `VALIDATE_ALL_CODEBASE` to true. This means it will scan **all** the files in the directory you have mapped. If you want to only validate a subset of your codebase, map a folder with only the files you wish to have linted
 
 ### Flags for running Locally
-You can add as many **Additional** flags as needed:
-
-| **ENV VAR** | **Command** | **Default Value** | **Notes** |
-| --- | --- | --- | --- |
-| **VALIDATE_YAML** | `-e VALIDATE_YAML=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_JSON** | `-e VALIDATE_JSON=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_XML** | `-e VALIDATE_XML=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_MD** | `-e VALIDATE_MD=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_BASH** | `-e VALIDATE_BASH=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_PERL** | `-e VALIDATE_PERL=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_PYTHON** | `-e VALIDATE_PYTHON=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_RUBY** | `-e VALIDATE_RUBY=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_COFFEE** | `-e VALIDATE_COFFEE=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_ANSIBLE** | `-e VALIDATE_ANSIBLE=<true\|false>` | `true` | Flag to enable or disable the linting process of the language |
-| **VALIDATE_JAVASCRIPT_ES** | `-e VALIDATE_JAVASCRIPT_ES=<true\|false>` | `true` | Flag to enable or disable the linting process of the language (Utilizing: eslint) |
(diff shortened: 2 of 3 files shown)


Hanse00:
A couple of examples of the code working:
```
philip@MacBook-Pro super-linter % docker run -e RUN_LOCAL=true -e ACTIONS_RUNNER_DEBUG=true admiralawkbar/super-linter
Gathering user validation information...
- Validating [YAML] files in code base...
- Validating [JSON] files in code base...
- Validating [XML] files in code base...
- Validating [MARKDOWN] files in code base...
- Validating [BASH] files in code base...
- Validating [PERL] files in code base...
- Validating [PYTHON] files in code base...
- Validating [RUBY] files in code base...
- Validating [COFFEE] files in code base...
- Validating [ANSIBLE] files in code base...
- Validating [JAVASCRIPT(eslint)] files in code base...
- Validating [JAVASCRIPT(standard)] files in code base...
- Validating [TYPESCRIPT(eslint)] files in code base...
- Validating [TYPESCRIPT(standard)] files in code base...
- Validating [DOCKER] files in code base...
- Validating [GOLANG] files in code base...
- Validating [TERRAFORM] files in code base...
```

```
philip@MacBook-Pro super-linter % docker run -e RUN_LOCAL=true -e ACTIONS_RUNNER_DEBUG=true -e VALIDATE_MD=true -e VALIDATE_PYTHON=true admiralawkbar/super-linter
Gathering user validation information...
- Excluding [YAML] files in code base...
- Excluding [JSON] files in code base...
- Excluding [XML] files in code base...
- Validating [MARKDOWN] files in code base...
- Excluding [BASH] files in code base...
- Excluding [PERL] files in code base...
- Validating [PYTHON] files in code base...
- Excluding [RUBY] files in code base...
- Excluding [COFFEE] files in code base...
- Excluding [ANSIBLE] files in code base...
- Excluding [JAVASCRIPT(eslint)] files in code base...
- Excluding [JAVASCRIPT(standard)] files in code base...
- Excluding [TYPESCRIPT(eslint)] files in code base...
- Excluding [TYPESCRIPT(standard)] files in code base...
- Excluding [DOCKER] files in code base...
- Excluding [GOLANG] files in code base...
- Excluding [TERRAFORM] files in code base...

**Comment by Hanse00** [comment]
@admiralAwkbar Could you please take a look when you have time? Cheers.

File: README.md [hunk 1]
```diff
@@ -101,7 +101,14 @@ jobs:
 ```
 
 ## Environment variables
-The super-linter allows you to pass the following `ENV` variables to be able to trigger different functionality:
+The super-linter allows you to pass the following `ENV` variables to be able to trigger different functionality.
+
+*Note:* All the `VALIDATE_[LANGAUGE]` variables behave in a specific way.
```
admiralAwkbar, line 106: Very nice cleanup of the language

docs/run-linter-locally.md
```diff
@@ -23,29 +23,7 @@ Once the container has been downloaded to your local environment, you can then b
   - **NOTE:** The flag:`RUN_LOCAL` will set: `VALIDATE_ALL_CODEBASE` to true. This means it will scan **all** the files in the directory you have mapped. If you want to only validate a subset of your codebase, map a folder with only the files you wish to have linted
 
 ### Flags for running Locally
-You can add as many **Additional** flags as needed:
-
-| **ENV VAR** | **Command** | **Default Value** | **Notes** |
```
admiralAwkbar, line 28: I have been wanting to remove this as it was a duplicate and wasn't sure if it should be here

lib/linter.sh
```diff
@@ -706,275 +706,375 @@ GetValidationInfo()
   ######################
   PRINT_ARRAY=()
 
-  ###############################
-  # Convert string to lowercase #
-  ###############################
+  ################################
+  # Convert strings to lowercase #
```
admiralAwkbar, line 710: Much cleaner to do all at once, instead of before each loop. good fix

admiralAwkbar:
@Hanse00 Very cool work, my only edge case:
- We have users who have multiple languages in a repo, let's say `go` and `ansible`, and they want to be able to lint `go`, but **not** `ansible`

What would they need to set their flags too, for this to work in that fashion?

docs/run-linter-locally.md
Hanse00: I personally felt that a description of all flags was relevant in the main "how to use it" part of the documentation. The fact that you are running it locally doesn't change that.

Hanse00:
> @Hanse00 Very cool work, my only edge case:
> 
> * We have users who have multiple languages in a repo, let's say `go` and `ansible`, and they want to be able to lint `go`, but **not** `ansible`
> 
> What would they need to set their flags too, for this to work in that fashion?
> 
> If they had other languages in the repo would they be linted, or excluded?
> 
> * Let's say they had just some `xml`, `yml`, etc... would those be excluded, or would they need to explicitly say to lint them?
> 
> The goal is for as many languages to be linted as possible, with as few excludes or includes. Just trying to understand the logic a little better :)

Yeah so the way I've implemented it here, the logic is:

If none of the VALIDATE_X flags are set: Lint everything. (We assume that's what people want by default)

If *any* of those have a value, we assume the end user wants to be in control of which linters to run, to we fall back to defaulting to none of them, and do whatever the flags say.

So let's say you have go and ansible, and only want to lint go: You add `VALIDATE_GO=true`, that's it.

You do of course have a point that in the case someone wants all but one linter running, the changes here are more cumbersome. Instead of `VALIDATE_ANSIBLE=false`, you'd have to provide `VALIDATE_GO=true VALIDATE_XML=true VALIDATE_YAML=true` and so forth.

This is more work when configuring the linter, *but* it provides the advantage that nothing here is implicit: By looking at the linter flags set, someone could determine without question which linters are being run. Whereas if the flags are "negating" (eg. `VALIDATE_ANSIBLE=false`), one still has to go look up which linters the tool supports, to know which ones are going to run (As in "What does all but Ansible mean? Which are *all*?").

The second potential advantage here is that you can add support for new linters down the road, without impacting existing users. If you were to add `VALIDATE_HTML` for example, anyone who has a set of linters to run, won't automatically run the new linter without noticing. Because they would have to *opt in* to using it by adding `VALIDATE_HTML=true`. Whereas in the old model, when the update is pushed out, users would have to remember to now add `VALIDATE_HTML=false` to continue the same behavior they had previously.

admiralAwkbar:
@zkoppert thoughts? I feel like this is a good move and will only require a few users to update some configs to move to a better format...

admiralAwkbar:
also, I noticed the Ci doesn't work for Forks... forgot that its a small security issue still and being worked towards. Will try to build locally and validate as well

Hanse00:
> @zkoppert thoughts? I feel like this is a good move and will only require a few users to update some configs to move to a better format...

Speaking of migration, there's a way people can change the config now, that will work both with the current implementation, and with what I am proposing.

Eg. Someone has Go and Ansible, only want to lint Go:

1. Currently they would have the flag `VALIDATE_ANSIBLE=false` set, to disable Ansible.
2. If they change their flags to `VALIDATE_ANSIBLE=false VALIDATE_GO=true` nothing will change, the current implementation would still run just Go.
3. The new implementation would similarly be able to read these flags, and continue to simply validate Go.
4. Once the migration is done, they could either remove the Ansible flag, leaving simply `VALIDATE_GO=true`, or leave them both as is.

zkoppert:
I agree that this is a good move towards an additive approach. Since it is a breaking change we will need to rev the version major number on the release. We should also have some information on the wiki dedicated to the config change between v1.x and v2.x .

admiralAwkbar:
Started wiki info at:
https://github.com/github/super-linter/wiki/Versions

Hanse00:
> Started wiki info at:
> https://github.com/github/super-linter/wiki/Versions
> 
> @zkoppert @Hanse00 I would love some input on the wiki doc

I think having some prose describe the versions, as you have currently, is great. But I also think it would be good to complement it with more of a "change log" bullet style of differences, for quick skimming.

zkoppert:
@admiralAwkbar @Hanse00 check this out and suggest any improvements. 

### Upgrading from v1.x to v2.x
- If no env configs option were being used in v1.x, no action needed
- If env configs options were being used in v1.x, take action below
  - For all languages in your codebase, you should explicitly set a true or false value for each
  - EX: codebase with Ansible and Go code: `VALIDATE_ANSIBLE=false`  ➡ `VALIDATE_ANSIBLE=false VALIDATE_GO=true`

admiralAwkbar:
@zkoppert

zkoppert:
Let's merge this into a different branch besides master so we can ensure checks pass. I will retarget for **staging**

PR #101, opened 2020-06-16
Run checks on env restructure work by @Hanse00

zkoppert:
cc/ @Hanse00 @admiralAwkbar


zkoppert:
check failed based on info below.

```
In lib/linter.sh line 103:
DEFAULT_VALIDATE_LANGUAGE='true'                      # Default to validate language
^-----------------------^ SC2034: DEFAULT_VALIDATE_LANGUAGE appears unused. Verify use (or export if used externally).
```

lib/linter.sh
```diff
@@ -706,275 +706,375 @@ GetValidationInfo()
...
-  # Convert string to lowercase #
-  ###############################
-  VALIDATE_JSON=$(echo "$VALIDATE_JSON" | awk '{print tolower($0)}')
   ######################################
-  # Validate we should check all files #
+  # Validate if we should check COFFEE #
   ######################################
-  if [[ "$VALIDATE_JSON" != "false" ]]; then
-    # Set to true
-    VALIDATE_JSON="$DEFAULT_VALIDATE_LANGUAGE"
```
zkoppert, line 734: looks like we forgot to remove the initialization of this variable.

Hanse00: My bad! The change did indeed involve a bit of copy and pasting, due to bash not making it trivial to loop over a list of variables. So I might have missed changing a variable name somewhere. Would definitely ask you guys give it a proper read over, to make sure I didn't miss something.

Hanse00:
> check failed based on info below.
> 
> ```
> In lib/linter.sh line 103:
> DEFAULT_VALIDATE_LANGUAGE='true'                      # Default to validate language
> ^-----------------------^ SC2034: DEFAULT_VALIDATE_LANGUAGE appears unused. Verify use (or export if used externally).
> ```

Good catch!

zkoppert:
now working on the following rubocop error. https://github.com/github/super-linter/runs/778435297#step:5:231

zkoppert:
Logging off for the day. no progress made on rubocop errors ☝🏻 if some else has a second to look at those. Something about safe mode being uninitialized.

admiralAwkbar:
@zkoppert @Hanse00 so the error is that for some reason, rubocop is installing version `0.85.1` as well as `0.74`... and then defaults to the wrong version...
so I'm trying to figure it out and get a fix to prevent that from happening again

admiralAwkbar:
So it looks like `0.74` needs a lib thats a default in `0.85.1` which in turns, causes `0.85.1` to be the default as its newer.... So the trick:
- Install `0.74.0` which also installs `0.85.1` and sets it as default
- Run step to reset `0.74.0` as default, and force remove `0.85.1`
- Reinstall `0.74.0` to fix linkage in libs
- walk away like nothing happened...

admiralAwkbar:
@zkoppert @Hanse00 we have lift off... lol

admiralAwkbar:
@zkoppert you want to cut the releases and call this gold?

zkoppert:
> So it looks like `0.74` needs a lib thats a default in `0.85.1` which in turns, causes `0.85.1` to be the default as its newer.... So the trick:
> 
> * Install `0.74.0` which also installs `0.85.1` and sets it as default
> * Run step to reset `0.74.0` as default, and force remove `0.85.1`
> * Reinstall `0.74.0` to fix linkage in libs
> * walk away like nothing happened...

🤯

Hanse00:
> So it looks like `0.74` needs a lib thats a default in `0.85.1` which in turns, causes `0.85.1` to be the default as its newer.... So the trick:
> 
> * Install `0.74.0` which also installs `0.85.1` and sets it as default
> * Run step to reset `0.74.0` as default, and force remove `0.85.1`
> * Reinstall `0.74.0` to fix linkage in libs
> * walk away like nothing happened...

Oof.

Perhaps long term it would be wise to update the linter used, to avoid all this?
But if it works for now, I guess that's cool.

PR #113, opened 2020-06-18
Use public URL for the Docker container

eunomie:
The previous one in the README is the URL to manage the repository (when you are the owner).


admiralAwkbar:
@eunomie thank you very much for the catch on this!

admiralAwkbar:
As only text fix, we don't need to pull this branch in to formally run automation

PR #114, opened 2020-06-18
README example: `actions/checkout@master` -> `actions/checkout@v2`

gr2m:


zkoppert:
> Should we update the workflows in this readme as well?

Yes, that would be good.

joshmgross:
> > Should we update the workflows in this readme as well?
> 
> It looks like this PR is updating the workflow in the readme. Is there another one somewhere you are referring to @joshmgross?

Apologies, I meant the workflows in this repository.
https://github.com/github/super-linter/blob/master/.github/workflows/deploy-PROD.yml#L38
https://github.com/github/super-linter/blob/master/.github/workflows/stack-linter.yml#L38
https://github.com/github/super-linter/blob/master/.github/workflows/deploy-DEV.yml#L41

admiralAwkbar:
@joshmgross I would update them as well
I think there was no reason we were not using v2...

zkoppert:
agreed i will pull this and those changes into a staging branch

PR #126, opened 2020-06-18
Staging

zkoppert:
Set checkout to version 2 as the current master could introduce breaking changes.


zkoppert:
Thanks to @joshmgross and @gr2m

PR #130, opened 2020-06-18
Fix typo (combination on → combination of)

waldyrious:
What the title says :)


admiralAwkbar:
@waldyrious thank you for the fix!

admiralAwkbar:
As this is text only, no need to pull in for automation

PR #131, opened 2020-06-18
fixing automation

admiralAwkbar:
This PR will cover:
- Not failing the cleanup step when merging a PR that was a fork
- Since the fork is not in the Org/Repo, we cant pull secrets, which breaks the step, but also, we never formally build the test container, so there is no cleanup to complete...


admiralAwkbar:
@zkoppert quick fix from seeing the failed jobs that were from merging the forks :)

PR #140, opened 2020-06-18
Allow disabling of errors

### LINKED ISSUE #138: Add Support for disabling errors (info-only runs)

I have a repository of legacy markdown files that super-linter would be really useful for. I've also encountered similar situations with other repositories in the past. Sometimes, I'm not looking for a pass/fail, but rather for counts of issues so that I can monitor a trend. 

I'd love to see an option that doesn't fail the build when a linter doesn't pass -- that surfaces the information but returns an exit code of `0`. This way, I can see the counts for informational purposes, and also verify as I'm getting things into shape that the counts really are dropping. Then, I can turn on the linter for "warnings as errors" essentially. 

* Using a different, specific linter that has this option. 
* Attempting to capture the exit code myself
* Not using a linter in my build until I've fixed it up locally.

SeanKilleen:
Resolves #138. 

* [x] Add environment variable to container
* [x] Update bash script to change exit code


lib/linter.sh
```diff
@@ -1094,6 +1096,30 @@ GetValidationInfo()
...
+  ###############################
+  # Convert string to lowercase #
+  ###############################
+  DSIABLE_ERRORS=$(echo "$DISABLE_ERRORS" | awk '{print tolower($0)}')
+
+  ############################
+  # Set to false if not true #
+  ############################
+  if [ "$DISABLE_ERRORS" != "true" ]; then
+    ACTIONS_RUNNER_DEBUG="false"
```
SeanKilleen, line 1119:
```suggestion
    DISABLE_ERRORS="false"
```

lib/linter.sh
```diff
@@ -1869,6 +1895,14 @@ Footer()
     fi
   done
 
+  ##################################
+  # Exit with 0 if errors disabled #
+  ##################################
+  
+  if ${ERRORS_DISABLED} then
```
SeanKilleen, line 1902:
```suggestion
  if ${DISABLE_ERRORS} then
```

SeanKilleen:
OK, so at this point I think the draft is in decent shape.

However, my bash is super rusty and this might need a little work so I'm holding off until #138 is seen as good to go.

lib/linter.sh
```diff
@@ -108,6 +109,7 @@ DEFAULT_ACTIONS_RUNNER_DEBUG='false'                  # Default value for debugg
 RAW_FILE_ARRAY=()                                     # Array of all files that were changed
 READ_ONLY_CHANGE_FLAG=0                               # Flag set to 1 if files changed are not txt or md
 TEST_CASE_FOLDER='.automation/test'                   # Folder for test cases we should always ignore
+DEFULAT_DISABLE_ERRORS='false'                        # Default to enabling errors
```
SeanKilleen, line 112:
```suggestion
DEFAULT_DISABLE_ERRORS='false'                        # Default to enabling errors
```

lib/linter.sh
```diff
@@ -1094,6 +1096,30 @@ GetValidationInfo()
...
+    ##################################
+    # No flag passed, set to default #
+    ##################################
+    DISABLE_ERRORS="$DEFAULT_DISABLE_ERRORS"
+  fi
+
+  ###############################
+  # Convert string to lowercase #
+  ###############################
+  DSIABLE_ERRORS=$(echo "$DISABLE_ERRORS" | awk '{print tolower($0)}')
```
SeanKilleen, line 1113:
```suggestion
  DISABLE_ERRORS=$(echo "$DISABLE_ERRORS" | awk '{print tolower($0)}')
```

admiralAwkbar:
This is being merged into #191 for CI

PR #141, opened 2020-06-18
Apply valid syntax to CODEOWNERS file

antleblanc:
According to the following resource there's no comma separator in the CODEOWNERS file.

See: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax


zkoppert:
That explains why the branch protection isn't working!! haha. Thanks!!

PR #159, opened 2020-06-18
Update to latest golangci-lint

iggy:
In setting super-linter up on one of my repo's, I ran into
golangci/golangci-lint#827 which is apparently fixed in v1.23.8 (and has
improved error messages in v1.24.0). I'm not sure if there's a reason for
using the version that is there now, but if the latest doesn't work,
please at least update to 1.24.0.

Signed-off-by: iggy <iggy@theiggy.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,7 +89,7 @@ RUN wget -qO- "https://github.com/koalaman/shellcheck/releases/download/stable/s
 #####################
 # Install Go Linter #
 #####################
-ARG GO_VERSION='v1.23.7'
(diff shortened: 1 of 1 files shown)


marbar3778:
would love to see this merged, its currently a blocker for me to update all our go products. It would be nice to use: https://github.com/marketplace/renovate to update the docker version on releases of various linters. 

Moar Bots

admiralAwkbar:
Merged into #181 to run CI

admiralAwkbar:
Pulled into #183 for full ci run and build out

PR #171, opened 2020-06-19
Allow for workspace location to be provided by the end user

fastloris:
This will give more flexibility to anyone who doesn't happen to keep their files in /tmp/lint and is running it locally.


admiralAwkbar:
Pulling this into #186 for full CI

PR #177, opened 2020-06-19
Update the notation of RuboCop

koic:
RuboCop is the proper notation as below.

> **RuboCop** is a Ruby static code analyzer (a.k.a. `linter`) and code formatter.

https://github.com/rubocop-hq/rubocop/blob/v0.85.1/README.md


admiralAwkbar:
@koic thank you for the catch!

admiralAwkbar:
As text only, no need for CI

PR #182, opened 2020-06-19
Fix: Repair Links in CONTRIBUTING.MD

JustinGrote:
Links pointed to github-demo which is 404, updated to point to super-linter


admiralAwkbar:
@JustinGrote thank you for the catch!

admiralAwkbar:
As an only text, no need to pull into formal CI

PR #186, opened 2020-06-19
Default workspace

admiralAwkbar:
This will give more flexibility to anyone who doesn't happen to keep their files in `/tmp/lint` and is running it locally.


admiralAwkbar:
@fastloris I agree to give the user the flexibility to change the default location is much more helpful than forcing them :)

admiralAwkbar:
@zkoppert I think just need your eyes on this and were good...

fastloris:
Thanks everyone!

PR #191, opened 2020-06-19
Disable flag

### LINKED ISSUE #138: Add Support for disabling errors (info-only runs)

admiralAwkbar:
Resolves #138. 

* [x] Add environment variable to container
* [x] Update bash script to change exit code


zkoppert:
Add documentation somewhere?

admiralAwkbar:
@zkoppert was adding it to the `README.md` but wasn't sure if should be referenced anywhere else

admiralAwkbar:
@SeanKilleen thanks for the great addition!

admiralAwkbar:
@zkoppert this is golden... make it so..

SeanKilleen:
Thanks for adding the docs I missed! Glad this is useful.

PR #213, opened 2020-06-20
Add support more variants names of env files

mgrachev:
Hi there!

Right now, super-linter finds only these variants of `env` files:
```
foo.env
.env
```

In this PR I add support for additional variants:
```
.env.foo
.env.foo.example
```


admiralAwkbar:
@mgrachev very nice! Moving to internal branch for CI

PR #217, opened 2020-06-20
Turn the pipe operator into the proper or operator

ntchambers:
Currently the `linter.sh` file uses:

```
    if [ $ERROR_CODE -ne 0 ] | [ -z "${GET_VERSION_CMD[*]}" ]; then
```

which is effectively `test | test`. This will cause the first `test` to be ignored, and only the second test's exit status will be considered. This changes it to `test || test` where it considers both

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -223,7 +223,7 @@ GetLinterVersions()
     ##############################
     # Check the shell for errors #
     ##############################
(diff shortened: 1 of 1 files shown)


zkoppert:
Need to merge this to a staging branch when I get back to my laptop.

PR #221, opened 2020-06-21
Add `editorconfig-checker`

mstruebing:
see: https://github.com/editorconfig-checker/editorconfig-checker/

I hope this kind of linter is allowed as it is not for linting a specific language.


mstruebing:
Any news on this?

admiralAwkbar:
@mstruebing this is interesting...
With all the various languages being listed, wouldn't this cause issues?
I.e. I have python and bash in a repo, one uses tabs, one uses spaces.

mstruebing:
As long as the `.editorconfig` have the correct rules this isn't a problem at all.
With an `.editorconfig` you are able to define basic formatting for different filetypes.
See: https://editorconfig.org/

lib/linter.sh
```diff
@@ -2504,6 +2521,17 @@ if [ "$VALIDATE_KOTLIN" == "true" ]; then
...
 
+########################
+# EDITORCONFIG LINTING #
+########################
+if [ "$VALIDATE_EDITORCONFIG" == "true" ]; then
+  ####################################
+  # Lint the files with editorconfig #
+  ####################################
+  # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILE_ARRAY"
+  LintCodebase "EDITORCONFIG" "editorconfig-checker" "editorconfig-checker" "^.*$" "${FILE_ARRAY_ENV[@]}"
```
nemchik, line 2532: Is this going to lint all dotfiles?

Is this linter supposed to lint all dotfiles? Or just the .editorconfig files?

Ex: should this lint .gitignore ? It looks like it will.

.automation/test/editorconfig-checker/editorconfig-checker_bad_1.ext
```diff
@@ -0,0 +1,3 @@
+ some line
```
nemchik, line 1: Depending on the answers to my other questions these files might need to be split into subfolders named good and bad, and the files might need to be named `.editorconfig` in each folder.

lib/linter.sh
mstruebing: Isn't that a regex which says: "start with any character n times and then end"?
So it will lint all files, by that it will consider your `.editorconfig` file if present.
So to be clear: the `.editorconfig` is the file which describes how your files should be formatted (basically, tabs/spaces, spaces amount, line ending character, like that), which can be used for any filetype and programming language. It's language agnostic. You can have wildcard rules or make rules for specific files or file globs.
So it will lint every file.

Does this explain it properly?

.automation/test/editorconfig-checker/editorconfig-checker_bad_1.ext
mstruebing: See above

lib/linter.sh
nemchik: I misread the regex. Sorry. You are correct that pattern will match any file in the repo.

So my question then changes: Is this linter meant to lint `.editorconfig` files, or is it meant to check that all files comply with the settings specified in the `.editorconfig` that applies to them?

mstruebing: It's exactly the second one. It checks that your files comply with your settings specified in `.editorconfig`.
Sorry for not explaining it properly.

nemchik: This makes more sense. What happens in a repo with no `.editorconfig`?

mstruebing: No error should occur, if there is no definition for a specific check it is skipped and not assuming something defaultish.

.automation/test/editorconfig-checker/editorconfig-checker_bad_1.ext
nemchik: Makes sense now with the other conversation we've had.

mstruebing:
Currently rebased :)

zkoppert:
resolved conflicts caused by the html PR getting merged in first.

zkoppert:
@nemchik Do you believe this one is ready?

nemchik:
LGTM.

zkoppert:
fixing up my merge mistakes and then sending this through if it passes CI.

PR #233, opened 2020-06-22
Adding Rust support via Clippy

### LINKED ISSUE #108: Rust support
Using https://github.com/rust-lang/rust-clippy

plauche:
Resolves #108 

Questions...
- This currently defaults to the latest version `rustup` installs. Should it be hard coded to a specific version?
- `clippy` has some support for [configuration files](https://github.com/rust-lang/rust-clippy#configuration) but it looks like they need to be included in the Rust project itself, not passed in as arguments to `clippy`. I wasn't sure how or if I should try to add these config files to the super linters config file setup.


plauche:
Well in my enthusiasm I forget/ignored that `clippy` is designed to be run against crates, not individual rust source files...currently rethinking how this might work.

PR #220, opened 2020-06-21
Spelling

jsoref:
Misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

It reported: https://github.com/jsoref/super-linter/commit/19c02d60fafd71fd945a76242c47e6868d5e44c1#commitcomment-40055878

And it validated that the changes in this PR made it happy: https://github.com/jsoref/super-linter/commit/19c02d60fafd71fd945a76242c47e6868d5e44c1

Note: this PR does not include the action, if you're interested, that can be offered separately.

PR #235, opened 2020-06-22
Add check spelling action

### LINKED ISSUE #228: Consider using check-spelling action
We seem to have a tendency to introduce spelling errors. We might consider adding https://github.com/marketplace/actions/check-spelling

GaboFDC:
This adds the check spelling actions to avoid some typos. Fixes #228 

Note:
This is sill WIP, as I would like early feedback on the way I'm handling the expected words, and all the tweaking needed.
Also, ping @jsoref as you did the initial PR #220 fixing some things (and are the creator of the action), I'm not sure the config you used. Fell free to leverage my branch to add up any work you may already done on this.

Update: This is now reade for review


jsoref:
This is the config I used: https://github.com/jsoref/super-linter/commit/19c02d60fafd71fd945a76242c47e6868d5e44c1
It's the laziest form.

Splitting things out by files as you did is a totally reasonable choice.

Note that `expect` is for words that you're temporarily tolerating. If you want a word to be treated as part of the dictionary, you can use `allow` instead. So, the things in `.github/actions/spelling/expect/words.txt` would probably be better served in `.github/actions/spelling/allow/words.txt`.

.github/actions/spelling/allow/Dockerfile.txt
```diff
@@ -0,0 +1,9 @@
+musl
+nvq
+rdoc
+setuptools
+uninstall
+xJv
```
jsoref: I'm surprised by `xJv` -- I don't think that the tool would generate this (it should see `x` + `Jv`)
And if I were adding it, I'd put this in `expect` as it really isn't a "word" -- `setuptools` and `uninstall` are closer to words, whereas `Jv` is a tangential flag...

(The same for `xzv`)

.github/actions/spelling/allow/ghe-config-apply.sh.txt
```diff
@@ -0,0 +1,3 @@
+aef
```
jsoref, line 1: >      echo "Were good to move forward, no .pid file found at:[$GHE_CONFIG_PID]"

This should be `We're` (the spell checker is not a grammar checker and won't catch this...)

Again, I wouldn't put this into `allow`,

I might add `\bps -\w+\b` into `patterns`.

.github/actions/spelling/allow/gitignore.txt
```diff
@@ -0,0 +1,8 @@
+cov
```
jsoref, line 1: You might choose to just add `^\.gitignore$` or `ignore$` to an `exclude` -- if it's a file you're just importing and not really changing.

That said, it's a pretty short list and might be useful this way.

.github/actions/spelling/allow/Dockerfile.txt
GaboFDC: Yes this was a mistake on my side. Thinking about adding `\btar -\w+\b` into patters, WDYT?

.github/actions/spelling/allow/ghe-config-apply.sh.txt
GaboFDC: I resolved to add `\b -\w+\b` to catch any cli args not sure if this patter is safe enough.

.github/actions/spelling/patterns.txt
```diff
@@ -3,3 +3,5 @@ https?:\S*
 ([A-Za-z])\1{3,}
 # Any CLI args (-xzf -aef)
 \ -\w+\b
+# Hex colors (dummy group to not be treated as comment)
+(#)([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})
```
jsoref, line 7: I tend to use `(?:...)` instead of `(...)` for such cases

GaboFDC:
Thanks for the feedback @jsoref, I implemented the change from expect to allow, and some general patterns for CLI args. About gitignore, I think it doesn't hurt to check it, so like this is fine.
About expect.words.txt I added it because it seemed like the action was complaining about it. But after testing seems like it fails anyways, can you give it a look? not sure why it still fails.

.github/actions/spelling/allow/ghe-config-apply.sh.txt
jsoref: Should be ok.

As long as you use `--longFlags` you'd still catch `--loongFlags`.

Offhand, I can't think of any particular cases where ` -foo` appears in a normal English context. (There are one or two outlier tools which do `-longFlags`, but as long as you don't have those and don't specifically have them in a comment/rare-if-branch/document your code should break if those are misspelled 😄 .)

The tool is best effort, you should still have people reviewing your code, and hopefully they'll catch the rest 😄 .

jsoref:
I probably need to rethink mandating `expect` (it add a different name in the past as well...). I might relax that for `0.17`.

I only recently added the ability to supplement the dictionary (`allow`), so the original model was:
1. User runs and gets a failure identifying which things they should add to the list of expected words file (because nothing short of a *very* empty repository would have nothing, -- just adding the workflow adds `ubuntu` which isn't in the dictionary...)
1. User adds their `expect` file
1. Future runs complain about things diverging

At this point, many things have changed, and it's probably not really a big deal if there isn't an expect file...

The file should be called either `expect.txt` or `expect/something.txt` -- the name `expect.words.txt` in `/tmp` is an internal thing... I should perhaps not use that name, as it's confusing.

It looks like it's missing a step to pick the proper filename... I'll have to look into that.

jsoref:
Heh, it really wants a blank line in that file:
```github-log
##[warning].github/actions/spelling/expect.txt: line 1, columns 0-1, Warning - no newline at eof (no-newline-at-eof)
```

.automation/test/ansible/ghe-initialize/templates/ghe-config-apply.sh
```diff
@@ -190,7 +190,7 @@ CheckGHEProcess
 RunConfigApply
 
 ##########################################
-# Were going to run it again after a nap #
+# We're going to run it again after a nap #
 # to make sure there is no crazy actions #
 ##########################################
```
zkoppert, line 195: Can we fix alignment?
```suggestion
# We're going to run it again after a nap #
# to make sure there is no crazy actions  #
###########################################
```

.automation/test/ansible/ghe-initialize/templates/ghe-config-apply.sh
```diff
@@ -190,7 +190,7 @@ CheckGHEProcess
 RunConfigApply
 
 ##########################################
```
zkoppert, line 192:
```suggestion
###########################################
```

.github/workflows/check-spelling.yml
```diff
@@ -0,0 +1,47 @@
...
+      # Checkout the code base #
+      ##########################
+      - name: Checkout Code
+        uses: actions/checkout@v2
+
+      #############################
+      # Run check spelling action #
+      #############################
+      - name: Check spelling
+        uses: check-spelling/check-spelling@0.0.16-alpha
```
zkoppert, line 43: Is this project relatively stable even though its in alpha?

lib/linter.sh
```diff
@@ -3,7 +3,7 @@
 
 ################################################################################
 ################################################################################
-########### Super-Linter (Lint all the code) @AdmiralAwkbar ####################
+########### Super-Linter (Lint all the code) @Admiralawkbar ####################
```
zkoppert, line 6: inconsistent with other changes
```suggestion
########### Super-Linter (Lint all the code) @admiralawkbar ####################
```

.github/workflows/check-spelling.yml
jsoref: I make releases somewhat periodically, but I try not to break things. The biggest change I've made was the terminology for `expect`/`allow`/`reject` and I did that in a way that didn't break compatibility (it just added a warning and I try to make PRs to consumers to migrate them).

I'm planning to make a 0.0.17-alpha release shortly which will offer [act](https://github.com/nektos/act) support.

I have a few more [planned features](https://github.com/check-spelling/check-spelling/wiki/Possible-features), but they shouldn't cause backwards compatibility issues.

zkoppert:
@admiralAwkbar If adding this functionality is okay with you, I think this one is good to go. If your cool with it I can pull it into a staging branch.

admiralAwkbar:
This has been pulled in for full CI

PR #239, opened 2020-06-22
Fix shellcheck exclusions

nemchik:
Let's fix all the things so we don't need `# shellcheck disable=` :smile:


admiralAwkbar:
Awesome work @nemchik. Pullin in for CI

PR #241, opened 2020-06-22
Fix shellcheck exclusions

admiralAwkbar:
Let's fix all the things so we don't need `# shellcheck disable=` :smile:

Pulling in #239 for CI

PR #245, opened 2020-06-22
Fix ansible find

nemchik:
Following up on https://github.com/github/super-linter/pull/241 and https://github.com/github/super-linter/pull/239

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -1944,7 +1944,7 @@ TestCodebase()
     #################################
     # Get list of all files to lint #
     #################################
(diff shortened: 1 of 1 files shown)


nemchik:
It took a bit of digging to understand what was going on here (thus the multiple commits, I can squash if needed). If I have correctly understood the need: ignore `$GITHUB_WORKSPACE/$TEST_CASE_FOLDER/ansible/ghe-initialize/*`.

This fixes all of the issues except https://github.com/github/super-linter/pull/241/checks?check_run_id=796505252#step:5:326 and I'm not 100% sure what to make of these errors. It looks like a file being tested that should be good but the linter says it's not.

admiralAwkbar:
@nemchik can you try this again in a new PR, its all tore up...

PR #224, opened 2020-06-21
Add kotlin support to super-linter

Kim1Jun:
based on [wiki](https://github.com/github/super-linter/wiki/Adding-new-language-support)

close #107

PR #246, opened 2020-06-22
Kotlin support

### LINKED ISSUE #107: Kotlin support
I believe klint is the most popular kotlin linter and shouldn't be too opinionated:

> ktlint tries to capture (reflect) official code style* from kotlinlang.org and Android Kotlin Style Guide (+ we respect your .editorconfig and support additional ruleset|s).

https://github.com/pinterest/ktlint

If you were to create a super-static-analyser 
It could use https://github.com/detekt/detekt

admiralAwkbar:
based on [wiki](https://github.com/github/super-linter/wiki/Adding-new-language-support)

close #107

This closes #224


sschuberth:
Sad to see this only uses `ktlint` instead of `detekt` plus `ktlint` plugin as suggested [here](https://github.com/github/super-linter/issues/107#issuecomment-646558756).

Issue #204, opened 2020-06-19
Clearify what is the intended workflow for using Super Linter locally


essenmitsosse:
Coming from [ESLint](https://github.com/eslint/eslint) my workflow is kind of like that (super simplified), if you clone a repo, that is setup with ESLint, here is basically what you have to do:

- `npm install`
- `npm run lint` (or whatever it is called)

Something as simple as this doesn`t seem to be possible with Super Linter for obvious reasons, because you would have to setup the docker container, and then maybe you could add the docker script as a NPM script. But it seems like, that is not how Super Linter is even intended to use.

So the question is: **How is Super Linter meant to integrate with day-to-day local development.**

I am still a bit confused. For me a Linter mainly is used locally while developing - auto fixing being a must have. Having a linter check run as part of a CI in a VM is an added bonus. Super Linter seems to completely focused on CI.

My best guess would be: Keep on using ESLint through NPM locally and just use Super Linter action for CI.

Either way, would be great if that could be clarified in the Readme.

kkweon:
+1 Please add a documentation on how to run it locally. There could be a discrepancy in versions or configs. Running a local linter could cause a trouble.

essenmitsosse:
[There already is one](https://github.com/github/super-linter/blob/master/docs/run-linter-locally.md), but it seems to be more for debugging super linter, then to actually use it everyday in a project.

zkoppert:
Currently my workflow is to commit and push often to a pull request and let the ci do its thing while I keep working. I would be open to suggestions on how we might accommodate more local use if it could integrate nicely. Have to spend some time brainstorming that.

essenmitsosse:
I think the major difference is how immediately you want feedback. For me a linter is not just supposed to make sure no low-quality code is submitted, its purpose is also to help me get readable code right as I write it, as well as give instant feedback for errors like missing variables. I have a shortcut in the IDE, that formats the whole file I am working on. This way I can just hack down code and then let the linter do its magic regarding formatting. Have to submit the code to CI every time would be some orders of magnitude slower than the current workflow. Also I'd be pushing a huge amount of unnecessary commits to origin.

Or to put it in other words: I let a linter clean up code way more often, then I need feedback from the whole CI pipeline, even when doing tiny, granular commits.

It should be noted, this is not just about auto fixing, but also about integration with the IDE.

I think the question is: Does Super Linter want to be a replacement for everything linter related, or does it just want to streamline the CI process?

GaboFDC:
[...]

PR #267, opened 2020-06-24
Add disclaimer about local development

zkoppert:
This addition to the `README.md` is to help clarify incoming requests about why run linting at the time of pull request (ci) instead of at the time of code writing (local).

closes https://github.com/github/super-linter/issues/204


essenmitsosse:
This is great. Definitely clears up any confusion about that matter.

PR #269, opened 2020-06-24
Debug with vscode

zkoppert:
staging changes to configure vscode debugging

This adds: 

- [x] Support for [ Visual Studio Code Docker Remote Development](https://code.visualstudio.com/docs/remote/containers). 
- [x] Support for [Github Codespaces](https://github.com/features/codespaces/) once it becomes available for the repo.
- [x] Default debug task (F5) using Bash Debug so that developers can step through the linter script for debugging and validation

Capture

cc/ @JustinGrote

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,37 @@
+// For format details, see https://aka.ms/vscode-remote/devcontainer.json or this file's README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.123.0/containers/docker-existing-dockerfile
+{
+	"name": "SUPER-LINTER",
+
+	// Sets the run context to one level up instead of the .devcontainer folder.
+	"context": "..",
+
+	// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
+	"dockerFile": "..\\Dockerfile",
+
+	// Set *default* container specific settings.json values on container create.
+	"settings": { 
+		"terminal.integrated.shell.linux": "/bin/bash"
+	},
+
+	// Add the IDs of extensions you want installed when the container is created.
+	"extensions": [
+		"github.vscode-pull-request-github",
+		"rogalmic.bash-debug"
+	],
+
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	// "forwardPorts": [],
+
+	// Uncomment the next line to run commands after the container is created - for example installing curl.
+	//"postCreateCommand": "ln -s /workspaces/super-linter/.automation/test /tmp/lint",
+
+	// Uncomment when using a ptrace-based debugger like C++, Go, and Rust
+	// "runArgs": [ "--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined" ],
+
+	// Uncomment to use the Docker CLI from inside the container. See https://aka.ms/vscode-remote/samples/docker-from-docker.
+	// "mounts": [ "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind" ],
+
+	// Uncomment to connect as a non-root user. See https://aka.ms/vscode-remote/containers/non-root.
+	// "remoteUser": "vscode"
+}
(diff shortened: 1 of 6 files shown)


zkoppert:
I could use some help resolving linting errors if anyone has some free time. 😁

zkoppert:
Looks like 
- JSON comments are an issue (https://github.com/github/super-linter/pull/269/checks?check_run_id=805142548#step:4:87)
- missing double quotes (https://github.com/github/super-linter/pull/269/checks?check_run_id=805142548#step:4:137)
- trailing spaces in the readme (https://github.com/github/super-linter/pull/269/checks?check_run_id=805142548#step:4:126)

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,37 @@
+// For format details, see https://aka.ms/vscode-remote/devcontainer.json or this file's README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.123.0/containers/docker-existing-dockerfile
+{
+	"name": "SUPER-LINTER",
+
+	// Sets the run context to one level up instead of the .devcontainer folder.
+	"context": "..",
+
+	// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
+	"dockerFile": "..\\Dockerfile",
```
GaboFDC, line 10: @JustinGrote in what OS did you tested? This fails for me and I needed to change for slashes `..//Dockerfile`

README.md
```diff
@@ -185,6 +185,13 @@ Below are a list of the known limitations for the **GitHub Super-Linter**:
 ## How to contribute
 If you would like to help contribute to this **GitHub** Action, please see [CONTRIBUTING](https://github.com/github/super-linter/blob/master/.github/CONTRIBUTING.md)
 
+### Visual Studio Code
+You can checkout this repository using [Container Remote Development](https://code.visualstudio.com/docs/remote/containers), and debug the linter using the `Test Linter` task.
```
GaboFDC, line 189: @JustinGrote Maybe also, we should specify that https://marketplace.visualstudio.com/items?itemName=rogalmic.bash-debug extension is needed

JustinGrote: @GaboFDC it's autoinstalled by the devcontainer.json, so it shouldn't be explicity needed, though it can be added to the extensions.json file as well to auto-recommend it, but I don't think it needs to be in the readme here.

.devcontainer/devcontainer.json
JustinGrote: @GaboFDC Windows, good catch. The // should work fine on windows.

GaboFDC:
@zkoppert I gave it a got to try and fix linters.

Not sure if for https://github.com/GaboFDC/super-linter/runs/807834326?check_suite_focus=true#step:4:160 we should just disable it, or if I'm doing something wrong

And for JSON, we would need to remove all comments, not sure what are your thoughts on that?

README.md
zkoppert: Since this is autoinstalled, lets just leave this documentation as is.

zkoppert:
> Not sure if for https://github.com/GaboFDC/super-linter/runs/807834326?check_suite_focus=true#step:4:160 we should just disable it, or if I'm doing something wrong

There is another solution to try before disabling. Check out the recommendation [here](https://github.com/koalaman/shellcheck/wiki/SC1091#correct-code) that suggests adding a comment `# shellcheck source=somefile`. If that doesn't work then we can just disable the offending line.

> And for JSON, we would need to remove all comments, not sure what are your thoughts on that?

I think the comments in there are helpful. My suggestion would be moving those comments in the JSON file to a `README.md` file in that `.vscode` directory.

GaboFDC:
@zkoppert yes my test included the [source line](https://github.com/github/super-linter/compare/Debug-with-vscode...GaboFDC:Debug-with-vscode#diff-a09c393e7eaeb0250d6d2628eee3a7cfR11)

That's the only error remainig. Not sure if I should open another PR, or what would be the workflow here.

zkoppert:
Yeah go ahead and open a new PR targeting to merge into this branch. That PR can move comments out of the JSON file and into a `.vscode/README.md`.

GaboFDC:
https://github.com/github/super-linter/pull/293

admiralAwkbar:
Pulling into #304

PR #273, opened 2020-06-25
workflows/stack-linter: run on forks

MikeMcQuaid:
As discussed with @zkoppert, this should enabled CI to run on forks to this repository.


MikeMcQuaid:
This may created both `push` and `pull_request` CI jobs. In that case we can filter out branches by using something like:

https://github.com/Homebrew/brew/blob/adbc50f15130b18dd1358d24923921a7a199ee21/.github/workflows/tests.yml#L1-L5

Which branches would we want to run this on?

.github/actions/spelling/expect.txt
```diff
@@ -32,6 +32,7 @@ pwsh
 rcu
 rhc
 ry
+tada
```
jsoref, line 35: @zkoppert: #277 is an alternative to this

jsoref:
@zkoppert: there's always a race edge between PRs made before the action is added and integrating the spell checker. For some projects, I've actually created merges/rebases and offered PRs to the pending PRs to address this. I can do that here if you'd like...

I'm open to advice on how to avoid this race.

zkoppert:
> > Which branches would we want to run this on?
> 
> @zkoppert Nudge on this.

We want to run this on all pull requests to master branch. @MikeMcQuaid I'm struggling to find the right syntax to accomplish that so we don't have duplicate jobs. advice?

MikeMcQuaid:
@zkoppert Sorry, will rephrase the question: which branches do you wish to run this on _without_ pull requests e.g. `master`, any others?

jsoref:
So, you can choose to run it only on `push`, at the risk of forks not having actions enabled. You can run it only on `pull_requests`, or you can run it on both. You might choose only to run it on `pull_request`.

My initial deployments were on push to enable devs to test things w/o having to open a pull request, and to update things in a local branch before updating their PR.

Over time, I'm probably going to move to recommending using only `pull_request`, and suggest devs use `act` to test locally, I'm close to that, but not quite there yet (hopefully 0.0.17, and it looks like at least one more release in nektos/act).

nemchik:
Is there really any harm to having both push and pull? If the response is "it makes two checks on each PR" I would ask what significance that has? Both checks should pass before merging, and having both allows the most flexibility for people forking and contributing.

jsoref:
The only real harm is that you see the same comment twice. I'm hoping to eventually teach the bot to collapse identical comments, but I haven't gotten to that (it requires browsing the list of comments, and I worry about it scaling poorly for long lived/commented PRs).

zkoppert:
@MikeMcQuaid none others besides master

MikeMcQuaid:
@zkoppert Gotcha. I believe this should be what you want, then (as it's only running a `pull_request` job now).

zkoppert:
Thanks a ton @MikeMcQuaid ! Every time I looked at the documentation to try to see how this works it was late and brain was toast. Looks great! ❤️

MikeMcQuaid:
@zkoppert Happy to help!

PR #283, opened 2020-06-26
Specify dockerfilelint config path - feedback requested

past-due:
- LintCodebase: Support linter cmd trailing options
- Add dockerfilelint config option to actually use the `.dockerfilelintrc`
   - Reference: https://github.com/replicatedhq/dockerfilelint/pull/138

Help testing & feedback / improvements / refactoring welcome.

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -1681,6 +1681,7 @@ LintCodebase()
   LINTER_NAME="$1" && shift     # Pull the variable and remove from array path  (Example: jsonlint)
   LINTER_COMMAND="$1" && shift  # Pull the variable and remove from array path  (Example: jsonlint -c ConfigFile /path/to/file)
   FILE_EXTENSIONS="$1" && shift # Pull the variable and remove from array path  (Example: *.json)
+  LINTER_TRAILING_OPTIONS="$1" && shift  # Pull the variable and remove from array path  (Example: -c ConfigFile)
   FILE_ARRAY=("$@")             # Array of files to validate                    (Example: $FILE_ARRAY_JSON)
 
   ######################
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@past-due I think this is great but i think the path we will have e to go will be to have a `mapping.yml` file that will allow for users to specify if they want trailing arguments, additional pre arguments, etc...
That would be the best way to solve this, but its a major change, and will need some work.

Something like `mappings.yml`:
```yml
---
Docker:
  docker_post_arguements: -o somestuff
JSON:
  json_additional_extensions:
    - .file
    - .config
...
```

past-due:
A `mapping.yml` sounds like a great long-term enhancement, but I believe this current (minimal) change to specify the  dockerfilelint config path is needed.

By default, `dockerfilelint` only picks up a `.dockerfilelintrc` that is located in the same directory as the Dockerfile that's being analyzed (see: https://github.com/replicatedhq/dockerfilelint/pull/138#issue-361609817). So specifying the `-c $DEFAULT_RULES_LOCATION/` for the call to `dockerfilelint` would appear to be required to get it to actually use the `.dockerfilelintrc` for all Dockerfiles, as is presumably expected.

GaboFDC:
I think the idea behind is good, but right now it differs from the standard that some linters have, where the config file is already specified, like [MD](https://github.com/github/super-linter/pull/283/files#diff-2b006ea0cfcfac29e33a1b9eef9974c1R2319), [Python](https://github.com/github/super-linter/pull/283/files#diff-2b006ea0cfcfac29e33a1b9eef9974c1R2319), [Ruby}(https://github.com/github/super-linter/pull/283/files#diff-2b006ea0cfcfac29e33a1b9eef9974c1R2319), and so on. 

So I think it is better if for docker we use the same pattern that we already have in the others. ANd you can use the `LINTER_PATH` to specify where the dockerfilelint is

past-due:
> So I think it is better if for docker we use the same pattern that we already have in the others. And you can use the `LINTER_PATH` to specify where the dockerfilelint is

The issue is that this currently does not work, because the parameter that tells dockerfilelint where to find the `.dockerfilelintrc` is located is not passed.

admiralAwkbar:
@past-due it seems we found the bug and its been  merged into master now

past-due:
> @past-due it seems we found the bug and its been merged into master now

@admiralAwkbar: I just tested 62717c09c755da78a83a4d8957093bf44c5ad6de (which includes 368a82290ec3f79bca4bb1f4805b3cb2fe3add7a) and it does **_not_** fix the problem, for the reasons I outlined in https://github.com/github/super-linter/pull/289#issuecomment-650231760

past-due:
Superseded by #397

PR #289, opened 2020-06-26
fixed the open Docker issue

### LINKED ISSUE #285: Configuration file of Dockerfilelint ignored
**Describe the bug**
Dockerfilelint seems ignoring the configuration file `.dockerfilelintrc` in  `.github/linter/`.

After some research, it appears that Dockerfilelint does not search for a configuration file at the root of the repository or where Dockerfilelint is run, but expects a configuration file in the same directory than the Dockerfile to lint. If a repository contains several Dockerfile in separates directories, each directory is expected to contain its `.dockerfilelintrc`. So the file  in the `.github/linter/` seems to be logically ignored by Dockerfilelint, even if correctly taken into account by super-linter (with the message "User provided file:[.dockerfilelintrc], setting rules file..." in debug mode).

**To Reproduce**
1. Create a repository with a Dockerfile triggering a warning (`FROM ubuntu:lastest` will trigger "Base images should not use the latest tag").
2. Add a `.dockerfilelintrc` in `.github/linters/` containing:

    ---
    rules:
      latest_tag: off
    ...
3. Run the super-linter

**Expected behavior**
No error should be reported by Dockerfilelint.

**Workaround**
Instead of modifying super-linter (or Dockerfilelint or both), maybe mentioning in the documentation to place the `.dockerfilelintrc` file next to the Dockerfile instead of in `.github/linter/` could be enough.

admiralAwkbar:
When we introduced `dockerfilelint` there was no way to pass a config file, and the file needed to be in the same directory. Once we made PR to the main directory, they introduced the feature. We can now add the logic to view the `.dockerfilelintrc`

closes #285

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -2289,7 +2289,7 @@ RunTestCases()
   TestCodebase "JAVASCRIPT_STANDARD" "standard" "standard $JAVASCRIPT_STANDARD_LINTER_RULES" ".*\.\(js\)\$"
   TestCodebase "TYPESCRIPT_ES" "eslint" "eslint --no-eslintrc -c $TYPESCRIPT_LINTER_RULES" ".*\.\(ts\)\$"
   TestCodebase "TYPESCRIPT_STANDARD" "standard" "standard --parser @typescript-eslint/parser --plugin @typescript-eslint/eslint-plugin $TYPESCRIPT_STANDARD_LINTER_RULES" ".*\.\(ts\)\$"
-  TestCodebase "DOCKER" "/dockerfilelint/bin/dockerfilelint" "/dockerfilelint/bin/dockerfilelint" ".*\(Dockerfile\)\$"
+  TestCodebase "DOCKER" "/dockerfilelint/bin/dockerfilelint" "/dockerfilelint/bin/dockerfilelint -c $DOCKER_LINTER_RULES" ".*\(Dockerfile\)\$"
   TestCodebase "ANSIBLE" "ansible-lint" "ansible-lint -v -c $ANSIBLE_LINTER_RULES" "ansible-lint"
   TestCodebase "TERRAFORM" "tflint" "tflint -c $TERRAFORM_LINTER_RULES" ".*\.\(tf\)\$"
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@jwiebalk I think this is the final work needed on your PR from some time ago :)

admiralAwkbar:
@past-due I think this is the best way to achieve the fix needed for the config file for the docker file lint

past-due:
@admiralAwkbar: My impression from https://github.com/replicatedhq/dockerfilelint/pull/138 was that:
- dockerfilelint's `-c` option expects the _path_ to the `.dockerfilelintrc`, i.e. _not including the filename_. `DOCKER_LINTER_RULES` includes the filename, so presumably would not work properly.
- dockerfilelint [expects options to be provided at the end of the command line](https://github.com/replicatedhq/dockerfilelint#command-line-options)

Hence the adjustments in #283

past-due:
@admiralAwkbar: I just tested 62717c09c755da78a83a4d8957093bf44c5ad6de (which includes 368a82290ec3f79bca4bb1f4805b3cb2fe3add7a) and it does **_not_** fix the problem. The `.dockerfilelintrc` is not being used by dockerfilelint. See my comment above: https://github.com/github/super-linter/pull/289#issuecomment-650231760

admiralAwkbar:
@past-due i tested and validated it, the command runs: `/dockerfilelint/bin/dockerfilelint -c .dockerlintrc <Dockerfile>` which ius using the config file as intended.

Am I missing something else on this?

past-due:
> @past-due i tested and validated it, the command runs: `/dockerfilelint/bin/dockerfilelint -c .dockerlintrc <Dockerfile>` which ius using the config file as intended.
> 
> Am I missing something else on this?

`-c .dockerlintrc`
should be
`-c <path to folder _containing_ .dockerlintrc>`

The command it is executing will execute without error, but it's not using the actual `.dockerlintrc` in my tests. In https://github.com/replicatedhq/dockerfilelint/blob/master/lib/index.js#L85 you can see it will fail to open `.dockerlintrc/.dockerlintrc` and just silently use the defaults.

LukeMondy:
I can confirm I'm seeing the same behavior at the moment. `.dockerfilelintrc` is ignored still.

zkoppert:
opened up an issue ☝🏻 #305 so we can work to get this fixed.

zkoppert:
Let's continue the conversation in that issue

PR #281, opened 2020-06-26
Add JAVASCRIPT_ES_CONFIG_FILE to specify eslintrc filename / format

past-due:
See: https://eslint.org/docs/user-guide/configuring#configuration-file-formats

ESLint supports many different configuration filenames and formats. This PR adds a new ENV variable `JAVASCRIPT_ES_CONFIG_FILE` (which defaults to the previous fixed `.eslintrc.yml`) to support specifying which file to use.

(In our case, we use `.eslintrc.json`)

PR #292, opened 2020-06-26
Add JAVASCRIPT_ES_CONFIG_FILE to specify eslintrc filename

admiralAwkbar:
See: https://eslint.org/docs/user-guide/configuring#configuration-file-formats

ESLint supports many different configuration filenames and formats. This PR adds a new ENV variable `JAVASCRIPT_ES_CONFIG_FILE` (which defaults to the previous fixed `.eslintrc.yml`) to support specifying which file to use.

(In our case, we use `.eslintrc.json`)

this closes #281


admiralAwkbar:
@past-due Great work on this, I just updated to handle fore `TypeScript` as well

thank you for all the solid work!

PR #294, opened 2020-06-26
Fix honoring pylint config file

dsayling:
The pylint config file is correctly being applied to
the pylint command, however, the `-E` flag is actually
not honoring any of the disabled messages in the
`.python-lint` file. By making this change to the
default lint file only the errors are output as expected
but honoring the disabled messages

For testing I did, see this gist: https://gist.github.com/dsayling/950a128dd771a11d0663c5bf63ea4ad3

### DIFF
--- a/.github/linters/.python-lint
+++ b/.github/linters/.python-lint
@@ -1,4 +1,5 @@
 [MASTER]
+errors-only=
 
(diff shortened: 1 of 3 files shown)


admiralAwkbar:
Pulling into a branch for full CI

PR #301, opened 2020-06-29
Fix honoring pylint config file

admiralAwkbar:
The pylint config file is correctly being applied to
the pylint command, however, the `-E` flag is actually
not honoring any of the disabled messages in the
`.python-lint` file. By making this change to the
default lint file only the errors are output as expected
but honoring the disabled messages. Also, disable the score
as the -E flag also does.

closes #294


admiralAwkbar:
@dsayling thanks for the catch and cool fix!

PR #300, opened 2020-06-29
Read Clojure filetypes from $FILE_TYPE, not $FILE

Reefersleep:
`super-linter` would previously attempt to compare Clojure file endings (such as `.clj` and `.cljc`) to `$FILE`, not `$FILE_TYPE`.

When running with `VALIDATE_CLOJURE: true` for a repo containing Clojure code, this would result in warnings like these in the linting output:
```
  - WARN! 
Failed to get filetype for:[project/src/my_ns.clj]!
```
And no validation of the mentioned Clojure files.

PR #302, opened 2020-06-29
Read Clojure filetypes from $FILE_TYPE, not $FILE

admiralAwkbar:
`super-linter` would previously attempt to compare Clojure file endings (such as `.clj` and `.cljc`) to `$FILE`, not `$FILE_TYPE`.

When running with `VALIDATE_CLOJURE: true` for a repo containing Clojure code, this would result in warnings like these in the linting output:
```
  - WARN! 
Failed to get filetype for:[project/src/my_ns.clj]!
```
And no validation of the mentioned Clojure files.

This closes #300

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -1695,7 +1695,7 @@ BuildFileList()
       # Set the READ_ONLY_CHANGE_FLAG since this could be exec #
       ##########################################################
       READ_ONLY_CHANGE_FLAG=1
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@Reefersleep Thanks for the catch on this one!

Reefersleep:
Will this be working after a version bump? And if so, what's your cadence for that?

PR #306, opened 2020-06-29
Creating more functions

admiralAwkbar:
This PR will cover:
- Breaking the monolith `linter.sh` into a main, and sub-functions files

This should help with maintaining the structure and allow for easier growth


zkoppert:
Do we need to edit the wiki page on how to add a new language support to reference which of these new files to edit?

zkoppert:
Love that this has its own readme. The main one is getting large.

admiralAwkbar:
@zkoppert all functions are still labeled the same, but I think we should update the doc to show the new file locations as well :)

admiralAwkbar:
@zkoppert Updated wiki

admiralAwkbar:
@zkoppert yes I think this will be v3...

PR #322, opened 2020-06-30
Better customisation instructions

### LINKED ISSUE #316: Explain how to tweak the rules better
I started using this great tool as a GitHub action but was initially confused how to configure and tweak the various linters.

There is a [Disabling linters and Rules](https://github.com/github/super-linter/blob/master/docs/disabling-linters.md) page with some great links to the various tools, but it still took me a bit of figuring out how to apply any customisations.

I eventually figured out that the answer is to copy the entire TEMPLATE configuration file for the linter you are using from this repo into `.github/linters` and then tweak it. It seems it is not possible to keep using the base config (and benefit from any updates to it) and then override certain rules. Is that correct?

If so, it seems to me we could benefit from an introductory paragraph at the top if this page. So instead of this:

> If you find you need to ignore certain errors and warnings, you will need to know the format to disable the Super-Linter rules.
> Below are examples and documentation for each language and the various methods to disable.

We change it to something along the lines of:

> Linters can often require configuring to ensure they work with your codebase and your team's coding style, to avoid flagging false-positives. Super-Linter has set up some default configuration for each linter which should work reasonably well with common code bases but many of the linters can be configured to disable certain rules, or configure the rules to ignore certain pieces of codes.
>
> To run with your own configuration for a linter, copy the relevant [TEMPLATE configuration file for the linter you are using from this repo](https://github.com/github/super-linter/tree/master/TEMPLATES) into the `.github/linters` folder in your own repository, and then edit it to modify, disable - or even add - rules and configuration to suite how you want your code checked.
>
> How the changes are made differ for each linter, and also how much Super-Linter has decided to change the linter's defaults. So, for some linters (e.g. [pylint for python](https://github.com/github/super-linter/blob/master/TEMPLATES/.python-lint)), there may be a large configuration file. For others (e.g. [eslint for JavaScript](https://github.com/github/super-linter/blob/master/TEMPLATES/.stylelintrc.json)) the default configuration file may initially be nearly empty. And for some (e.g. StandardJS) it may not be possible to change configuration at all so there is not Template file.
>
> Where a configuration file exists in your repo, it will be used in preference to the default one in the Super-Linter TEMPLATES directory (not in addition to it), and where one doesn't exist the TEMPLATES version will be used. So you should copy al configurations required from the template and not just the config you want to change.
>
> It is possible to have custom configuration for some linters, and continue to use the default from TEMPLATES directory for others so if you use Python and JavaScript and only need to tweak the Python rules, then only have a custom configuration for pylint and continue to use the default TEMPLATE from the main repo for ESLint, for example.
>
> For some linters it is also possible to override rules on a case by case level with directives in your code. Where this is possible we try to note how to do this in the specific linter sections below, but the official linter documentation will likely give more detail on this.

Happy to submit a PR for this if you think it would be useful and agree in general with my understanding of how this works?

bazzadp:
Fixes #316


admiralAwkbar:
@bazzadp Very nice!
 Pulling in for full ci

PR #326, opened 2020-06-30
Don't show shyaml errors to the console

nemchik:
Redirect stderr here. stdout will still pipe to grep.


admiralAwkbar:
Pulling in for full CI

PR #274, opened 2020-06-25
Build image locally for dev

nemchik:
This should eliminate the need to pull into local branches for CI.

PR #332, opened 2020-06-30
Build image locally for dev

nemchik:
Ref: https://github.com/github/super-linter/pull/274

Explanation:
https://github.com/github/super-linter/pull/273/files uses https://github.com/github/super-linter/blob/f1e46e8e76e65e3f96d06e780a9eebeeb701348e/.github/workflows/stack-linter.yml#L44 which pulls the latest release of the tool published to docker hub. This does not build the image locally and test using the image built. So you can lint the code prior to building the tool (which is good and necessary) but if you make a change to the code that would affect how the tool runs you would not be able to see validation that your changes work by pulling the latest release from docker hub.

The intention of this PR is to overcome the above challenge, and negate the need for contributors with permissions to merge on this repo to pull changes in for CI to run. CI will just run when a user opens a PR.

_Originally posted by @nemchik in https://github.com/github/super-linter/pull/274#issuecomment-652031578_


admiralAwkbar:
@nemchik At first I was against this idea, but the more I understand the use case for the community, this makes sense...

Pulling in for full CI, likely the last time we have to do that :)

nemchik:
Yeah at first the use case can be hard to explain or take in, but my next PR will just build on the spot and you'll be able to see if it's valid without pulling in for internal CI, and more importantly I'll be able to see if it passes or needs changes before I even open a PR, or during the review of the PR.

Thanks for taking the time to consider this!

PR #330, opened 2020-06-30
DockerFile: Add missing AllUsers Scope when install PSSA

alagoutte:
Fix #279

Issue #318, opened 2020-06-30
PSScriptAnalyzer Error: Invoke-ScriptAnalyzer (not recognized)


krowlandson:
**Describe the bug**
When running a vanilla implementation of Super-Linter (v3.0.1 / v.3.1.0), we get the following error within the PSScriptAnalyzer tests:

```bash
File:[<redacted>]
Invoke-ScriptAnalyzer: The term 'Invoke-ScriptAnalyzer' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
ERROR! Found errors in [pwsh] linter!
ERROR:[]
```

**To Reproduce**
1. Using Super Linter v3.0.1 / v.3.1.0 in a GitHub Workflow
2. With all default settings (no VAR defined)
3. Against a repository with PowerShell code present
4. See error against ALL detected files

**Expected behavior**
Expect PSScriptAnaylzer tests to run against all PowerShell files and only raise errors detected by linting.

**GitHub Actions YAML**
The below is the YAML file content used to control this GitHub Action:

```yaml
---
name: Super-Linter

on:
  push

jobs:
  super_linter_on_push:
    name: Run GitHub/Super-Linter on Push
    runs-on: ubuntu-latest
    steps:
      - name: Checkout GitHub Repository
        uses: actions/checkout@v2
      - name: Run GitHub/Super-Linter
        uses: docker://github/super-linter:v3.1.0
```

**Additional context**
Seems to work on local machine using same versions.

Rubemlrm:
@krowlandson there is another issue relative to this bug. The issue is the following #279

krowlandson:
Thank you @Rubemlrm - was literally just looking at that issue and was about to post back to say I think this is a duplicate.

Any thoughts on when this will be resolved, and why it doesn't appear to be a problem within a local run?

krowlandson:
Just to add to my findings, I've been experimenting and found that the following works fine on GitHub Actions:

```yaml
---
name: Super-Linter

on:
  push

jobs:
  super_linter_on_push:
    name: Run GitHub/Super-Linter on Push
    runs-on: ubuntu-latest
    steps:
      - name: Checkout GitHub Repository
        uses: actions/checkout@v2
      - name: Run GitHub/Super-Linter
        run: |
          docker run -e RUN_LOCAL=true -v $GITHUB_WORKSPACE:/tmp/lint github/super-linter:v3.1.0
```

[...]

PR #336, opened 2020-07-01
GitHub issue 318 (additional updates)

krowlandson:
This PR covers the issue raised in Issue #318, adding additional updates over those already covered in PR #330

Includes:

- Minor update to PowerShell command formatting
- Using version ARGs for PWSH and PSSA versions (currently set to latest but may want to pin to specific versions?)
- Added ARG for PWSH installation directory and removed version number (7)
- Added logic for additional PWSH and PSSA testing
- Treating PSSA as the Linter instead of PWSH in worker.sh
- Resolved minor error discovered when `"ACTIONS_RUNNER_DEBUG" == "true"` where `LINTER_ARRAY` specified value `powershell` instead of `pwsh` so `mapfile -t GET_VERSION_CMD < <("$LINTER" --version 2>&1)` returned an error.<br /><ul>`powershell --version` (/bin/sh: powershell: not found) --> `pwsh --version` (PowerShell 7.0.2)</ul>
- Added `${LINTER_NAME}` to debug output for binary check


Dockerfile
```diff
@@ -39,15 +39,18 @@ RUN apk add --no-cache \
 #########################################
 # Reference: https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-7
 # Slightly modified to always retrieve latest stable Powershell version
+# Specify PSScriptAnalyzer Repository and Version for stability
+ARG PWSH_VERSION='latest'
+ARG PSSA_VERSION='latest'
 RUN mkdir -p /opt/microsoft/powershell/7 \
```
alagoutte, line 45: remove the 7 ?

krowlandson: I'd assume that we only use PowerShell v7.x for now, but a good call as will protect against a future major release making this incorrect. I've set the PowerShell installation directory path using `ARG PWSH_DIRECTORY='/opt/microsoft/powershell'` so this can always be easily updated if desired.

I considered using the value from `PWSH_VERSION` in the directory path, but don't think this adds value, especially as targeting a specific PowerShell version requires a slightly modified path for the releases (note added in comments to explain).

Ideally I'd target a release tag 'v7' for PowerShell, but unfortunately the PowerShell team don't seem to maintain a release tag for major versions.

admiralAwkbar:
@krowlandson amazing.. pulling into a branch for full CI check

PR #341, opened 2020-07-01
Powershell fix

admiralAwkbar:
This PR covers the issue raised in Issue #318, adding additional updates over those already covered in PR #330

Includes:

- Minor update to PowerShell command formatting
- Using version ARGs for PWSH and PSSA versions (currently set to latest but may want to pin to specific versions?)
- Added ARG for PWSH installation directory and removed version number (7)
- Added logic for additional PWSH and PSSA testing
- Treating PSSA as the Linter instead of PWSH in worker.sh
- Resolved minor error discovered when `"ACTIONS_RUNNER_DEBUG" == "true"` where `LINTER_ARRAY` specified value `powershell` instead of `pwsh` so `mapfile -t GET_VERSION_CMD < <("$LINTER" --version 2>&1)` returned an error.<br /><ul>`powershell --version` (/bin/sh: powershell: not found) --> `pwsh --version` (PowerShell 7.0.2)</ul>
- Added `${LINTER_NAME}` to debug output for binary check

This closes #336


admiralAwkbar:
@krowlandson Amazing work! I just updated a bit to make it a function call and pull it out of the main linter function. A little cleaner, but does the exact same work you knocked out :) super cool...

PR #342, opened 2020-07-01
Build image locally for dev

admiralAwkbar:
Ref: https://github.com/github/super-linter/pull/274

Explanation:
https://github.com/github/super-linter/pull/273/files uses https://github.com/github/super-linter/blob/f1e46e8e76e65e3f96d06e780a9eebeeb701348e/.github/workflows/stack-linter.yml#L44 which pulls the latest release of the tool published to docker hub. This does not build the image locally and test using the image built. So you can lint the code prior to building the tool (which is good and necessary) but if you make a change to the code that would affect how the tool runs you would not be able to see validation that your changes work by pulling the latest release from docker hub.

The intention of this PR is to overcome the above challenge and negate the need for contributors with permissions to merge on this repo to pull changes in for CI to run. CI will just run when a user opens a PR.

_Originally posted by @nemchik in https://github.com/github/super-linter/pull/274#issuecomment-652031578_
This closes #332


admiralAwkbar:
We likely need to find a solution to the double jobs for `push` and `pull_request`

admiralAwkbar:
@MikeMcQuaid Thoughts on a fix to the double jobs?

admiralAwkbar:
I read somewhere you can set it up like an `or` statement, but can't find it anywhere...

.github/workflows/deploy-DEV.yml
```diff
@@ -20,6 +20,7 @@ on:
   push:
     branches-ignore:
       - 'master'
```
MikeMcQuaid, line 22: Needs to be `branches: master` if you want to avoid the double jobs.

PR #343, opened 2020-07-01
Were going to fix it all

### LINKED ISSUE #282: Support .rubocop_todo.yml
**Describe the bug**
We have the following entry in our `.rubocop.yml`:
```
inherit_from: .rubocop_todo.yml
```
The rubocop_todo file lists offences in current project to be addressed at a latter stage, as described in https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration

The linter fails if such file is included with
```
ERROR! Found errors in [rubocop] linter!
ERROR:[Configuration file not found: /action/lib/.automation/.rubocop_todo.yml]
``` 

**To Reproduce**
Add a `.rubocop_todo.yml` in the directory specified in `LINTER_RULES_PATH` and add `inherit_from: .rubocop_todo.yml` to the main config file.

**Expected behavior**
Rubocop TODO should be properly loaded by the main config file.

**Additional context**
I have `LINTER_RULES_PATH` set to root directory of the project and `.ruby-lint.yml -> .rubocop.yml` symlink.

### LINKED ISSUE #320: tflint fails on external source modules
**Describe the bug**
I am attempting to use `VALIDATE_TERRAFORM` (tflint) on a main.tf that contains a source module of `terraform-google-modules/scheduled-function/google`. The lint fails due to:

```
Error: `pubsub_scheduled_example` module is not found. Did you run `terraform init`?

  on main.tf line 59, in module "pubsub_scheduled_example":
  59: module "pubsub_scheduled_example" {]
```

I'm not aware of a way for super-linter to run terraform init during runtime. 

**To Reproduce**
1. Create a `main.tf` that contains a source module.
2. Execute super-linter with `VALIDATE_TERRAFORM` enabled.

**Expected behavior**
The super-linter should either run `terraform init` before executing `tflint` or support the `tflint` flag `--ignore-module`

### LINKED ISSUE #339: [doc] sample linter.yml doesn't actually lint anything
The sample `linter.yml` in https://github.com/github/super-linter/blob/master/README.md sets `VALIDATE_ANSIBLE: false`. According to the very same file setting a `VALIDATE_[LANGUAGE]` variable makes the other languages default to `false`:

> Note: All the `VALIDATE_[LANGUAGE]` variables behave in a specific way. If none of them are passed, then they all default to true. However if any one of the variables are set, we default to leaving any unset variable to false. This means that if you run the linter "out of the box", all languages will be checked. But if you wish to select specific linters, we give you full control to choose which linters are run, and won't run anything unexpected.

so the sample config file has all the linters disabled. I don't think this is intended.

### LINKED ISSUE #340: Work better for forks by updating example yml
At present the [Example Workflow file](https://github.com/github/super-linter/blob/master/README.md#example-connecting-github-action-workflow) works on pushes only and ignores master branch:

```
#############################
# Start the job on all push #
#############################
on:
  push:
    branches-ignore:
      - 'master'
```

If a repo is forked then this potentially leads to missing linting in the following circumstances:

- If the fork does not have Actions enabled (they are not enabled by default) and then they open a pull request.
- if the contributor works on the `master` branch in their fork and then opens a pull request.

As forks are really key to how GitHub works for a lot of projects, and in fact contributions for non-core contributors are probably ones you definitely want to lint this is not great IMHO.

What [we did](https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/.github/workflows/linter.yml), to improve this, was enable it on all `push` and `pull_requests` (including `master`) but put an exception for local pull requests to avoid running twice:

```
name: Lint Code Base
on:
  - push
  - pull_request
jobs:
  lint:
    name: Lint Code Base
    # Skip duplicate job for local branches
    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
    runs-on: ubuntu-latest
```

If the repo is forked, and GitHub actions are enabled then linting happens on `push` (good for contributor), and either way it is run again when the `pull_request` is opened (good for maintainers). So linting always happens (where GitHub Actions are enabled) and the results of the linting can be seen in the pull request issue for review before merging. Which is what we want.

The existing use case (linting from branches in the mean repo) still works as the `push` condition still triggers is used. The `if` statement prevents the action running twice when the Pull Request is opened or added to (as that is technically both a `push` and a `pull_request` as I understand it). It does put 2 checks in place in this case (one for the `push` which is run, and one for the `pull_request` which is skipped). A slight UX issue but the important thing is linting is run - and the results of the last `push` run on that branch are shown in the pull request (even if this happened before the pull requests is opened), which is what we want.

One thing to note with removing the ignoring of master branch, is that this means linting happens again when a pull request is merged into `master`. Again a slight inconvenience (or extra check depending how you look at it!) but think that is worth it to handle forks. Could possibly prevent this happening with a more convoluted `if` statement but, as I say, I think it is a good thing to run this extra check so in our repo we're happy to have this.

What's your thoughts on this? Is my understanding correct (I'm not an expert in GitHub Actions so could be wrong here!)? Should we update the example `linter.yml` file in the README.md to suggest this config instead? If so, I'm happy to submit a PR to change this but thought I'd open an issue first to discuss in case this is not wanted or I've got it totally wrong.

### LINKED ISSUE #349: openapi has double path
**Describe the bug**
when it tries to find openapi files, it doubles the path

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

admiralAwkbar:
This closes #339 
This closes #340
This closes #320
This closes #282 
This closes #349

### DIFF
--- a/.github/workflows/deploy-DEV.yml
+++ b/.github/workflows/deploy-DEV.yml
@@ -31,6 +31,8 @@ jobs:
     name: Deploy Docker Image - DEV
     # Set the agent to run on
     runs-on: ubuntu-latest
+    # Prevent duplicate run from happening when a forked push is committed
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     ##################
     # Load all steps #
     ##################
--- a/README.md
+++ b/README.md
@@ -125,7 +125,8 @@ jobs:
         uses: docker://github/super-linter:v3
         env:
           VALIDATE_ALL_CODEBASE: false
-          VALIDATE_ANSIBLE: false
+          DEFAULT_BRANCH: master
+
 ...
 ```
 
--- a/TEMPLATES/.tflint.hcl
+++ b/TEMPLATES/.tflint.hcl
@@ -1,6 +1,6 @@
 // https://github.com/terraform-linters/tflint/blob/master/docs/guides/config.md
 config {
-  module = true
+  module = false
   deep_check = false
   force = false
 
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -100,11 +100,11 @@ function BuildFileList()
     #####################
     # Get the CFN files #
     #####################
-    if [ "$FILE_TYPE" == "json" ] || [ "$FILE_TYPE" == "yml" ] || [ "$FILE_TYPE" == "yaml" ] && DetectCloudFormationFile "$FILE"; then
+    if [ "$FILE_TYPE" == "yml" ] || [ "$FILE_TYPE" == "yaml" ]; then
       ################################
       # Append the file to the array #
       ################################
-      FILE_ARRAY_CFN+=("$FILE")
+      FILE_ARRAY_YML+=("$FILE")
       ##########################################################
       # Set the READ_ONLY_CHANGE_FLAG since this could be exec #
       ##########################################################
@@ -118,11 +118,12 @@ function BuildFileList()
         # Append the file to the array #
         ################################
         FILE_ARRAY_CFN+=("$FILE")
+
+        ##########################################################
+        # Set the READ_ONLY_CHANGE_FLAG since this could be exec #
+        ##########################################################
+        READ_ONLY_CHANGE_FLAG=1
       fi
-      ##########################################################
-      # Set the READ_ONLY_CHANGE_FLAG since this could be exec #
-      ##########################################################
-      READ_ONLY_CHANGE_FLAG=1
     ######################
     # Get the JSON files #
     ######################
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -417,7 +417,7 @@ DetectOpenAPIFile()
   ###############################
   # Check the file for keywords #
   ###############################
-  grep -E '"openapi":|"swagger":|^openapi:|^swagger:' "$GITHUB_WORKSPACE/$FILE" > /dev/null
+  grep -E '"openapi":|"swagger":|^openapi:|^swagger:' "$FILE" > /dev/null
 
   #######################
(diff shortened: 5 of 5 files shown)


zkoppert:
This looks concerning: [L66](https://github.com/github/super-linter/pull/343/checks?check_run_id=827022378#step:4:66). Not sure why your changes would have caused this though. 🤔

zkoppert:
going to try rerunning the action

admiralAwkbar:
@zkoppert might need some help figuring out the error;
```
You are not currently on a branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>
```

admiralAwkbar:
@zkoppert I think this works...

admiralAwkbar:
@nemchik any chance you can help validate all is working as expected?

What I have done:
- We used to copy all config files to a config folder when we would run the action.
- This works except if you reference other configs from your main location, then all is toast...
- So the solution was to find the files and then make the reference from their location instead
  - This resulted in some fun variables...

nemchik:
I'll have a look. I've got a branch running CI with what I think might be the issue now. If that's not it I'll keep looking later, otherwise I'll PR to this branch.

admiralAwkbar:
found a fun bug with open API where the path was being doubled and failing... deep in the weeds...

nemchik:
https://github.com/github/super-linter/pull/347 may help with the "not on a branch" issue. I think you've already resolved it another way, but this would double down and just make sure that all `git ...` commands are run within the `$GITHUB_WORKSPACE` directory.

I assume this was the issue, but I am not positive that is what was going on.

admiralAwkbar:
I think were gold on this now...

PR #351, opened 2020-07-01
shfmt files

nemchik:
I ran:
```
docker run --rm -v ${PWD}:${PWD} mvdan/shfmt -s -i 2 -ci -sr -w ${PWD}/.automation/*.sh
docker run --rm -v ${PWD}:${PWD} mvdan/shfmt -s -i 2 -ci -sr -w ${PWD}/.vscode/*.sh
docker run --rm -v ${PWD}:${PWD} mvdan/shfmt -s -i 2 -ci -sr -w ${PWD}/lib/*.sh
```
(does not recurse through subfolders, so the intentionally bad shell files for test purposes are untouched)

This is just a code formatter. It can be used as a linter as well (I've mentioned it in some issue discussion). Most of the changes are whitespace changes. Generally all of the changes are meant to improve readability and ensure the shell (bash) does not incorrectly interpret anything.

If there are any specific questions feel free to comment inline.

I won't be too hurt if this doesn't merge. It took seconds to create this branch, and a valid reason for NOT merging it would be that this style of formatting would not be enforced in the future (unless it were to be mandated via pre-commit hooks or GHA checks). But it can still carry value as a one-off.


lib/buildFileList.sh
```diff
@@ -26,7 +25,10 @@ function BuildFileList()
   #################################################################################
   # Switch codebase back to the default branch to get a list of all files changed #
   #################################################################################
-  SWITCH_CMD=$(git -C "$GITHUB_WORKSPACE" pull --quiet; git -C "$GITHUB_WORKSPACE" checkout "$DEFAULT_BRANCH" 2>&1)
+  SWITCH_CMD=$(
```
admiralAwkbar, line 28: Very cool!

lib/validation.sh
```diff
@@ -77,40 +76,40 @@ function GetValidationInfo()
...
-        -n "$VALIDATE_GO" || \
-        -n "$VALIDATE_TERRAFORM" || \
-        -n "$VALIDATE_POWERSHELL" || \
-        -n "$VALIDATE_CSS" || \
-        -n "$VALIDATE_ENV" || \
-        -n "$VALIDATE_CLOJURE" || \
-        -n "$VALIDATE_PROTOBUF" || \
-        -n "$VALIDATE_OPENAPI" || \
-        -n "$VALIDATE_KOTLIN" ]]; then
+  if [[ -n $VALIDATE_YAML || -n \
```
admiralAwkbar, line 79: This is an odd way of changing it.. haven't seen in this format...

lib/worker.sh
```diff
@@ -273,12 +275,12 @@ function TestCodebase()
...
     ############################
     # Example: markdown_good_1.md -> good
-    FILE_STATUS=$(echo "$FILE_NAME" |cut -f2 -d'_')
+    FILE_STATUS=$(echo "$FILE_NAME" | cut -f2 -d'_')
 
     #########################################################
     # If not found, assume it should be linted successfully #
     #########################################################
-    if [ -z "$FILE_STATUS" ] || [[ "$FILE" == *"README"* ]]; then
+    if [ -z "$FILE_STATUS" ] || [[ $FILE == *"README"* ]]; then
```
admiralAwkbar, line 283: It's odd how some of these need to be quoted and some do not... the rules on this from spellcheck are interesting...

lib/validation.sh
nemchik: The logic here is that a line should not start with an operator (`-` or `|` etc).

lib/worker.sh
nemchik: When using bash specific conditional braces (`[[ double ]]` is bash specific, `[ single ]` is sh but works in bash) quotes on the left side of the operator are unnecessary, but quotes on the right side of the operator are required. https://stackoverflow.com/a/15581159/1384186

nemchik:
> I really like the cleanup of the code! It seems to be all spacing fixes and cleanup of some formatting. The only thing is just the de-quoting of some of the vars... I understand that you should quote the vars to prevent parameter expansion issues, but it's not complaining about some of them being unquoted...

See https://github.com/github/super-linter/pull/351#discussion_r448993344

nemchik:
P.S. I use https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format and install shfmt from https://github.com/mvdan/sh either natively on linux or using https://scoop.sh/ on windows. I would use chocolatey but there is no package for shfmt in the chocolatey repositories.

lib/validation.sh
admiralAwkbar: dope

PR #354, opened 2020-07-02
Auto draft releases

### LINKED ISSUE #203: Make writing changelogs easier
Running through git logs or closed pull requests is not ideal as its manual.

Use something like [release drafter](https://github.com/apps/release-drafter) to help draft the releases so we can just review for accuracy.

- https://github.com/marketplace/actions/automatic-changelog-generator
- https://github.com/CookPete/auto-changelog
- https://github.com/github-changelog-generator/github-changelog-generator

**Additional context**
We should discuss pros and cons of the alternatives vs the suggested release drafter.

zkoppert:
fixes #203 

- Adds the release drafter action, configured to attempt to guess the next version number
- Updates documentation to request labels on all pull requests before merge

Preview of what a draft release will look like:
**Note that where it says @zkoppert, it will put the author of the change
Screen Shot 2020-07-01 at 9 08 51 PM

### DIFF
--- a/.github/pull_request-template.md
+++ b/.github/pull_request-template.md
@@ -1,0 +1,13 @@
+<!-- Link to issue if there is one -->
+Fixes #
+
+<!-- Describe what the changes are -->
+## Proposed Changes
+
+  -
+  -
+  -
+
+## Readiness Checklist
+ - [ ] Label as `breaking` if this is a large fundamental change
+ - [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`
--- a/.github/pull_request-template.md
+++ b/.github/pull_request-template.md
@@ -1,4 +1,4 @@
-<!-- Please ensure your PR title is breif and descriptive for a good changelog entry -->
+<!-- Please ensure your PR title is brief and descriptive for a good changelog entry -->
 <!-- Link to issue if there is one -->
 <!-- markdownlint-disable -->
 Fixes #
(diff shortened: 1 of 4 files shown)


IAmHughes:
@zkoppert I'm not super familiar with this action but it seems cool. Does this create draft releases only, or as configured is it going to create actual releases? If the latter, this will break other automation we have that is triggered off of a release such as the deployments and version tag shuffling.

Also, every push to master may be a little too much, no? We do a doc update (README for example) often currently and I wouldn't create releases for those. If they are drafts we can just delete them, but that sounds noisy. Open to thoughts and discussions for sure!

zkoppert:
This action will create a draft release or update the existing draft. It never creates an actual one reserving that for humans. 😄  Luckily draft releases do not create tags until they become official so that helps too.

nemchik:
Questions

* I gather Draft Releases are not Releases (or Pre Releases), so their purpose is to stage a Release (or Pre Release) and a human can then convert it from a Draft Release to a Release (or Pre Release) presumably "with the click of a button"?
* If a Draft is created, and new changes are merged without a new Release being made, this will update the Draft with information about the changes merged after the draft was created? (so it always includes all changes since the last release)

.github/pull_request-template.md
```diff
@@ -0,0 +1,16 @@
+<!-- Please ensure your PR title is breif and descriptive for a good changelog entry -->
```
yg, line 1: Typo

```suggestion
<!-- Please ensure your PR title is brief and descriptive for a good changelog entry -->
```

zkoppert:
@nemchik yes to both your questions

Issue #356, opened 2020-07-02
Super linter not picking up user specified lint config files


vishad-slice:
**Describe the bug**
In the latest version v3.2.0, the user specified lint config file locations are not working. It is always using the default files.
**To Reproduce**
1. Set these envs:
    JAVASCRIPT_ES_CONFIG_FILE: .eslintrc.json
    LINTER_RULES_PATH: /path/to/config/directory
**Expected behavior**
User specified lint config file should be used

**Screenshots**
<img width="1383" alt="Screenshot 2020-07-02 at 6 11 08 PM" src="https://user-images.githubusercontent.com/60728725/86360213-ab63a180-bc8f-11ea-92c1-e786ffe64d49.png">

**Additional context**
Working as expected in v3.1.1

techotron:
I've noticed the same issue this morning. I've added a comment to this issue which at the time looked similar: https://github.com/github/super-linter/issues/234#issuecomment-652903651

techotron:
I think it's a scoping thing with the `declare` statement. If I change [this line](https://github.com/github/super-linter/blob/9d5d792af8d1c365b2bf34535908c6df92921704/lib/linter.sh#L320) to:

`declare -g "${LANGUAGE_LINTER_RULES}=$GITHUB_WORKSPACE/$LINTER_RULES_PATH/${!LANGUAGE_FILE_NAME}"` 

then the custom linter works for me

Issue #234, opened 2020-06-22
Super linter does not respect .ruby-lint.yml


nhattan:
**Describe the bug**
I followed [this doc](https://github.com/github/super-linter/blob/master/docs/disabling-linters.md#ruby) creating `.github/linters/.ruby-lint.yml` with the following content but Super linter is still checking the excluded directories

```yml
AllCops:
  EnabledByDefault: true
  Exclude:
    - 'bin/**/*'
    - 'config/**/*'
    - 'config.ru'
    - 'db/**/*'
    - 'node_modules/**/*'
    - 'spec/rails_helper.rb'
    - 'spec/spec_helper.rb'
    - 'tmp/**/*'
```

My `linter.yml`
```yml
name: Lint Code Base

on:
  push:
    branches: [ master, develop ]
  pull_request:
    branches: [ master, develop ]

jobs:
  build:
    name: Lint Code Base
    runs-on: ubuntu-latest

    steps:
      - name: Checkout Code
        uses: actions/checkout@v2

      - name: Lint Code Base
        uses: docker://github/super-linter:v2.1.1
        env:
          VALIDATE_ALL_CODEBASE: false
          VALIDATE_YAML: true
          VALIDATE_JSON: true
          VALIDATE_MD: true
          VALIDATE_BASH: true
          VALIDATE_RUBY: true
          VALIDATE_JAVASCRIPT_ES: false
          VALIDATE_DOCKER: true
          VALIDATE_CSS: true
          VALIDATE_ENV: true
```

**Expected behavior**
Super linter should skip checking the excluded directories.

**Screenshots**
<img width="1437" alt="Screen Shot 2020-06-22 at 08 43 36" src="https://user-images.githubusercontent.com/4226759/85241012-82a80480-b464-11ea-9c18-64c3c9835b65.png">

**Additional context**
I ran `rubocop -c .github/linters/.ruby-lint.yml` on my local and it's fine, no offenses detected.

GaboFDC:
I think this is because the script logics, runs rubocop against every single file? not sure what would be the way to fix this 🤔

nhattan:
@GaboFDC script logics?

dsayling:
I'm having the same issue on python. I've added a new exclusion to my own .python-lint file. The output states the local config file is being loaded - however the pylint issue is still not ignored.

annarailton:
I'm having the same issue with a `dockerlintrc` file being ignored.

nhattan:
[...]

PR #359, opened 2020-07-02
Use a global scope when setting the LANGAUGE_LINTER_RULES variable

techotron:
This PR changes the scope for the $LANGUAGE_LINTER_RULES variable so that it is seen when executing the `LintCodebase` function. Without this, the variable isn't changed outside of the `GetLinterRules` scope and therefore the default rules are applied.

Addresses the following issue: https://github.com/github/super-linter/issues/356
Possibly addresses this issue: https://github.com/github/super-linter/issues/234

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -317,7 +317,7 @@ GetLinterRules()
     ########################################
     # Update the path to the file location #
     ########################################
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@techotron Thank you!

Issue #143, opened 2020-06-18
Invoking golangci-lint run on individual files breaks linters in packages.


RussellRollins:
**Describe the bug**
When golangci-lint is invoked against a single file, as it is here:

https://github.com/github/super-linter/blob/1e08ff6363cb45c5714b1c30039d896972287173/lib/linter.sh#L1939

This fails to properly lint a package if types are defined in another file.

**To Reproduce**
1. Create two Go files, main.go and foo.go.
2. Define `Foo` in foo.go.
3. Run `golangci-lint run` in that package, it lints fine.
4. Run the super-linter against that package, it fails to lint with: `ERROR:[level=warning msg="[runner] Can't run linter unused: buildssa: analysis skipped: errors in package: [/github/workspace/main.go:8:8: undeclared name: Foo]"` 

See an example run here: https://github.com/RussellRollins/super-lint-error/runs/785695971

**Expected behavior**
This passes linting without failures.

**Screenshots**
![Screen Shot 2020-06-18 at 3 34 24 PM](https://user-images.githubusercontent.com/3237030/85064207-35344a80-b179-11ea-9811-81972b08250a.png)

**Additional context**
I can see how the current solution is used to enable the "lint only changed files" feature, but this will fall down for relatively simple Go setups.

bznein:
We encountered the same issue and this makes SuperLinter with Golang basically unusable.
An option to run the linter on whole directories when `VALIDATE_ALL_CODEBASE ` is True would be extremely helpful

bznein:
Hi @RussellRollins, I created a PR #370 which aims at fixing this issue. The owners of the repo want to gather some more feedback from other GO developer. If you have some time to spare can you have a look at it and leave your opinion there? Thanks!

iggy:
Slow your roll bot. This is still valid.

Sorry for the noise. Trying to keep this from getting closed.

pwmcintyre:
still valid; super-linter is totally broken for golang :(

jtbonhomme:
Still valid!

WendelHime:
still valid

GaboFDC:
Related to #888

karolistamutis:
What would be the workaround besides disabling Super-Linter completely for go? Having the same issue.

masih:
Not stale

WendelHime:
not stale

bznein:
not stale

RussellRollins:
[...]

PR #370, opened 2020-07-03
Fix GO linting

bznein:
As pointed out in #143, running super-linter on Golang code breaks as soon as packages are split in more than one files, which can happen even with moderately simple projects. The idea in this PR is to fix this issue by running `golangci-lint` on whole directories instead of singular files, when `VALIDATE_ALL_CODEBASE == false` and recursively on the whole repo otherwise.

Feedback is greatly appreciated on all points, but especially on the following:

- The array of go files to analyze will potentially contain multiple repeated entries. To get rid of them I am using the following commnad: `  mapfile -t FILE_ARRAY_GO < <(printf '%s\n' "${FILE_ARRAY_GO[@]}" | sort -u)` Any suggestion on better approaches from more experienced bash developers is highly appreciated.
- With this solution, the number of errors reported by SuperLinter will be always heavily underestimated. If running against all codebase, it will at most be 1; otherwise, it will be at most 1 for every directory analysed. Please note that, however, at the moment the number of errors is already not exactly the number of errors found by the linter but only the number of files with one or more errors (example [here](https://github.com/bznein/AoC2019-Go/runs/834384999?check_suite_focus=true) when the number of errors is equal to 1 while there are 3 errors in a single file)

A Docker image built from the code in this PR is available at https://hub.docker.com/r/bznein/superlinter

### DIFF
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -281,7 +281,7 @@ function BuildFileList() {
       ################################
       # Append the file to the array #
       ################################
-      FILE_ARRAY_GO+=("$FILE")
+        FILE_ARRAY_GO+=("$(dirname "${FILE}")" )
       ##########################################################
       # Set the READ_ONLY_CHANGE_FLAG since this could be exec #
       ##########################################################
@@ -414,6 +414,9 @@ function BuildFileList() {
   done
 
   echo ${READ_ONLY_CHANGE_FLAG} > /dev/null 2>&1 || true # Workaround SC2034
+
+  FILE_ARRAY_GO=($(echo "${FILE_ARRAY_GO[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' '))
+
 
   #########################################
   # Need to switch back to branch of code #
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -281,7 +281,7 @@ function BuildFileList() {
       ################################
       # Append the file to the array #
       ################################
-        FILE_ARRAY_GO+=("$(dirname "${FILE}")" )
+      FILE_ARRAY_GO+=("$(dirname "${FILE}")" )
       ##########################################################
       # Set the READ_ONLY_CHANGE_FLAG since this could be exec #
       ##########################################################
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -415,8 +415,7 @@ function BuildFileList() {
 
   echo ${READ_ONLY_CHANGE_FLAG} > /dev/null 2>&1 || true # Workaround SC2034
 
-  FILE_ARRAY_GO=($(echo "${FILE_ARRAY_GO[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' '))
-
+  mapfile -t FILE_ARRAY_GO < <(printf '%s\n' "${FILE_ARRAY_GO[@]}" | sort -u)
 
   #########################################
   # Need to switch back to branch of code #
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -955,8 +955,12 @@ if [ "$VALIDATE_GO" == "true" ]; then
   #########################
   # Lint the golang files #
   #########################
-  # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILE_ARRAY"
-  LintCodebase "GO" "golangci-lint" "golangci-lint run -c $GO_LINTER_RULES" ".*\.\(go\)\$" "${FILE_ARRAY_GO[@]}"
+    # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILE_ARRAY"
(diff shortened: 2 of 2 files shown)


admiralAwkbar:
@bznein this is a fairly elegant way of solving the issue. I like the sort removes duplicates, and shouldn't be much heavier for anyone unless they are going to really be working on Go files... So from my understanding, this would work... I would love some others with more Go experience to chime in and make sure the community agrees with the approach. Can you ping any open devs from the Go issues and see if they agree?

LouisPlisso:
That looks right to me

RussellRollins:
This definitely seems like an improvement to me, and appears to work as intended when `VALIDATE_ALL_CODEBASE` is true.

However, I think the solution for validating diffs isn't quite right. In this code, if any file in a package/directory changes, the whole package/directory will be linted. I think this breaks the assumptions of how the superlinter is meant to operate.

`golangci-lint` has a few flags that might be used to fix that issue directly. Instead of using a list of changed files, you could generate a git patch in `buildFileList.sh`, and then use `--new-from-patch`. Or you could invert the current list of files (so find all files that _didn't_ change) and try to build a regular expression for use with the `--skip-files` flag.

As is, this seems like a good change to me, but with some more work might better fit the goals of the superlinter.

admiralAwkbar:
@bznein let me know when you have those updates so i can go over and get this worked in :)

bznein:
Thanks for the comments @RussellRollins!

I'll work on it this weekend hopefully :)

sayboras:
@bznein Hope you don't mind if I chime in

Just quickly test your changes [here](https://github.com/sayboras/cilium/runs/869113449?check_suite_focus=true). Seems like even there is failure, github action still returns as success. Please refer to the below screen shot of local run.

image

lib/linter.sh
```diff
@@ -956,7 +956,11 @@ if [ "$VALIDATE_GO" == "true" ]; then
   # Lint the golang files #
   #########################
   # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILE_ARRAY"
-  LintCodebase "GO" "golangci-lint" "golangci-lint run -c $GO_LINTER_RULES" ".*\.\(go\)\$" "${FILE_ARRAY_GO[@]}"
+  if [ "$VALIDATE_ALL_CODEBASE" == "true" ]; then
+      LintCodebase "GO" "golangci-lint" "golangci-lint run -c $GO_LINTER_RULES" ".*\.\(go\)\$" "$GITHUB_WORKSPACE/..."
+  else
+      LintCodebase "GO" "golangci-lint" "golangci-lint run -c $GO_LINTER_RULES" ".*\.\(go\)\$" "${FILE_ARRAY_GO[@]}"
```
sayboras, line 962: I assume that you are trying to derive the list of dirs/files, so that we can still use generic _LintCodebase_ function. This might work for other linters, but might not work for golang in general. 

Consider that we can have few go versions with go mod, dep, vendoring, etc. Just curious if we can just run _golangci-lint run -c <config file> ./..._ on project root directory (or in the diretory having go.mod, dep, vendor, etc).

lib/buildFileList.sh
```diff
@@ -415,6 +415,8 @@ function BuildFileList() {
 
   echo ${READ_ONLY_CHANGE_FLAG} > /dev/null 2>&1 || true # Workaround SC2034
 
+  mapfile -t FILE_ARRAY_GO < <(printf '%s\n' "${FILE_ARRAY_GO[@]}" | sort -u)
```
nhutran1201, line 418:
```suggestion
  mapfile -t FILE_ARRAY_GO < <(printf '%s\n' "${FILE_ARRAY_GO[@]}" | sort -u)
```

lib/buildFileList.sh
```diff
@@ -281,7 +281,7 @@ function BuildFileList() {
       ################################
       # Append the file to the array #
       ################################
-      FILE_ARRAY_GO+=("$FILE")
+      FILE_ARRAY_GO+=("$(dirname "${FILE}")" )
```
nhutran1201, line 284:
```suggestion
      FILE_ARRAY_GO+=("$(dirname "${FILE}")" )
```

lib/linter.sh
```diff
@@ -956,7 +956,11 @@ if [ "$VALIDATE_GO" == "true" ]; then
   # Lint the golang files #
   #########################
   # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILE_ARRAY"
-  LintCodebase "GO" "golangci-lint" "golangci-lint run -c $GO_LINTER_RULES" ".*\.\(go\)\$" "${FILE_ARRAY_GO[@]}"
+  if [ "$VALIDATE_ALL_CODEBASE" == "true" ]; then
```
nhutran1201, line 959:
```suggestion
  if [ "$VALIDATE_ALL_CODEBASE" == "true" ]; then
```

migolfi:
Hey, just wanting to give a friendly bump, can this be reopened? 
While working in Go I'm unable to split files as usual.

bznein:
Hi @migolfi,

I'll try to find some time in the next couple of weeks to look at the feedback and push a new version!

PR #372, opened 2020-07-03
Use dependabot for linters dependencies

### LINKED ISSUE #134: Upgrade Rubocop to Latest
We were looking to move from running Rubocop locally to super-linter but super-linter uses a version from about 1 year ago (0.74). https://github.com/rubocop-hq/rubocop/releases has 0.85.1 as the latest right now.

Upgrade to a more recent version of rubocop, or enable this to be configured by user.

Thanks!

### LINKED ISSUE #201: Keep dependencies up to date
How can we automate this project staying up to date with the latest stable version of the linters it references?

Minimally cool: Alert or issue created every time we need to update a version of a linter
Maximally cool: Automatically open a PR to rev the version once a stable release from a linter is available

GaboFDC:
This fixes #201 
I did a cleanup as good as I can, this is still a draft but only missing part is adding dependabot, that after #367 will only need to add npm and pyenv to the config, so I'll like feedback on the way this was done and any other comments.

About ruby, we need to fix #134 before we can start using the Gemfile properly as I stated in a comment there.

I also used what @nemchik suggested, of copy binaries from maintained docker images, for more clarity, and ease of "install"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,10 +28,6 @@ ARG PSSA_VERSION='latest'
 # arm-ttk
 ARG ARM_TTK_URI='https://github.com/Azure/arm-ttk.git'
 ARG ARM_TTK_DIRECTORY='/opt/microsoft/arm-ttk'
-# clj-kondo
-ARG CLJ_KONDO_VERSION='2020.06.21'
-# Go Linter
-ARG GO_VERSION='v1.27.0'
 # Raku Linter
 ARG RAKU_VER="2020.06"
 ARG RAKU_INSTALL_PATH=/usr
@@ -41,19 +37,49 @@ ARG RAKUBREW_HOME=/tmp/rakubrew
 # Run APK installs #
 ####################
 RUN apk add --no-cache \
-    bash git git-lfs musl-dev curl gcc jq file\
+    bash curl \
+    gcc go \
+    icu-libs jq libxml2-utils make musl-dev \
     npm nodejs \
-    libxml2-utils perl \
-    ruby ruby-dev ruby-bundler ruby-rdoc make \
-    py3-setuptools ansible-lint \
-    go \
     openjdk8-jre \
-    php7 \
-    ca-certificates less ncurses-terminfo-base \
-    krb5-libs libgcc libintl libssl1.1 libstdc++ \
-    tzdata userspace-rcu zlib icu-libs lttng-ust \
-    libffi-dev openssl-dev
+    perl php7 \
+    py3-setuptools ansible-lint \
+    ruby ruby-dev ruby-bundler ruby-rdoc
 
+########################################
+# Copy dependencies files to container #
+########################################
+COPY dependencies/* /
+
+################################
+# Installs python dependencies #
+################################
+RUN pip3 install --no-cache-dir pipenv
+RUN pipenv install --system
+
+####################
+# Run NPM Installs #
+####################
+RUN npm config set package-lock false \
+    && npm config set loglevel error \
+    && npm --no-cache install
+
+#############################
+# Add node packages to path #
+#############################
+ENV PATH="/node_modules/.bin:${PATH}"
+
+####################
+# Run GEM installs #
+####################
+RUN gem install rubocop:0.74.0 rubocop-rails rubocop-github:0.13.0
+
+# Need to fix the version as it installs 'rubocop:0.85.1' as a dep, and forces the default
+# We then need to promote the correct version, uninstall, and fix deps
+RUN sh -c 'INCORRECT_VERSION=$(gem list rhc -e rubocop | grep rubocop | awk "{print $2}" | cut -d"(" -f2 | cut -d"," -f1); \
+  gem install --default rubocop:0.74.0; \
+  yes | gem uninstall rubocop:$INCORRECT_VERSION -a -x -I; \
+  gem install rubocop:0.74.0'
 
 #########################################
 # Install Powershell + PSScriptAnalyzer #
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -22,83 +22,37 @@ LABEL com.github.actions.name="GitHub Super-Linter" \
...
-    py3-setuptools ansible-lint \
-    go \
     openjdk8-jre \
-    php7 \
-    ca-certificates less ncurses-terminfo-base \
-    krb5-libs libgcc libintl libssl1.1 libstdc++ \
-    tzdata userspace-rcu zlib icu-libs lttng-ust
+    perl php7 \
+    py3-setuptools ansible-lint \
+    ruby ruby-dev ruby-bundler ruby-rdoc
```
nemchik, line 32: Just for the sake of future sanity and cleanliness, can you put one per line here?
ex:
```
bash \
curl \
```
etc

Dockerfile
```diff
@@ -112,46 +66,54 @@ RUN sh -c 'INCORRECT_VERSION=$(gem list rhc -e rubocop | grep rubocop | awk "{pr
...
+    | tar -xzC ${PWSH_DIRECTORY} \
+    && ln -s ${PWSH_DIRECTORY}/pwsh /usr/bin/pwsh -f \
+    && pwsh -c 'Install-Module -Name PSScriptAnalyzer -RequiredVersion ${PSSA_VERSION} -Scope AllUsers -Force'
+
 ######################
 # Install shellcheck #
 ######################
-RUN wget -qO- "https://github.com/koalaman/shellcheck/releases/download/stable/shellcheck-stable.linux.x86_64.tar.xz" | tar -xJv \
-    && mv "shellcheck-stable/shellcheck" /usr/bin/
+COPY --from=koalaman/shellcheck:stable /bin/shellcheck /usr/bin/
```
nemchik, line 91: Use `v0.7.1` (the latest from https://hub.docker.com/r/koalaman/shellcheck/tags). Ideally we want to be specific with the version used and let dependabot update this dockerfile with new versions.

dependencies/Pipfile
```diff
@@ -0,0 +1,16 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
```
nemchik, line 8: I do not have enough experience with dependabot to confirm it should keep versions in this file up to date, but ideally in any place where dependabot is capable of updating versions for us we should specify exact version numbers. This makes it easier for users in the future to track down the commit associated with a release and see specific versions of dependencies used.

nemchik:
The only thing I think left in this PR to address is https://github.com/github/super-linter/pull/372#discussion_r449727037

Everything else here is looking really good!

nemchik:
https://github.com/github/super-linter/pull/367 merged and seems to have created a conflict. Please rebase.

admiralAwkbar:
@GaboFDC this looks good to me! I found a few issues and cleaned them up

GaboFDC:
Hye @admiralAwkbar thanks for the help. Seems like the rebase was not clean enough, I see duplicate npm install. And the packages install was reverted, I had cleaned it to the minimus as lots of things were not needed anymore.
I'll re-rebase and force push if that's ok with you?

Also, I'll test about https://github.com/github/super-linter/pull/372#discussion_r449727037 to be sure dependabot works as expected.

admiralAwkbar:
@GaboFDC yeah that rebase went badly, my bad. Please make any adjustments and push to the branch so we can get it all validated and up to date

dependencies/Pipfile
GaboFDC: From some tests, this is fine, as the locked version is on the .lock file. 

So here we just say that there is no restriction in the version (like a minimum), and dependabot will correcly update the lock file when new releases are available

nemchik: If dependabot cannot update these then leaving `*` is fine, but if it can then my opinion is we should put the specific versions in here. Another benefit is that new versions for any versioned dependency will have to be approved by a human via a PR like this one https://github.com/github/super-linter/pull/381 which sounds like more human interaction but actually ends up meaning a build is triggered for each new version of the dependency, and so we end up with a lot more testing to keep things running smooth. I think dependabot might even be able to auto merge PRs with passing builds if it becomes a lot to manage, but then the ability to intentionally hold back a specific version is lost. Another reason for doing it in this file and not relying on the lock file is the lock file is a lot less human readable to basic end users. In this case it's pretty small, but in npm for example it is nearly pointless for a human to read the lock file with how complex it is.

GaboFDC: Just to clarify, using the `*` does not forbid dependabot from working, it works just fine, see my test [here](https://github.com/GaboFDC/super-linter/pull/10/files)
But you still prefer to keep the specific version on the normal file, right? If so, I'll put it with minimum the latest version we are using at the moment, does that seems fine?

nemchik: Awesome to see dependabot working in your testing! My only reason for still considering it being in this file is user readability. I don't think it's a sword worth dying on, so if keeping `*` is easier i'm fine with that.

nemchik:
GitHub Actions seemed broken earlier today. If you run the following it should re-push your latest commit and trigger new checks:
```
git commit --amend --no-edit
git push -f
```

admiralAwkbar:
@GaboFDC Amazing work out there... it needed some cleanup :)

admiralAwkbar:
@GaboFDC did you have any more work on this branch or do you feel like you got to where you wanted it to be?

GaboFDC:
@admiralAwkbar I think this is ready

PR #377, opened 2020-07-06
Add basic Raku support

nbrown:
Adding support for the language `Raku`
- https://rakubrew.org/


admiralAwkbar:
@nbrown Amazing work.

One fix?
Can you set it to install in `/usr/bin/` instead of `root`

Other than that it seems great!

admiralAwkbar:
Can you also use the installer from https://rakubrew.org/perl/rakubrew instead of the other one that says deprecated?

nbrown:
I think I fixed both of your comments. Please let me know if there's anything else you'd like me to look at.

admiralAwkbar:
@nbrown amazing work! thank you so much for the contributions!

Issue #120, opened 2020-06-18
Dart support


shyndman:
Dart support would be useful. 

The official (and only I believe) linter is https://github.com/dart-lang/linter
Rules: https://dart-lang.github.io/linter/lints/

mit-mit:
Hi @cvega are you working on this? If so, I'm on the Dart team, and I'm happy to help with any questions you might have.

cvega:
Hello @shyndman & @mit-mit 

I appreciate the suggestion and the offer to provide feedback. I have the basis for adding dart or more specifically `dartanalyzer` to super-linter. The only big concern I had was the size of the SDK in relation to the container (not a huge deal, but less is more). I was wondering if you could provide any feedback on the dart binary? Ideally I'd like to move the dart binary and `dartanalyzer` into `/usr/bin` and call it day but I'm not certain that would suffice. I'm in the process of testing that as time permits, but if you have the details I'm all ears. I will also tag you both in the PR which I plan on getting out later this evening. Thanks in advance.

mit-mit:
@devoncarew WDYT about embedding just a few binaries from the SDK; is that likely to cause issues longer-term? And if we go that route, would we rather rely on the files we need for `dart analyze`?

devoncarew:
Hi @cvega - exciting to get dart support here! Some thoughts:

- for longevity of the integration, I think we should prefer using `dart analyze` to lint rather than `dartanalyzer`. `dart analyze` has some (important) bug fixes that `dartanalyzer` likely won't ever get, and in the medium term we'll remove dartanalyzer completely (in preference to `dart analyze`)
- I'd be a little leery of just grabbing a few binaries from the SDK - I'm not sure how well that would work, and it seems likely to break as we evolve the SDK. I'd be a lot less concerned about removing files from the sdk - that seems a lot less fragile. Deleting files like bin/model/lexeme/model.tflite, bin/snapshots/dart2js.dart.snapshot, ... will cut the size of the sdk in half.

cvega:
@mit-mit @devoncarew 

This is very helpful. 

- I like longevity too. Done.
- I've discussed sizing of the SDK. Not a huge deal. For whats it worth, It seems to work fine with just the binary (very limited testing) but I rather not copy pieces of the SDK or give Dart users/developers a subpar experience because something they expect is missing. I am grabbing the entire SDK. For the time being, I'll add the entire package and we can trim it down once I produce the PR.

Again, thank you both for your feedback. Much appreciated.

[...]

PR #379, opened 2020-07-06
add dart support

cvega:
Initial commit for adding dart support with default pedantic rules. I'm using `dartanalyzer` however @mit-mit and @devoncarew mentioned the ability to `dart analyze` in #120 . Would be beneficial to get feedback from the dart community regarding pedantic/default rules the differences between the two or how to implement the later using dart best practices.

Reference `dart analyze`:
https://github.com/dart-lang/sdk/tree/master/pkg/dartdev


.automation/test/dart/README.md
```diff
@@ -0,0 +1,13 @@
+# Dart Test Cases
```
devoncarew, line 1: Is it normal for these super-linter linters to put the test cases in an `.automation` folder? If this was a normal dart project, they might go in a folder like `test/sample` (and then likely have the negative samples be excluded from analysis via an entry in the analysis_options.yaml file).

devoncarew: Ah, nevermind, I see lots of other languages listed in `.automation/test/`

.automation/test/dart/dart_bad_1.dart
```diff
@@ -0,0 +1,6 @@
+// empty_constructor_bodies bad {}
+
+class Point {
```
devoncarew, line 3: You may want the `dartfmt` file file (and the following one).

Dockerfile
```diff
@@ -181,6 +181,22 @@ RUN curl -sLO https://github.com/borkdude/clj-kondo/releases/download/v${CLJ_KON
...
+# Install dart-sdk #
+####################
+# install alpine-pkg-glibc (glibc compatibility layer package for Alpine Linux)
+ARG GLIBC_VERSION='2.31-r0'
+RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub
+RUN wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-${GLIBC_VERSION}.apk
+RUN apk add glibc-${GLIBC_VERSION}.apk
+
+# install dart sdk
+ARG DART_VERSION='2.8.4'
```
devoncarew, line 194: You may want to prefer to pull the latest stable version, instead of using a fixed version.

README.md
```diff
@@ -39,6 +39,7 @@ Developers on **GitHub** can call the **GitHub Action** to lint their code base
 | **CSS**          | [stylelint](https://stylelint.io/)                                       |
 | **Clojure**      | [clj-kondo](https://github.com/borkdude/clj-kondo)                       |
 | **CoffeeScript** | [coffeelint](https://coffeelint.github.io/)                              |
+| **Dart**         | [dartanalyzer](https://dart.dev/tools/dartanalyzer)                      |
```
devoncarew, line 42: Instead of pointing to the `dartanalyzer` tool, perhaps just point to https://dart.dev?

If we do want to point to a page related to static analysis, perhaps this one:
https://dart.dev/guides/language/analysis-options

docs/disabling-linters.md
```diff
@@ -660,6 +660,35 @@ import package.b.*
 
 --------------------------------------------------------------------------------
 
+## Dart
+- [dartanalyzer](https://dart.dev/tools/dartanalyzer)
+
+### dartanalyzer standard Config file
```
devoncarew, line 666: Hmm, I don't pretend to know how the super-linter project works, but for configuring dart's statis analysis, we'll generally want to use the `analysis-options.yaml` file in the root of every dart project. That's often written to refer to a common lint rule set, like `package:pedantic`. They either use that directly, or reference it, and then have a few project specific overrides.

.github/linters/.dart-lint.yml
```diff
@@ -0,0 +1,57 @@
+---
```
devoncarew, line 1: See my comment below for standard ways to configure analysis for dart projects, but generally, I think we should avoid in-lining rules from pedantic here; they'll get out of date, and we'd be analyzing a project with a different rule set then was intended by the project authors.

Dockerfile
cvega: I imagine super-linter will want to tag the versions of the linters and languages it supports. I had a more dynamic url for grabbing latest, but this would result in users not being certain of the version installed, and the container possibly changing without a related commit/version bump.

devoncarew: Ah, ok, fixing to a specific version then makes sense, so the container image is reproducable.

We may then want to add some in-line about where to find the current stable version number for dart (https://dart.dev/get-dart#release-channels).

devoncarew:
One more comment - we'll likely want to run `pub get` on the user's project before analysis. Otherwise we'll see lots of analysis errors related to packages not being resolved.

docs/disabling-linters.md
cvega: Out of the box, I import the pedantic rules in `.dart-lint.yml`. I can change this to `analysis-options.yaml` to match dart conventions one would expect. I notice the linter still reports `empty_constructor_bodies` when no `--options` are specified which leads me to believe a default set of rules exists. Typically we want users to provide their own `analysis-options.yaml`. You'll notice most of the linters have a default state, If I provide an empty `analysis-options.yaml` without default rules the bad rule test doesn't get caught. I think (could be wrong)  we want to provide a sane default for people that haven't propagated their own rules yet. What do you think is the best way to do that is short of not using the `--options` flag?

Dockerfile
cvega: Recommendation added.

README.md
cvega: Recommendation added

.automation/test/dart/dart_bad_1.dart
cvega: Ran `dartfmt` on both test files.

docs/disabling-linters.md
cvega: Updated to use `analysis_options.yaml`

admiralAwkbar:
@cvega really nice work on this!
We can add the flag to allow users to name the linter file what they choose as well :)
that should help with that small issue
Other than that, the code looks great!

cvega:
@admiralAwkbar thanks for the positive feedback. From my perspective this is almost there. Couple of outstanding items/decisions. 

- default pedantic rules, I can add provisions on my side to keep the rules synchronized with latest if this is a concern for the dart community. I planned on doing this for the dart-sdk anyways and making sure super-linter had stable dart releases added ASAP:
  - rules: https://github.com/dart-lang/pedantic/blob/master/lib/analysis_options.1.9.0.yaml
  - might be worth noting this in docs, this only applies to users without a `analyze_options.yaml` file
- Adding `pub get`. Not sure if we want this to be an isolated stage in CI or get called by super-linter. For what it's worth, pub is part of the sdk and easy enough to add. I like the idea, but I'm also cautious because this is two behaviors versus just linting.
- add `--fatal-infos` to exit `1` on lints

devoncarew:
Hey - my thoughts on the above questions:

> * default pedantic rules, I can add provisions on my side to keep the rules synchronized with latest if this is a concern for the dart community. I planned on doing this for the dart-sdk anyways and making sure super-linter had stable dart releases added ASAP:
>   
>   * rules: https://github.com/dart-lang/pedantic/blob/master/lib/analysis_options.1.9.0.yaml
>   * might be worth noting this in docs, this only applies to users without a `analyze_options.yaml` file

It sounds like a project's analyze_options.yaml will take precedence, so I have a lot less concerns about in-lining a copy of the pedantic rules.

Note that we do have plans to come up with a new recommended rule set for Dart. This would act similar to pedantic; it would likely have two variants to the rules - one set of rules for very common, high value lints. And another, larger set of more proscriptive lints. We'd transition all new projects over to using this rules set in our project templates.

We don't have a hard time frame for that, but would expect it to be perhaps a quarter (3-4 months) out. We'd almost certainly want to evolve this integration once that rule set is available to use those rules instead of the pedantic ones.

> * Adding `pub get`. Not sure if we want this to be an isolated stage in CI or get called by super-linter. For what it's worth, pub is part of the sdk and easy enough to add. I like the idea, but I'm also cautious because this is two behaviors versus just linting.

I don't know enough about the super-linter to say, but you definitely want some process to run `pub get` before running the analyzer.

> * add `--fatal-infos` to exit `1` on lints

Is there a way to configure the super-linter for projects? If yes, this would make a good project config option. If no, I suspect we should default this to off.

admiralAwkbar:
@cvega you got any cycles this week to close this up?

cvega:
Absolutely. I think for the time being we need to stick with the `--fatal-infos` for CI purposes otherwise `dartanalyzer` exits `0` even when it discovers a lint. Will finalize this evening (PST).

PR #384, opened 2020-07-06
Fixed config load issue caused by quotes

### LINKED ISSUE #368: release 3.2.0 seems to ignore user provided lint override files
**Describe the bug**
We have two issues that we must override in our codebase - one in markdown and one in python. I have merged all the ongoing changes to the template files while preserving our two overrides. Super Linter acknowledges the two user provided files and then seems to not actually use them when running the markdown and python lints.

**To Reproduce**
1. Go to the debug-superlint branch in chia-blockchain - https://github.com/Chia-Network/chia-blockchain/tree/debug-superlint
2. Cause a push or pull (feel free to fork)
3. Hit the super-lint action
4. See Markdown MD024 errors that should be overriden by https://github.com/Chia-Network/chia-blockchain/blob/d1a8a3dc7e3963836d73c3550edcf8a8b313d3cb/.github/linters/.markdown-lint.yml#L26 and python errors that should be overriden by https://github.com/Chia-Network/chia-blockchain/blob/d1a8a3dc7e3963836d73c3550edcf8a8b313d3cb/.github/linters/.python-lint#L246

**Expected behavior**
.markdown-lint.yml and .python-lint should override the lint errors (as they have in earlier releases.

**Screenshots**

**Additional context**
You can follow my debugging attempts in this PR: https://github.com/Chia-Network/chia-blockchain/pull/302
A recent run - https://github.com/Chia-Network/chia-blockchain/runs/833154863?check_suite_focus=true#step:4:56

leviem1:
Fixes #368

## Proposed Changes

- Removed quotes surrounding declare causing variable to not read in correctly

## Readiness Checklist

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -336,7 +336,7 @@ GetLinterRules() {
     ########################################
     # Update the path to the file location #
     ########################################
-    declare -g "${LANGUAGE_LINTER_RULES}=$GITHUB_WORKSPACE/$LINTER_RULES_PATH/${!LANGUAGE_FILE_NAME}"
+    eval "${LANGUAGE_LINTER_RULES}=$GITHUB_WORKSPACE/$LINTER_RULES_PATH/${!LANGUAGE_FILE_NAME}"
   else
     ########################################################
     # No user default provided, using the template default #


nemchik:
I don't think shellcheck will like this (linter will probably return an error). Can you see if this as an alternative would work:
```bash
eval "${LANGUAGE_LINTER_RULES}=$GITHUB_WORKSPACE/$LINTER_RULES_PATH/${!LANGUAGE_FILE_NAME}"
```

leviem1:
Thanks! Good catch! That was my backup plan as well. ~~Is there any reason why we wouldn't want to use eval? I saw some negative things about it, but then again, that's StackOverflow.~~ Beat me to it.

Also, do you prefer a squash, amend, or just commit?

nemchik:
Not my call on squash, but my opinion is amend. It leaves a much more logical commit history and can be referenced in the PR even if the branch is squashed when merged.

As for the negatives of eval, that is sometimes true with other languages, particularly JavaScript. bash is much less of a concern with regards to using eval for various reasons.

nemchik:
Nice, as long as tests pass this LGTM.

leviem1:
I appreciate the help! I'm always thankful for a good learning opportunity and your explanations were excellent and constructive. Have a good day!

leviem1:
Can I get any insight? I did test this but I guess I missed a step in testing? I'll look at it locally until I hear back

nemchik:
There seems to have been an issue with GitHub actions. If you add another commit to the branch it should trigger another run. You amend an existing commit and reword the commit message and then force push and it'll count.

leviem1:
Just curious, would a force push also work?

PR #391, opened 2020-07-07
Editorconfig special case

GaboFDC:
Following the logic from [here](https://github.com/github/super-linter/pull/221/files#diff-c7b422f2f140a91b67618e5aa86d8d88R477) if we don't specify any linter flag (i.e use all linters) then, editorconfig will run against the entire code base, even if there is no `.editorconfig` present, and it will even check the `.git/` folder.

## Proposed Changes
In some sense this is expected because we are not specifying any linter, so all should be run. But as this one is special and lints the entire code base, I think is worth putting an extra check to only do it if either the flag is set, or `.editorconfig` is present as a special case because this behavior is different from other linters (run only against files from that extension/type).

I'm open to all feedback here. specially @mstruebing 

PD: I think another needed change (independent of this one) is to exclude `.git/` from editor config.

PD2: Not sure if this is a typo or what is the purpose of [this](https://github.com/github/super-linter/pull/221/files#diff-2b006ea0cfcfac29e33a1b9eef9974c1R1138)?

## Readiness Checklist

### DIFF
--- a/lib/validation.sh
+++ b/lib/validation.sh
@@ -75,6 +75,12 @@ function GetValidationInfo() {
   VALIDATE_EDITORCONFIG=$(echo "$VALIDATE_EDITORCONFIG" | awk '{print tolower($0)}')
   VALIDATE_HTML=$(echo "$VALIDATE_HTML" | awk '{print tolower($0)}')
 
+  #############################
+  # Editorconfig special case #
+  #############################
+  LINTER_RULES_PATH="${LINTER_RULES_PATH:-.github/linters}"               # Linter Path Directory
+  EDITORCONFIG_FILE_NAME='.editorconfig'
+
   ################################################
   # Determine if any linters were explicitly set #
   ################################################
@@ -484,10 +490,13 @@ function GetValidationInfo() {
       VALIDATE_EDITORCONFIG="false"
     fi
   else
-    # No linter flags were set - default all to true
-    VALIDATE_EDITORCONFIG="true"
+    # No linter flags were set
+    # special case cehcking for .editorconfig
+    if [ -f "$GITHUB_WORKSPACE/$LINTER_RULES_PATH/$EDITORCONFIG_FILE_NAME" ]; then
+      VALIDATE_EDITORCONFIG="true"
+    fi
   fi
-  
+
   ####################################
   # Validate if we should check HTML #
   ####################################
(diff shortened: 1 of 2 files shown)


nemchik:
https://github.com/github/super-linter/pull/221/files#diff-c7b422f2f140a91b67618e5aa86d8d88R477 your link seems incorrect

nemchik:
I am guessing this branch was based on the branch used in https://github.com/github/super-linter/pull/372
could you rebase on master from this repo so we can see the changes separately?

GaboFDC:
> 
> 
> I am guessing this branch was based on the branch used in #372
> could you rebase on master from this repo so we can see the changes separately?

Yes sorry, my mistake, it is fixed now

mstruebing:
I think it should be done to not run the editorconfig-checker when no `.editorconfig` is present, even if it could be that you could have `.editorconfig`-files in subdirectories, but tbh I never saw such a project without also a root `.editorconfig`-file.

To hide it behind a flag: I think it is okay because it is somehow a bit different behaviour than other linters of this project.

nemchik:
> PD: I think another needed change (independent of this one) is to exclude `.git/` from editor config.

Agreed.

> PD2: Not sure if this is a typo or what is the purpose of [this](https://github.com/github/super-linter/pull/221/files#diff-2b006ea0cfcfac29e33a1b9eef9974c1R1138)?

Seems like a typo.

lib/validation.sh
```diff
@@ -75,6 +75,12 @@ function GetValidationInfo() {
   VALIDATE_EDITORCONFIG=$(echo "$VALIDATE_EDITORCONFIG" | awk '{print tolower($0)}')
   VALIDATE_HTML=$(echo "$VALIDATE_HTML" | awk '{print tolower($0)}')
 
+  #############################
+  # Editorconfig special case #
+  #############################
+  LINTER_RULES_PATH="${LINTER_RULES_PATH:-.github/linters}"               # Linter Path Directory
```
nemchik, line 81: If I am understanding this correctly, if users don't set `LINTER_RULES_PATH` manually it'll be set to `.github/linters` and if there is an `.editorconfig` there then the linter will still run?

Logically I think it makes more sense to not include this block of changes and instead change the logic here https://github.com/github/super-linter/pull/391/files#diff-c7b422f2f140a91b67618e5aa86d8d88R495 to `if [ -f "$GITHUB_WORKSPACE/.editorconfig" ];`

mstruebing:
> PD2: Not sure if this is a typo or what is the purpose of this?

This was for debugging purpose while developing, I think I missed to remove it and it doesn't get caught while reviewing.

GaboFDC:
Filled #395 for the `.git/` folder

nemchik:
LGTM now.

admiralAwkbar:
@GaboFDC @nemchik @mstruebing amazing work in here all!
Really thank you so much for all the hard work

Issue #352, opened 2020-07-01
Linting .tsx and .jsx files


bjarkehs:
Hey!

Is there something I'm missing in order for it to lint my .tsx files? (Typescript JSX files)

It seems only my .ts files are linted with eslint.

StalkAltan:
Hey @bjarkehs !

It looks like the reason super-linter isn't linting .jsx/.tsx files is because the JS/TS linter configurations are only searching for .js/.ts

See here for an example from the code: [link](https://github.com/github/super-linter/blob/6c152d271e400f0db27514e0d1a647b41937741b/lib/linter.sh#L1032)

admiralAwkbar:
@bjarkehs should they be linted the exact same way? if that is the case, it should be easy enough for us to make a PR that also searches for those extensions :)

Sorry my JS/TS is not the strongest...

PR #394, opened 2020-07-07
Added support for jsx and tsx files

thematchless:
I added support for linting jsx and tsx files (Utilizing: eslint) https://github.com/github/super-linter/issues/352

## Changes
- added the recognition of .jsx and .tsx files
- added two config params to control the linting


thematchless:
converting to draft because the test cases are missing

admiralAwkbar:
@thematchless since it uses the same linting command, wouldn't it be more simple to just add those file extensions to the standard js loops? or do they really need to be treated as another language to manage? Just trying to understand better the path you took. Thanks! Also, my knowledge of those is super limited...

thematchless:
@admiralAwkbar that was my first attempt to solve this. But i thought it would be nicer if you could control if you really want to lint the .jsx or .tsx files. There might be some guys who want their javascript checked but not their .jsx files.

admiralAwkbar:
@thematchless makes sense.. and agree :)

Issue #305, opened 2020-06-29
.dockerlintrc is ignored


zkoppert:
> @past-due i tested and validated it, the command runs: `/dockerfilelint/bin/dockerfilelint -c .dockerlintrc <Dockerfile>` which ius using the config file as intended.
> 
> Am I missing something else on this?

@admiralAwkbar:
`-c .dockerlintrc`
should be
`-c <path to folder _containing_ .dockerlintrc>`

The command it is executing will run, but it's not using the actual `.dockerlintrc` in my tests. In https://github.com/replicatedhq/dockerfilelint/blob/master/lib/index.js#L85 you can see it will fail to open `.dockerlintrc/.dockerlintrc` and just silently use the defaults.

I am testing with a `.dockerlintrc` that disables `latest_tag`:
https://github.com/past-due/warzone2100/blob/2020_06_ci_lint_1/.dockerfilelintrc#L78

And Dockerfiles that use, for example, `FROM archlinux:latest`:
https://github.com/past-due/warzone2100/blob/2020_06_ci_lint_1/docker/archlinux-latest/Dockerfile

If this merge worked, it should ignore `latest_tag`. Instead, it's reporting it:
```
File:[./docker/archlinux-latest/Dockerfile]
ERROR! Found errors in [/dockerfilelint/bin/dockerfilelint] linter!
ERROR:[
File:   ./docker/archlinux-latest/Dockerfile
Issues: 1

Line 1: FROM archlinux:latest
Issue  Category      Title                 Description
    1  Clarity       Base Image Latest     Base images should not use the latest tag.
                     Tag]
```

And just to confirm, super-linter is properly detecting the `.dockerfilelintrc` and copying it to the appropriate location:
```
User provided file:[.dockerfilelintrc], setting rules file..
```

_Originally posted by @past-due in https://github.com/github/super-linter/pull/289#issuecomment-650280142_

LukeMondy:
Yes, I think a perhaps naive fix would be to change:
https://github.com/github/super-linter/blob/2b189bf16feb85b077890c2af15b008b4938b011/lib/linter.sh#L44
to be: 
```bash
DOCKER_LINTER_RULES="$DEFAULT_RULES_LOCATION/"     # Path to the Docker lint rules 
```
Might be worth testing to see if it helps or not (sorry, I don't have a good setup to test it myself)

julie-garrone:
I tried setting up a dockerfilelint configuration file using the latest super-linter image and it is still not working.
[...]

PR #397, opened 2020-07-08
Fix dockerfilelint -c option usage

### LINKED ISSUE #355: dockerfilelinter
Hi

so i want to lint my Dockerfile.
i've added the following:

```
- name: Super Linter
        uses: docker://github/super-linter:v3
        env:
          VALIDATE_DOCKER: true
          ACTIONS_RUNNER_DEBUG: true
          VALIDATE_YAML: true
```

i've copied the .dockerfilelintrc from the TEMPLATES directory and put it within .github/linters folder on my repo.
and disabled one of the rules (`latest_tag: on` to `latest_tag: off`)

when it got executed i see the line that indicate my overwrite:
`User provided file:[.dockerfilelintrc], setting rules file...`

but i still get the error:

```
ERROR! Found errors in [/dockerfilelint/bin/dockerfilelint] linter!
ERROR:[File:   /github/workspace/Dockerfile ]

Issue  Category      Title                 Description
    1  Clarity       Base Image Latest     Base images should not use the latest tag.
```

what am i missing?

thanks!

past-due:
dockerfilelint's "-c" option expects a path to the folder *containing* the DOCKER_LINTER_RULES file

Based on @LukeMondy 's suggestion: https://github.com/github/super-linter/issues/305#issuecomment-651571788

Should fix #355

## Proposed Changes

- Pass the path to the folder _containing_ the `DOCKER_LINTER_RULES` file, instead of the `DOCKER_LINTER_RULES` itself
   See: https://github.com/replicatedhq/dockerfilelint/pull/138
   > This option needs to be used with the path for the file not directly to the file. This is because https://github.com/replicatedhq/dockerfilelint/blob/master/lib/index.js#L85 adds the `.dockerfilelintrc` automatically.

## Readiness Checklist

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -54,8 +54,7 @@ TYPESCRIPT_STANDARD_LINTER_RULES=''                                     # ENV st
 ANSIBLE_FILE_NAME='.ansible-lint.yml'                                   # Name of the file
 ANSIBLE_LINTER_RULES="$DEFAULT_RULES_LOCATION/$ANSIBLE_FILE_NAME"       # Path to the Ansible lint rules
 # Docker Vars
-DOCKER_FILE_NAME='.dockerfilelintrc'                                    # Name of the file
-DOCKER_LINTER_RULES="$DEFAULT_RULES_LOCATION/$DOCKER_FILE_NAME"         # Path to the Docker lint rules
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@past-due since the file name cant change, there was no need to add to the path at the beginning, and then try to pull the path off later. I think this should solve the issue right?

PR #402, opened 2020-07-09
Make Raku Better

### LINKED ISSUE #401: Make Raku Install better
@nbrown As you can see from the updated **Dockerfile**, we have moved to a more elegant way of installing some of the dependancies.

example:
```
COPY --from=mstruebing/editorconfig-checker:2.1.0 /usr/bin/ec /usr/bin/editorconfig-checker
```

Could we utilize something similar to install Raku?
Possibly from `https://hub.docker.com/_/rakudo-star`?
Or does it not have all the binaries we need?

I have just noticed the build and compile times have gone up, and the Raku step is a bit long

thanks!

admiralAwkbar:
this closes #401 

Trying to make raku less expensive

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ ARG ARM_TTK_NAME='master.zip'
 ARG ARM_TTK_URI='https://github.com/Azure/arm-ttk/archive/master.zip'
 ARG ARM_TTK_DIRECTORY='/opt/microsoft'
 # Raku Linter
-ARG RAKU_VER="2020.06"
+ARG RAKU_VER="2020.01"
 ARG RAKU_INSTALL_PATH=/usr
 ARG RAKUBREW_HOME=/tmp/rakubrew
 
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@nbrown Any ideas whats the deal?

Dockerfile
```diff
@@ -30,7 +30,7 @@ ARG ARM_TTK_NAME='master.zip'
 ARG ARM_TTK_URI='https://github.com/Azure/arm-ttk/archive/master.zip'
 ARG ARM_TTK_DIRECTORY='/opt/microsoft'
 # Raku Linter
-ARG RAKU_VER="2020.06"
+ARG RAKU_VER="2020.01"
```
nemchik, line 33: With the way you're copying the binary from another docker container it should not be necessary to define this argument. I would say use the variable here in the reference to the image below but I don't think Dependabot would keep that up to date.

Dockerfile
```diff
@@ -162,17 +162,7 @@ RUN curl -sSLO https://github.com/pinterest/ktlint/releases/latest/download/ktli
...
-# Basic setup, programs and init
-RUN mkdir -p $RAKUBREW_HOME/bin \
-    && curl -sSLo $RAKUBREW_HOME/bin/rakubrew https://rakubrew.org/perl/rakubrew \
-    && chmod 755 $RAKUBREW_HOME/bin/rakubrew \
-    && eval "$($RAKUBREW_HOME/bin/rakubrew init Sh)"\
-    && rakubrew build moar $RAKU_VER --configure-opts='--prefix=$RAKU_INSTALL_PATH' \
-    && rm -rf $RAKUBREW_HOME/versions/moar-$RAKU_VER \
-    && rakubrew build-zef \
-    && rm -rf $RAKUBREW_HOME
+COPY --from=rakudo-star:2020.01 /usr/bin/raku /usr/bin/
```
nemchik, line 165: maybe try adding `RUN chmod +x /usr/bin/raku`

nbrown:
Sorry guys, I wasn't thinking about that cost and was using what I did in another docker image. Maybe we should use the alpine package from edge [1] instead or use rakubrew to download precompiled images (but they won't get installed to /usr/). Any preference

[1] https://pkgs.alpinelinux.org/packages?name=rakudo&branch=edge

Dockerfile
nbrown: The reason this doesn't work is because raku runs on the moar virtual machine and depends on NQP. I'll figure it out and propose a solution.

nbrown: So I thought that something like the following should work but I can't figure out what file is missing. I'm going to propose an alternative approach, but it is much faster than building moar, NQP and rakudo from scratch.

PR #403, opened 2020-07-09
Alternate approach to improve Rakudo install speed

### LINKED ISSUE #401: Make Raku Install better

nbrown:
I couldn't update the existing pull request  #402 but I believe this fixes #401 

I could not make copying from the other docker image work, but I sped it up significantly by using the rakudo package for alpine linux that is from the edge/testing repo.


zkoppert:
related to #402

PR #408, opened 2020-07-13
fine well do it this way

### LINKED ISSUE #407: Latest release 3.3.1 breaks workflow (error: pathspec <branch> did not match any file(s) known to git)
**Describe the bug**

Workflow throws the following error: 

`[error: pathspec <branch> did not match any file(s) known to git]`

The github action was being pulled with the `latest` tag.

Fixed the tag to a previous version (3.3.0) and it worked without any problems. The codebase was the same, the only change was the action version.

admiralAwkbar:
This closes #407 

this throws the fun little error message but also fixes the refs...
may need to find a better path eventually

### DIFF
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -26,7 +26,7 @@ function BuildFileList() {
   # Switch codebase back to the default branch to get a list of all files changed #
   #################################################################################
   SWITCH_CMD=$(
-    # git -C "$GITHUB_WORKSPACE" pull --quiet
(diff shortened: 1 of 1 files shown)


nemchik:
If this still doesn't work you might try
```
git -C "$GITHUB_WORKSPACE" reset --hard "$DEFAULT_BRANCH"
```
Since the branch should already be checked out and pulled via the checkout gha you should only need to reset the current branch to the default.

PR #415, opened 2020-07-14
Adding groovy

### LINKED ISSUE #363: Add Groovy Support
It should be here...
https://github.com/nvuillam/npm-groovy-lint

**Additional context**

**Note**
If this is a request to add a new language, after submitting this issue check out [the wiki](https://github.com/github/super-linter/wiki/Adding-new-language-support) for more info on how to accomplish that.

admiralAwkbar:
This closes #363 

This will add groovy support to the Super-Linter


lib/linter.sh
```diff
@@ -59,6 +59,9 @@ DOCKER_LINTER_RULES="$DEFAULT_RULES_LOCATION/$DOCKER_FILE_NAME"         # Path t
 # Golang Vars
 GO_FILE_NAME='.golangci.yml'                                            # Name of the file
 GO_LINTER_RULES="$DEFAULT_RULES_LOCATION/$GO_FILE_NAME"                 # Path to the Go lint rules
+# Groovy Vars
+GROOVY_FILE_NAME='.groovylintrc.json'                                   # Name of the file
+GROOVY_LINTER_RULES="$DEFAULT_RULES_LOCATION/$GROOVY_FILE_NAME"         # Path to the Go lint rules
```
adamdmharvey, line 64:
```suggestion
GROOVY_LINTER_RULES="$DEFAULT_RULES_LOCATION/$GROOVY_FILE_NAME"         # Path to the Groovy lint rules
```

admiralAwkbar: thank you!

adamdmharvey: you're welcome! note you resolved, though i don't see a commit pushed to ACTUALLY fix it :)

lib/buildFileList.sh
```diff
@@ -421,6 +421,15 @@ function BuildFileList() {
       # Set the READ_ONLY_CHANGE_FLAG since this could be exec #
       ##########################################################
       READ_ONLY_CHANGE_FLAG=1
+    elif [ "$FILE_TYPE" == "groovy" ] || [ "$FILE_TYPE" == "jenkinsfile" ]; then
+      ################################
```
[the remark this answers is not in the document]
admiralAwkbar: Added :)

nvuillam:
little comment: I see that `--failon error` is used
With default rules, only parse error and DeadCode are with "error" level
Maybe it would be more efficient to set the level to `warning` so more problems are catched ?

admiralAwkbar:
@nvuillam Would you propose to have it set to fail on error and warning? or will it show the warnings, and only fail on errors?

nvuillam:
It will show error & warnings :)

admiralAwkbar:
@nvuillam can you make a PR against the default config file to set the rules you believe should be marked as errors?

I think that would help a lot of users who take the defaults in and help establish some better practices

`/TEMPLATES/.groovylintrc.json`

nvuillam:
@admiralAwkbar  

- **error** means "no sense, it will crash"
- **warning** means "could be rewritten in a much cleaner way

What is the target of your superlinter ?
If response 2, you can just call using --failon warning

About pre-redefined rules, i already propose "recommended", and also the one I use for the package development ( that you used, which seems a good idea)

If someone wants to consider warnings as errors, he/she can submit a new present directly on npm-groovy-lint repo :)

PR #428, opened 2020-07-16
Use Docker instead of run-time build

zkoppert:
No related issue

## Proposed Changes

- When a user goes to the Marketplace to add the super linter, the button shown below should guide them to use the docker image and not the dockerfile which requires a large amount of build time at each run.

Screen Shot 2020-07-16 at 6 57 50 AM
low performant instructions show when you click the green button...Screen Shot 2020-07-16 at 6 45 57 AM

## Readiness Checklist

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -3,7 +3,7 @@ author: 'GitHub'
 description: 'It is a simple combination of various linters, written in bash, to help validate your source code.'
 runs:
   using: 'docker'
-  image: 'Dockerfile'
+  image: 'docker://github/super-linter'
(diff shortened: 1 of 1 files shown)


alerque:
@leviem1 No you're reading this right, the update completely eschews the version tags shown in the Marketplace, you can't get specific versions like that at all. This forces all uses of the normal Actions syntax to use whatever the hard coded Docker version is (in this case a sliding latest release tag, but still hard coded). You can only get the other versions if you use the Docker syntax yourself.

PR #435, opened 2020-07-17
Adjust handling of docker dependencies

nemchik:
Fixes #

## Proposed Changes

- The `COPY --from=` syntax functions correctly, but dependabot does not update it
- Use `FROM` build stages in the dockerfile so that dependabot will actually update the versions
- The last `FROM` in the dockerfile is used as the base image, any before it are used as build stages that you can copy from
- I also moved the editorconfig-checker copy line up to group it with the rest

## Readiness Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,17 @@
 ###########################################
 ###########################################
 
+#########################################
+# Get dependency images as build stages #
+#########################################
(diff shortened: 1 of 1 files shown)


nemchik:
P.S. at least one of these dependencies ( https://github.com/editorconfig-checker/editorconfig-checker/releases ) has a newer release than the version included in this repo, so the next time dependabot runs after this merges we should see immediate results.

PR #447, opened 2020-07-21
Automerge Dependabot dependencies

admiralAwkbar:
This will help make sure we're always up to date and taking in the latest security issues

### DIFF
--- a/.github/workflows/automerge-dependabot.yml
+++ b/.github/workflows/automerge-dependabot.yml
@@ -1,0 +1,47 @@
+---
+#######################################
+#######################################
+## Dependabot automerge dependencies ##
+#######################################
+#######################################
+
+#
+# Documentation:
+# https://medium.com/@toufik.airane/automerge-github-dependabot-alerts-with-github-actions-7cd6f5763750
+#
+
+######################
+# name of the action #
+######################
+name: automerge on pull request
+
+###############
+# When to run #
+###############
+on: [pull_request]
+
+#################
+# Start the job #
+#################
+jobs:
+  automerge:
+    name: automerge dependabot
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+    steps:
+      - name: merge
+        uses: actions/github-script@0.2.0
+        with:
+          script: |
+            github.pullRequests.createReview({
+              owner: context.payload.repository.owner.login,
+              repo: context.payload.repository.name,
+              pull_number: context.payload.pull_request.number,
+              event: 'APPROVE'
+            })
+            github.pullRequests.merge({
+              owner: context.payload.repository.owner.login,
+              repo: context.payload.repository.name,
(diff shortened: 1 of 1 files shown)


nemchik:
Can't we just use https://dependabot.com/docs/config-file/#automerged_updates in https://github.com/github/super-linter/blob/master/.github/dependabot.yml ?

jwiebalk:
@nemchik auto-merge isn't currently supported by the "native" GitHub dependabot configuration.

> Auto-merge will not be supported in GitHub-native Dependabot for the foreseeable future. 

https://github.com/dependabot/dependabot-core/issues/1973#issuecomment-640918321

admiralAwkbar:
@zkoppert give me the power!!!!!!!

admiralAwkbar:
@zkoppert

nemchik:
> @nemchik auto-merge isn't currently supported by the "native" GitHub dependabot configuration.
> 
> > Auto-merge will not be supported in GitHub-native Dependabot for the foreseeable future.
> 
> [dependabot/dependabot-core#1973 (comment)](https://github.com/dependabot/dependabot-core/issues/1973#issuecomment-640918321)

That's unfortunate. It would be a lot cleaner, and seems to already have some decent controls in place (preview did, with the name and type matching). Oh well. If this method is the only alternative currently we'll have to make it work.

.github/workflows/automerge-dependabot.yml
```diff
@@ -0,0 +1,47 @@
...
+on: [pull_request]
+
+#################
+# Start the job #
+#################
+jobs:
+  automerge:
+    name: automerge dependabot
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
```
zkoppert, line 30: @jasonmacgowan Do you know if this can be impersonated by a bad actor?

zkoppert:
>we'll have to make it work.

@nemchik Any specific concerns on how this might be problematic?

nemchik:
> > we'll have to make it work.
> 
> @nemchik Any specific concerns on how this might be problematic?

None at all! I have never seen this used before and my initial reaction came from it being "bigger" (more to look at and more to read through to understand what's going on) compared to a little extra yaml via the method dependabot preview has.

.github/workflows/automerge-dependabot.yml
admiralAwkbar: I don't think that would be possible on this... @jasonmacgowan give us your knowledge...

jasonmacgowan: I don't think this can easily spoofed since it's populated from the pull request and not a commit.

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

zkoppert: Thanks @jasonmacgowan

PR #449, opened 2020-07-21
Braces

nemchik:
Fixes #

## Proposed Changes

- Format markdown files (mostly README.md files in all subfolders)
- Adjust yaml formatting (github actions files)
- Use `${VAR}` instead of `$VAR` in shell scripts and others known to support this syntax.

## Readiness Checklist

### DIFF
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -16,7 +16,7 @@ function BuildFileList() {
   ################
   # print header #
   ################
-  if [[ $ACTIONS_RUNNER_DEBUG == "true" ]]; then
+  if [[ ${ACTIONS_RUNNER_DEBUG} == "true" ]]; then
     echo ""
     echo "----------------------------------------------"
     echo "Pulling in code history and branches..."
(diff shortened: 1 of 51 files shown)


admiralAwkbar:
@nemchik your doing the dirt work... I appreciate it...

nemchik:
Ok, most of this was regex replace and tooling, but i have manually reviewed all of the changes and double checked them. Should be good to go.

nemchik:
Added the only tag that seemed to fit lol

lib/buildFileList.sh
```diff
@@ -86,26 +86,26 @@ function BuildFileList() {
     ###########################
     # Extract just the file and extension, reverse it, cut off extension,
     # reverse it back, substitute to lowercase
-    FILE_TYPE=$(basename "$FILE" | rev | cut -f1 -d'.' | rev | awk '{print tolower($0)}')
+    FILE_TYPE=$(basename "${FILE}" | rev | cut -f1 -d'.' | rev | awk '{print tolower(${0})}')
```
GaboFDC, line 89: I think this broke the command, can you check?
This also shows that we don't have test for when `VALIDATE_ALL_CODEBASE` is false.
Kind of poetic that @admiralAwkbar broke the awk 😂

PR #455, opened 2020-07-22
Python flake8

### LINKED ISSUE #123: Flake8 support
I prefer flake8 over pylint and would love flake8 support for Python3 linting

Flake8 support

**Additional context**
Documentation here: https://flake8.pycqa.org/en/latest/

GaboFDC:
Fixes #123

## Proposed Changes

1. Allow to use flake8 to lint python
2. Based on how we do it for eslint and standard for JS
3. Any feedback is welcome
4. Some sorting for better maintainability. There are still some things that could be sorted but require more work, Maybe open an issue with that

Not 100% sure if I got the `VALIDATE_PYTHON*` vars correctly. Also as when running in local `VALIDATE_ALL_CODEBASE` is true, I couldn't find a reliable way of testing it.

## Readiness Checklist


GaboFDC:
@admiralAwkbar @nemchik this should be ready for a review

admiralAwkbar:
@GaboFDC so everything looks great!

## 1 Concern

- You leave the `VALIDATE_PYTHON` in the dockerfile but remove it from the `README`

For backward compatibility, we should leave the `VALIDATE_PYTHON` in both places, and add logic to see:

if python is set to validate, then it does both pylint and flake8, else only do what is selected

GaboFDC:
> For backward compatibility, we should leave the `VALIDATE_PYTHON` in both places, and add logic to see:
> 
> if python is set to validate, then it does both pylint and flake8, else only do what is selected

I think this will be not intended as for current users, it will start linitng both, and they may detect different things. So how it currently is, if you set `VALIDATE_PYTHON` it will lint with pylint, as it was doing before. And if you define either of the explicit ones, it will lint with the one defined.
For linting with both, you can define both, or leave `VALIDATE_ALL_CODEBASE` as true.
WDYT?

zkoppert:
>For linting with both, you can define both, or leave VALIDATE_ALL_CODEBASE as true.
This doesn't seem intuitive to me to have VALIDATE_ALL_CODEBASE enable both of them. I think it would be best to maintain the behavior you described [here](https://github.com/github/super-linter/pull/455/files#diff-04c6e90faac2675aa89e2176d2eec7d8R209) when VALIDATE_ALL_CODEBASE is enabled.

GaboFDC:
> This doesn't seem intuitive to me to have VALIDATE_ALL_CODEBASE enable both of them. I think it would be best to maintain the behavior you described [here](https://github.com/github/super-linter/pull/455/files#diff-04c6e90faac2675aa89e2176d2eec7d8R209) when VALIDATE_ALL_CODEBASE is enabled.

This is how currently works for JS, when VALIDATE_ALL_CODEBASE is enabled it lints with standard and with eslint. I kept the same behavior

rolandgeider:
Just a user interested in this. What is the behaviour if you want to validate all languages but want to use flake8? Won't setting VALIDATE_PYTHON_FLAKE8 mean that you have to manually specify all the other languages as well?

GaboFDC:
@rolandgeider 🤔 I think that would be the case, not sure how users of JS are dealing with similar cases.
Do you have any input on who would this be managed better?

rolandgeider:
I mean, it would work as it does now, it's just not very convenient because pylint and flake are more or less treated like language switches. Perhaps there could be another configuration option where you can set the prefered linter(s) for JS, python, etc, without triggering that behaviour, but that would go beyond the scope of this issue.

GaboFDC:
I think we can get this merge while we conclude #472 
wdyt?

artob:
> Perhaps there could be another configuration option where you can set the prefered linter(s) for JS, python, etc, without triggering that behaviour, but that would go beyond the scope of this issue.

As just another user's point of view, `VALIDATE_PYTHON=flake8` would be preferable to `VALIDATE_PYTHON_FLAKE8=true`.

GaboFDC:
> As just another user's point of view, `VALIDATE_PYTHON=flake8` would be preferable to `VALIDATE_PYTHON_FLAKE8=true`.

Given the current way it works is not really possible to do it that way. But that will be taken into account on #472

GaboFDC:
@admiralAwkbar were your concerns about `VALIDATE_PYTHON` resolved? or do you still think it needs a change?

PR #462, opened 2020-07-22
Simplify extension extraction

nemchik:
Fixes https://github.com/github/super-linter/pull/457 (not a fix, just a follow up)

## Proposed Changes

1. Use bash shell expansion to simplify the logic to extract the file extension and convert it to lowercase

No need to put my thing down flip it and reverse it.
image

## Readiness Checklist

### DIFF
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -84,9 +84,10 @@ function BuildFileList() {
     ###########################
     # Get the files extension #
     ###########################
-    # Extract just the file and extension, reverse it, cut off extension,
-    # reverse it back, substitute to lowercase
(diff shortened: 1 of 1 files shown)


nemchik:
p.s.
```bash
FILE="/some/other/place.with/myfile.TAR.GZ"
FILE_TYPE=$(basename "${FILE}" | rev | cut -f1 -d'.' | rev | awk '{print tolower($0)}')
echo "${FILE_TYPE}"

FILE="/some/other/place.with/myfile.TAR.GZ"
FILE_TYPE=${FILE##*.}
FILE_TYPE=${FILE_TYPE,,}
echo "${FILE_TYPE}"
```

PR #463, opened 2020-07-22
found the missing link

admiralAwkbar:
Found a missing var, also cleaning up some redundant code...


admiralAwkbar:
@jwiebalk and I paired to cleanup some loops and really make it easier to contribute

nemchik:
Lots of commits going on here. Still in progress?
image

admiralAwkbar:
@nemchik sorry for the prolonged... found a small error, that led to another small error, which leads to me making terrible decisions lol... almost

nemchik:
> @nemchik sorry for the prolonged... found a small error, that led to another small error, which leads to me making terrible decisions lol... almost

No worries, I'm not bothered, just don't want to suddenly see passing checks (once you get it ironed out) and merge before you've got everything you want in the branch.

admiralAwkbar:
@nemchik Would you solve this by doing:
ERRORS_FOUND_YML=0                 # Count of errors found
^--------------^ SC2034: ERRORS_FOUND_YML appears unused. Verify use (or export if used externally).

```bash
ERRORS_FOUND_YML=0                 # Count of errors found
echo "${ERRORS_FOUND_YML}" > /dev/null 2>&1 || true     # Workaround SC2034
```

admiralAwkbar:
@nemchik or I could put a block around it and disable for the block... not sure whats the protocol on `shellcheck` and if we should solve every case, or hide some :)

nemchik:
> @nemchik Would you solve this by doing:
> ERRORS_FOUND_YML=0 # Count of errors found
> ^--------------^ SC2034: ERRORS_FOUND_YML appears unused. Verify use (or export if used externally).
> 
> ```shell
> ERRORS_FOUND_YML=0                 # Count of errors found
> echo "${ERRORS_FOUND_YML}" > /dev/null 2>&1 || true     # Workaround SC2034
> ```


```shell
ERRORS_FOUND_YML=0                 # Count of errors found
export ERRORS_FOUND_YML
```

admiralAwkbar:
it looks like it works but I'm going to test it a bit more
it's a lot of changes :)

lib/buildFileList.sh
```diff
@@ -99,9 +99,9 @@ function BuildFileList() {
     #########
     #echo "FILE_TYPE:[${FILE_TYPE}]"
 
-    #####################
-    # Get the CFN files #
-    #####################
+    ################################
+    # Get the CLOUDFORMATION files #
+    ################################
     if [ "${FILE_TYPE}" == "yml" ] || [ "${FILE_TYPE}" == "yaml" ]; then
```
GaboFDC, line 105: I think there is a bug, here we only check yaml files. but on 
https://github.com/github/super-linter/pull/463/files#diff-2b006ea0cfcfac29e33a1b9eef9974c1R1214 and https://github.com/github/super-linter/pull/463/files#diff-409b55ad463cfac3e14fdaa2a919cfebR560 we check for json also.

admiralAwkbar: We check also in the json loop for cloud formation as well

GaboFDC: Sorry, can you point where? maybe I'm confused

PR #467, opened 2020-07-23
chore: remove redundant package-lock.json

tobiasbueschel:
Hi all,

Really appreciate all the work that already went into `super-linter`. I've noticed that there's a lonely `package-lock.json` file in the root of the project which was introduced by this commit: https://github.com/github/super-linter/commit/40ca4c5f96a1911ff7752a783ef8cc5cb8588589#diff-32607347f8126e6534ebc7ebaec4853d

Unless I'm missing some context, this file is not used and can be removed :)

I've also added it to `.gitignore` so it's not mistakenly added to the root again.


admiralAwkbar:
@tobiasbueschel nice catch!

tobiasbueschel:
Thanks @admiralAwkbar - have a good day!

PR #468, opened 2020-07-23
fix disabling-linters documentation

mil7:
On Disabling linters and Rules documentation I fixed two minor issues:
* Fix link to Clojure
* Everything after Clojure chapter was not properly defined (anymore)
* Remove surplus horizontal line


admiralAwkbar:
@mil7 Good catch on this!

PR #471, opened 2020-07-23
Better Powershell install

admiralAwkbar:
Making PowerShell install faster, cleaner.. better...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ RUN apk add --update --no-cache \
     ansible-lint \
     bash \
     curl \
-    gcc libstdc++\
+    gcc libstdc++6\
     git git-lfs\
     go \
     icu-libs \
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,7 +50,7 @@ RUN apk add --update --no-cache \
     ansible-lint \
     bash \
     curl \
-    gcc libstdc++6\
+    gcc \
     git git-lfs\
     go \
     icu-libs \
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@zkoppert on my local build it was about 2 min faster... so it should help
seems ARM ttk is where the next chunk is at...

Dockerfile
```diff
@@ -96,21 +93,20 @@ ENV PATH="/node_modules/.bin:${PATH}"
...
-    && curl -s https://api.github.com/repos/powershell/powershell/releases/${PWSH_VERSION} \
-    | grep browser_download_url \
-    | grep linux-alpine-x64 \
-    | cut -d '"' -f 4 \
-    | xargs -n 1 wget -O - \
-    | tar -xzC ${PWSH_DIRECTORY} \
-    && ln -sf ${PWSH_DIRECTORY}/pwsh /usr/bin/pwsh \
-    && pwsh -c 'Install-Module -Name PSScriptAnalyzer -RequiredVersion ${PSSA_VERSION} -Scope AllUsers -Force'
+COPY --from=powershell /usr/bin/pwsh /usr/bin/
+CMD /usr/bin/pwsh -c 'Install-Module -Name PSScriptAnalyzer -Scope AllUsers -Force'
```
nemchik, line 109: This should be `RUN` (not `CMD`)

admiralAwkbar:
@nemchik when set to `RUN` and not `CMD` it fails to see the bin dir object we just set... this is driving me a bit nuts lol

nemchik:
> @nemchik when set to `RUN` and not `CMD` it fails to see the bin dir object we just set... this is driving me a bit nuts lol

@admiralAwkbar
- `CMD` executes inside the container (after it is built). Ex: The default command to be run when an end user runs `docker run github/super-linter` would be whatever is in `CMD`.
-  `RUN` executes while building the container. Ex: I need X, Y, or Z installed inside the container so that when it runs it will have those things.

Ref:
https://docs.docker.com/engine/reference/builder/#cmd
https://docs.docker.com/engine/reference/builder/#run

The big question here is whether the `pwsh` binary executable is self contained and portable or if it requires everything else found in the tar.gz that was originally used to install. If it does require all the other stuff we might not be able to pull only the one file (might be best to keep the current method).

admiralAwkbar:
So too many deps not fulfilled.... may have to wait until the better image is available... still fun to try :)

PR #473, opened 2020-07-24
Adds optional security linting for Terraform

cesar-rodriguez:
## Proposed Changes

1. Adding optional security linting of terraform files
2. This uses Terrascan with the VALIDATE_TERRAFORM_TERRASCAN environment variable
3. The VALIDATE_TERRAFORM_TERRASCAN variable is set to "false" by default

## Readiness Checklist


admiralAwkbar:
@cesar-rodriguez Really nice work!

I took the liberty to remove a dead groovy loop that was bad merged in...

Also, you had it default to false, but all other linters default to true, so I wanted to keep it on the same level and consistency in action. Hope that isn't a big issue?

cesar-rodriguez:
> @cesar-rodriguez Really nice work!
> 
> I took the liberty to remove a dead groovy loop that was bad merged in...
> 
> Also, you had it default to false, but all other linters default to true, so I wanted to keep it on the same level and consistency in action. Hope that isn't a big issue?

Sounds good.

PR #477, opened 2020-07-24
Add PHPStan support

WyriHaximus:
## Proposed Changes

1. Add PHPStan
2. Add support for detecting 

## Readiness Checklist

- [ ] Label as `breaking` if this is a large fundamental change
- [X] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

## Todo list

- [X] Add PHPStan
- [x] Add PHPStan more securely by checking the phars signature
- [x] Add support for detecting PHPStan configuration within the code being checked
- [x] Support to chose one (or more) particular linter(s) for specific language.

Issue #472, opened 2020-07-23
Strategy for multiple linters


GaboFDC:
Right now, we manage 2 linters for javascript (eslint and standard) and python (pylint and flake8). There may be later other languages that need support for multiple linters. 

The current strategy, in a general way, treats the 2 linters as if they were 2 languages. This cause the issue referred below:

> I mean, it would work as it does now, it's just not very convenient because pylint and flake are more or less treated like language switches. Perhaps there could be another configuration option where you can set the preferred linter(s) for JS, python, etc, without triggering that behavior, but that would go beyond the scope of this issue.

_Originally posted by @rolandgeider in https://github.com/github/super-linter/pull/455#issuecomment-663191074_

So the general idea proposed is to have a switch for the language, and an independent switch for the linter

rolandgeider:
Specifically, the behaviour I mentioned in the issue linked above is that currently if a user wants to validate all languages but wants to use flake8 by setting VALIDATE_PYTHON_FLAKE8, it would mean that all other languages will have to be manually switched on as well

admiralAwkbar:
@GaboFDC I think this leads us back to the original way they were disabled.
All were enabled by default, and if you stated `VALIDATE_LANGUAGE=false` it would then omit that language.

The logic feels like it should be:
- No validate flags = all run if found
- validate flag set to true = only validate that language
- validate flag set to false - omit only that language

thoughts?

GaboFDC:
I think that would work fine, reading from #98 I think the problem was that you had to disable manually all linters you were not using, with this logic, you can either explicit define what you lint, lint all or lint all but something specific. 
Also maybe let's hear if @Hanse00 has some comments?

artob:
Would it perhaps make sense to support `VALIDATE_PYTHON=flake8` (instead of `VALIDATE_PYTHON_FLAKE8=true`) and default to Pylint (for compatibility with existing users and previous configuration) if the value is set to `true`?

So as to avoid a combinatorial explosion of distinct environment variables as language and feature support expands over time.

filips123:
> Would it perhaps make sense to support `VALIDATE_PYTHON=flake8` (instead of `VALIDATE_PYTHON_FLAKE8=true`) and default to Pylint (for compatibility with existing users and previous configuration) if the value is set to `true`?

[...]

Issue #122, opened 2020-06-18
PHP Support


lukeholder:
Possible linters:

- https://cs.symfony.com/ (9.1k Stars)
- https://psalm.dev/  (3.1k Stars)
- https://github.com/phpstan/phpstan (8.8k Stars)
- https://github.com/phan/phan (4.8k Stars)

fafnirical:
https://github.com/squizlabs/PHP_CodeSniffer is another very popular tool (7.7K stars, 64.5M installs through [Packagist](https://packagist.org/packages/squizlabs/php_codesniffer))

WyriHaximus:
Laid the foundation for this in #129 with just pure PHP syntax linting

morrisonlevi:
By the way, linting is built into PHP with `php -l`, so all you really need to do is run that across all the PHP files. Looks like the PR tries to do this in parallel. That's probably fine if other linters run in parallel too, but otherwise I'd simplify it to remove the composer dependency, etc.

WyriHaximus:
> By the way, linting is built into PHP with `php -l`, so all the project really needs to do is run that across all the PHP files. Looks like the PR tries to do this in parallel. That's probably fine if other linters run in parallel too, but otherwise I'd simplify it to remove dependencies.

Doesn't matter anyway. Just had a good look at how the linter works and it checks file by file so it can't even utilize the parallelization :(.

fafnirical:
#129 looks like a great foundation for PHP that I think we could build from. `php -l` performs only syntax checking, while many of the other linters in this project run both syntax and stylistic checks

ConradSollitt:
Personally for PHP I'm in favor of using the submitted `php -l` option over the other popular linters by default. Or if supporting multiple linters at least include the `php -l` option. I do see this project allows for multiple linters per language.

The reason is because PHP linters feel very opinionated to me compared to linters in other languages (particaullary JavaScript) and result in a lot of false positives. I ended up using `phpstan` on a project with about ~19,000 lines of PHP code https://github.com/fastsitephp/fastsitephp/blob/master/phpstan.neon and had to add a lot of exceptions to get the code to lint. All of the exceptions were valid as the project has more lines 25,000 lines of code for Unit Testing and specifically tested many of the lines I had to make exceptions for.

Although as I write this it would be nice for my project to lint with both `phpstan` and `php -l` and for many projects they would want one of the other popular linters.

ondrejmirtes:
[...]

PR #478, opened 2020-07-26
Support for PHPCS and Psalm

filips123:
## Proposed Changes

1. Part of #122 to add better PHP linters.
2. Add PHP CodeSniffer as supported linter.
3. Add Psalm as supported linter.

## Description

This adds PHP CodeSniffer and Psalm as supported PHP linters. They are installed using Phive which can automatically verify PHAR GPGP signatures. Supported linters are specified in `depedencies/phive.xml`. Some changes (PHIVE installation) are very similar to #477 so some conflicts will probably have to be resolved.

Note that I renamed `VALIDATE_PHP` to `VALIDATE_PHP_BUILTIN` for using `php -l`, so this is probably breaking change (at least until there is some better solution). I don't know how exactly should multiple linters for one language be handled, but there is already #472 for this which probably needs to be resolved first.

~~It is currently not possible to specify config files, so Psalm won't work because it requires config file. I will address this soon.~~ I also didn't add any test files yet because of this, but will also add them soon.

## TODO

- [x] Add basic support for PHP CodeSniffer and Psalm.
- [x] Add support to specify configuration file and using default one, make Psalm actually work.
- [ ] Add test files to check if linters are working correctly.
- [ ] Decide on strategy for supporting multiple linters for same language (#472).

## Readiness Checklist


santipg382s:
[](url)

admiralAwkbar:
@filips123 great looking work!

WyriHaximus:
@filips123 Awesome :+1: !

admiralAwkbar:
@filips123 looks great! let me know when your ready

filips123:
I will probably add some more PHPCS and Psalm specific tests. It seems `Deploy Docker Image - DEV` is currently just testing with old version which has just built-in PHP linter, so I will also have to check why.

But I don't know what is correct approch for supporting multiple linters. I replaced `PHP` with `PHP_BUILTIN` and added `PHP_PHPCS` and `PHP_PSALM` to support multiple linters. Is this correct for now (until something better is decided in #472)? Should I add `PHP` linter (with `VALIDATE_PHP` env var) back for backward compatibility and if yes, should it just lint code using built-in linter or should it use all supported linters?

WyriHaximus:
> I will probably add some more PHPCS and Psalm specific tests. It seems `Deploy Docker Image - DEV` is currently just testing with old version which has just built-in PHP linter, so I will also have to check why.
> 
> But I don't know what is correct approch for supporting multiple linters. I replaced `PHP` with `PHP_BUILTIN` and added `PHP_PHPCS` and `PHP_PSALM` to support multiple linters. Is this correct for now (until something better is decided in #472)? Should I add `PHP` linter (with `VALIDATE_PHP` env var) back for backward compatibility and if yes, should it just lint code using built-in linter or should it use all supported linters?

Yeah was looking into the same thing, just pushed some changes that should test with PHPStan: https://github.com/github/super-linter/pull/477/files#diff-409b55ad463cfac3e14fdaa2a919cfebR602

filips123:
I have problem with testing PHPCS and Psalm. They both include information about used memory and time for tests in the output. I can't just include this in TAP file, because they will be different every time. Is is possible to tell tester that this part of output will be different every time and should not affect test results?

WyriHaximus:
@filips123 is there no way to disable that by using a different output?

filips123:
At first glance, it seems not. However, time and memory are displayed in last few (different in PHPCS and Psalm) lines. Is it possible to just discard last few lines of output?

lib/linter.sh
```diff
@@ -1453,12 +1467,28 @@ fi
...
+  # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILE_ARRAY"
+  LintCodebase "PHP_BUILTIN" "php" "php -l" ".*\.\(php\)\$" "${FILE_ARRAY_PHP_BUILTIN[@]}"
+fi
+
+if [ "${VALIDATE_PHP_PHPCS}" == "true" ]; then
+  ############################################
+  # Lint the PHP files using PHP CodeSniffer #
+  ############################################
+  # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILE_ARRAY"
+  LintCodebase "PHP_PHPCS" "phpcs" "phpcs --standard=${PHP_PHPCS_LINTER_RULES}" ".*\.\(php\)\$" "${FILE_ARRAY_PHP_PHPCS[@]}"
```
WyriHaximus, line 1483:
```suggestion
  LintCodebase "PHP_PHPCS" "phpcs" "phpcs --standard=${PHP_PHPCS_LINTER_RULES} | grep -v FILTERWHATYOUWANTTOFILTER" ".*\.\(php\)\$" "${FILE_ARRAY_PHP_PHPCS[@]}"
```

WyriHaximus:
Well giving we're still shell scripting, you can probably pipe it into a `grep -v` as suggested here to filter it out: https://github.com/github/super-linter/pull/478/files#r461780650

lib/linter.sh
filips123: Shouldn't this be added to `lib/worker.sh` with `TestCodebase`, as that is command/file which is used in tests?

But it doesn't seem it is working. I have

```sh
  TestCodebase "PHP_PHPCS" "phpcs" "phpcs --standard=${PHP_PHPCS_LINTER_RULES} | grep -v 1" ".*\.\(php\)\$" "php"
  TestCodebase "PHP_PSALM" "psalm" "psalm --config=${PHP_PSALM_LINTER_RULES} | grep -v 2" ".*\.\(php\)\$" "php"
```

But this will fail with

```
ERROR! Found errors in [phpcs] linter!
ERROR:[ERROR: The file "|" does not exist
```

I also tried this in `lib/linter.sh` but then it doesn't have any effect.

WyriHaximus: Ah yes my bad, hmm so doesn't support piping from one command to another. Which shell is used to run this? 😱 

Sorry I just assumed this would work.

admiralAwkbar:
thatw as some gnarly merge conflicts for some reason... should have got it all worked out :)

admiralAwkbar:
@filips123 you can just add `.ignored` to the taps and skip for now, we need to come back and really make that much better :)

filips123:
Thanks! I will probably fix what needs to be fixed next week because I don't currently have time.

admiralAwkbar:
@filips123 Fixed the dep issues... I think your good to finish anything you had left :)

filips123:
Ok. I will fix few things that became messy with merge of PHPStan PR and make those ignored tests.

Where should ` .ignored` be added? Should it just be the only content of TAP file for that linter or somewhere else?

filips123:
I added linters to README and made that `VALIDATE_PHP` is the same as `VALIDATE_PHP_NUILTIN`. Now I think just tests (`expected-PHP_PHPCS.tap` and `expected-PHP_PSALM.tap`) need to be fixed/ignored.

admiralAwkbar:
@filips123 Are we ready to get this out of the draft and merged up?

filips123:
Yes, I will just wait until workflow finishes.

filips123:
Ok, I found two errors:

1. PHPStan TAP should not be ignored but instead fixed because I changed `php_bad_2.php` so output is different.
2. PHPStan fails because of `Uncaught TypeError: Return value of _HumbugBoxc7aa196c4c1e\Nette\DI\Config\Adapters\PhpAdapter::load() must be of the type array, int returned`.

I will check this later today.

admiralAwkbar:
@filips123 

Any ideas on this?

https://github.com/github/super-linter/pull/478/checks?check_run_id=945062540#step:4:426

admiralAwkbar:
@WyriHaximus Just ignoring the output for the tests as they have time info and will not likely work well...

admiralAwkbar:
@filips123 good catch on the linter rules!

admiralAwkbar:
@filips123 i think this is ready to get out of the draft after this build :)

WyriHaximus:
:tada:

Dockerfile
```diff
@@ -65,8 +65,8 @@ RUN apk add --update --no-cache \
     npm nodejs-current \
     openjdk8-jre \
     perl \
-    php7 php7-phar php7-json php7-mbstring \
-    php7-tokenizer php7-ctype php7-curl php7-dom \
+    php7 php7-phar php7-json php7-mbstring php-xmlwriter \
```
divinity76, line 68: shouldn't it be php7-xmlwriter ? at least in debia 10 buster it's called `php7.3-xmlwriter` (there's also a php-xmlwriter alias though)

WyriHaximus: It probably should be for explicity, but `php-*` is an alias to `php7-*` by now I hope 😂

divinity76: @WyriHaximus for now - php8 is approaching, scheduled for release in november 2020

filips123: Should I create another PR to fix this (so all of them use php7) or can this be commited directly to repository?

WyriHaximus: > @WyriHaximus for now - php8 is approaching, scheduled for release in november 2020

Yup, can't wait! But it's not going to be an issue until they bump alpine versions.

> (but this mixed version thing isn't good imo, either all of them should explicitly mention php7, or none of them should)

With future proofness in mind, maybe they should be `php` and not `php7` since most of those tools will be compatible (or at least 95%) with PHP 8 months before it is released.

WyriHaximus: @filips123 If you don't I will, but I would go for `php` instead so it doesn't have to be changed when php8 langs in Alpine :smiley: .

filips123: > compatible (or at least 95%) with PHP 8 months before it is released

Well, PHP 8 will be released on November 26, 2020, so we are already few months before release. But based on their `php` requirement in `composer.json`, they should already support PHP 8.

>  but I would go for php instead

Yes, maybe this would be better.

WyriHaximus: > > compatible (or at least 95%) with PHP 8 months before it is released
> 
> Well, PHP 8 will be released on November 26, 2020, so we are already few months before release. But based on their `php` requirement in `composer.json`, they should already support PHP 8.
> 

Exactly, they tend to make sure everything runs smoothly by the time a new major or minor version is released. (This is also why I switched by to PHPCS from PHPCSFixer a while ago. The latter could take weeks before supporting a new minor version.)

> > but I would go for php instead
> 
> Yes, maybe this would be better.

:+1:

PR #486, opened 2020-07-27
Log

### LINKED ISSUE #485: OUTPUT_LOG option
Browsing linter errors and warnings in the Actions GUI is not a pleasant or efficient experience. 

Something like an `OUTPUT_LOG` or `LOG_FILE` option that adds the `super-linter` output to a logfile. The case this would benefit is the ability to use `upload-artifact` to upload the linter logs as an artifact that could be downloaded directly or through another workflow/job and/or integration (for example, posting the log to a slack channel). 

Yes, "raw logs" options in a job enables a suboptimal workaround for this functionality, requiring 4 clicks and wading through all the other the step outputs (this is what I'm doing now) 

Abandon `super-linter` entirely and initialize linters through `bash` so that a logfile can be generated, for example: 
```bash
pylint | tee "pylint.log"
```

nemchik:
Fixes #485

## Proposed Changes

1. Add log functions so that we have different levels of logging, automatic formatting (colors), timestamps, and logging happens to the console (based on user defined verbosity) and a log file (regardless of verbosity - although we can consider user defined verbosity here as well)
2. Any output that the user should see (almost all `echo` calls) should use the log functions
3. Allow the user to define verbosity for the console output (default to full verbosity and allow users to lower it with a variable)

### Log Functions Added

- `fatal` will display a fatal message (red background, white text) AND will `exit 1`. Always displayed regardless of `LOG_LEVEL`
- `error` will display an error message (red text). Always displayed regardless of `LOG_LEVEL`
- `warn` will display a warn message (yellow text). Always displayed regardless of `LOG_LEVEL`
- `notice` will display a notice message (green text). Always displayed regardless of `LOG_LEVEL`
- `info` will display an info message (blue text). Only displays if 'LOG_LEVEL` is set to `VERBOSE`
- `debug` will display a debug message (blue text). Only displays if 'LOG_LEVEL` is set to `DEBUG`
- `trace` will display a trace message (blue text). Only displays if 'LOG_LEVEL` is set to `TRACE`

All messages (regardless of `LOG_LEVEL` are output to the `LOG_FILE`

## Readiness Checklist

## Remaining Tasks

- [x] Replace all lines that currently produce loggable output with log functions (so the log file is used)
- [x] Add the code to allow setting the verbosity level
- - [x] Default to full verbosity (output everything)
- - [x] Replace existing `notice` calls with `info` calls as needed
- [x] Document the new user facing variables


admiralAwkbar:
Looks good

zkoppert:
hey just checking in, Can you check off the boxes that you completed in the description.

nemchik:
> hey just checking in, Can you check off the boxes that you completed in the description.

Done, and added one last task unless anyone looks through this sooner and finds something else that should be adjusted.

I figure we may over time move around what is output as `fatal`, `error`, `warn`, `notice`, `info`, `debug`, and `trace`. Speaking of, I'll document in the OP a little about each of them.

lib/worker.sh
```diff
@@ -567,11 +562,11 @@ function RunTestCases() {
   #################
   # Header prints #
   #################
-  echo ""
-  echo "----------------------------------------------"
-  echo "-------------- TEST CASE RUN -----------------"
-  echo "----------------------------------------------"
-  echo ""
+  echo
```
zkoppert, line 565: I'm a little confused about why some things have the `echo` before and after and others don't. Can you help me understand that part?

nemchik: `echo ""` is the same as `echo` (it just prints a blank line).

I removed some of them because they are surrounded by log functions that would only output if the `LOG_LEVEL` is set high enough. I may need to comb through this a bit more diligently and either make some adjustments or just remove them all.

nemchik: P.S. I'm leaning towards removing them all because right now they are not conditionally output (like the log functions) and we can end up with multiple consecutive blank lines for seemingly no reason:
image

nemchik:
Ok, documentation is added. This should be ready to merge.

PR #517, opened 2020-08-04
Fix Hadolint

admiralAwkbar:
this fixes what i forgot to merge...


admiralAwkbar:
@GaboFDC Good catch... my bad for missing

PR #528, opened 2020-08-06
Adding SQL-Lint

### LINKED ISSUE #128: SQL Support
we have a lot of SQL in our codebase and would be nice to lint these as well.

not sure which linter would be the best

admiralAwkbar:
This PR will cover:
- Adding `sql-lint` to the codebase
- Being able to rest and lint SQL files

This closes #128


admiralAwkbar:
Seems this is all set and should work as expected. Reached out to `sql-lint` stream to see if they could add `-c` flag and then when added, we can update code to allow for configurations

joereynolds:
Hey all,
sql-lint dev here.
I've tagged the release that adds the option for `-c`/`--config` flags (v0.0.15)
Let me know if there are any problems, enjoy!

PR #529, opened 2020-08-06
Adding all codebase scan test

### LINKED ISSUE #475: Create test for validate all code base
There has been a lot of bugs and errors not caught because the test only run each linter individually.

There should be a test that runs the linter as a whole, and validate all code base, to test this case.

Running latest image against my repos

**Additional context**
I was thinking we may add a workflow that runs the built container against the `.automation/test` folder, and checks the output, it should lint all languages, and fail on the bad ones. But not really sure about the way to do this, maybe save the run output to a file, and compare it with an expected file? this was just my first thought but doesn't seem too maintainable, or maybe we could just grep the output for specific contents? 

I'm really looking forward for other thougths on this.

admiralAwkbar:
This PR will add:
- Automation to test the codebase for all files
- It will follow normal testing, then clean codebase of bad files, then run against all files

This will increase the testing time, but will greatly help reduce errors that are introduced to the tool

This closes #475

### DIFF
--- a/.automation/clean-code-base-for-tests.sh
+++ b/.automation/clean-code-base-for-tests.sh
@@ -151,7 +151,7 @@ CleanSHAFolder() {
   ##################
   # Find the files #
   ##################
-  REMOVE_CMD=$(cd "${GITHUB_WORKSPACE}" || exit 1; rm -rf "${GITHUB_SHA}" 2>&1)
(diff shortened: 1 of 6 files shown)


admiralAwkbar:
@GaboFDC @nemchik any idea what's going on here? Not sure why mapping is all messed up...

admiralAwkbar:
The clean code base doesnt really show output, and the next step complains about the workspace? seems lost

admiralAwkbar:
@michaelsainz any help with PowerShell and errors is super appreciated...

PR #530, opened 2020-08-06
Add elisp linter

ericdallo:
Fixes #

## Proposed Changes

1. ...
2. ...
3. ...

### Author/Contributor

### Reviewing Maintainer 


admiralAwkbar:
@ericdallo looks like you need to create the tests folder:
`.automation/test/emacs`

And then add a good and bad test case and we should be in good shape?

ericdallo:
Yeah, I was intending to do that after fixing the problem with emacs container not working correctly on this docker python:alpine image :/

PR #534, opened 2020-08-07
PerlCritic basic implementation

### LINKED ISSUE #232: Suggestion: use Perlcritic for perl
Perlcritic has 5 levels of code quality, with a reference manual behind it: Perl Best Pracices, from O'Reilly publishers.

marghidanu:
Use Perl::Critic to lint Perl files
https://github.com/github/super-linter/issues/232

Fixes #232

### Author/Contributor
- Perl::Critic documentation available here: https://metacpan.org/pod/Perl::Critic

### Reviewing Maintainer 


admiralAwkbar:
@marghidanu nice work :) just clean up that test case and I think we're in good shape :)

Really nice work on getting it added!

marghidanu:
Fixing it now, I'm also trying to include the results of Perl compilation checks.

marghidanu:
Unfortunately, there's no policy for the type of check that was previously used. Anway using `perl -c` is a big no-no because in some cases it can actually execute code (BEGIN blocks get evaluated during compile time), also it will try to load all the packages imported.

admiralAwkbar:
@marghidanu Awesome work!

PR #548, opened 2020-08-10
Only run deploy to Docker on original repo and not forks

bazzadp:
## Proposed Changes

This repo has [an action which runs on pushes to master to build a Docker image](https://github.com/github/super-linter/blob/master/.github/workflows/deploy-PROD.yml). When you fork this repo it also runs on pushes to master, and then fails as the forks don't have the appropriate secrets set.

This PR limits the action to just the `github/super-linter` repo.

An alternative would be to check if the `secrets.DOCKER_USERNAME` is set but I believe that can only be done at step level not job level whereas limiting based on repo name will exit the action earlier.

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request - N/A

### Reviewing Maintainer 


bazzadp:
> Do we want to hardcode the repository name though? Perhaps there's a more elegant solution?

It's a fair point, but I'd vote to hard code it for simplicity sake. It's obvious what that line does (especially with the comment) where as a more elegant solution may not be more readable.

Changing the repo name is not something that will happen often (if at all?) and there are plenty of other places where this is hardcoded too so any change would require a big search and replace anyway. Not that that is an excuse or that we should add to that list is a good idea, but just showing it won't be easy to migrate a repo.

However if someone has an elegant solution that doesn't compromise the simplicity too much then let me know. From a search all I found was this one: https://github.community/t/stop-github-actions-running-on-a-fork/17965 which is the solution I implemented.

bazzadp:
Also should we do this for any other workflows too?

I got a timeout from my fork last night for `workflows/deploy-DEV.yml` - does it need to run in my fork?

What about others?

Or do we merge this first, monitor it for a while in case of issues, then work on a complete list?

nemchik:
Any pull request that targets this repo (including PRs from forks) will meet the `github.repository == 'github/super-linter'` condition and will run with the context of whatever repo it came from (forks won't have secrets).

You may want the `pull_request_targe` event. That'll use this repo as context even if the PR was opened from a fork.

bazzadp:
Good point - and that affects `deploy-DEV-PR.yml` - though not `deploy-DEV.yml` nor `deploy-PROD.yml`.

So think we should leave `deploy-DEV-PR.yml` and `deploy-DEV.yml` to run on forks (they should work - not sure why it timed out for me last night). That also builds, tests and lints the code - like we should do.

So we just block `deploy-PROD.yml` in forks because it will never work. This is what is currently set in this PR. Only issue is if someone's set up DOCKER secrets and want it to run from fork but can't see why you'd want that.

WDYT?

admiralAwkbar:
@bazzadp The issue was trying to solve a logic problem on the dev deploy and we created 2 jobs, but that wasn't the right answer... So i removed the PR and it should go back to running more normally. I need to update the automerge to fix a deps issue.

PR #577, opened 2020-08-17
docs: remove note about using docker://

ahmadnassri:
it seems GitHub Actions now respects `actions.yml` directives when using docker actions, thus making the note to use `docker://` redundant.

in other words: using `github/superlinter@v3` no longer results in compiling the docker image, since `action.yml` has been updated 

https://github.com/github/super-linter/blob/5c75c2fdcd47301590fdd6c0c105285b8537a0dc/action.yml#L4-L6

## Proposed Changes

simplify docs, and use the standard gh action format of `github/super-linter@v3`

### Author/Contributor

### Reviewing Maintainer 

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,15 +148,12 @@ jobs:
       # Run Linter against code base #
       ################################
       - name: Lint Code Base
-        uses: docker://github/super-linter:v3
+        uses: github/super-linter@v3
         env:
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@ahmadnassri Very nice! I didn't know they finally got that worked out... seems easier...

PR #592, opened 2020-08-19
Fix minor typo in HTML language name

bondms:


admiralAwkbar:
@bondms thanks for the catch...

PR #611, opened 2020-08-24
add quotes

### LINKED ISSUE #602: Error while getting Lua linter version
**Describe the bug**
Lua linter version is not reported correctly.

**To Reproduce**

```shell
docker run \
    -v "$(pwd)":/workspace \
    -w="/workspace" \
    -e ACTIONS_RUNNER_DEBUG=true \
    -e DEFAULT_WORKSPACE=/workspace \
    -e MULTI_STATUS=false \
    -e RUN_LOCAL=true \
    github/super-linter:V3.8.0

```
**Expected behavior**

**Screenshots**
Verbose output (seen in super-linter v3.8.0):

```shell
[lua]: lua: unrecognized option '--version' usage: lua [options] [script [args]] Available options are:   -e stat  execute string 'stat'   -i       enter interactive mode after executing 'script'   -l name  require library 'name' into global 'name'   -v       show version information   -E       ignore environment variables   --       stop handling options   -        stop handling options and execute stdin
```

**Additional context**
This is not the first time this happens, see #543.  I highly suggest throwing an error and terminating the script, because if you cannot be sure of the linter version you're using, you might get inconsistent results.

### LINKED ISSUE #604: Superlinter incorrectly reports languages linted
**Describe the bug**
The super-linter API over reports languages linted when adding the status to a pull request.

For example a recent PR in my repo has the following:
Super-Linter Test Results
I have no Java or R code in my repo.

I noticed it particularly since the R linter was added and have a theory.

What I suspect is happening is to do with this line of code:

https://github.com/github/super-linter/blob/0e075ccba96327a6d5dec6cc264ad7c06862e352/lib/linter.sh#L1111

which then goes on to call CallStatusAPI.

I suspect it's saying "JAVA" is in "**JAVA**SCRIPT_ES", and "R" is in  "JAVASC**R**IPT_ES" and also "MA**R**KDOWN", hence why they are being include for reporting (with 0 errors since they weren't actually run).

[This Stackoverflow answer](https://stackoverflow.com/a/15394738/2144578) gives a better method to test if a value is in an array by adding spaces on either side.

**To Reproduce**

1. Lint a repo with no R files but with markdown files (README.md will do)
2. Open a PR with an edit to the README.md
3. Note that R is included as a linted language

**Expected behavior**
The linter should only list languages actually linted. Especially when you explicitly enable only the languages you want (as I do) and R is not enabled.

**Screenshots**
As per above

**Additional context**
Theory given above.

admiralAwkbar:
This closes #604
This closes #602 

This should help parse through the array with less false positives


lib/linter.sh
```diff
@@ -1144,7 +1148,31 @@ Footer() {
...
 
+  ######################################
+  # Check if Language was in the array #
+  ######################################
+  for LANG in "${UNIQUE_LINTED_ARRAY[@]}"; do
+    if [[ "${LANG}" == "${NEEDLE}" ]]; then
+      ############
+      # Found it #
+      ############
+      return 0
```
bazzadp, line 1167: Maybe I’m misunderstanding this but why is this returning 0 when it’s found and 1 when it’s not found? Isn’t that the wrong way around based on the function name and how it’s used above?

PR #666, opened 2020-09-02
fix: comment typo for GH Container Registry

f0rmiga:
## Proposed Changes

1. Fixes the typo on `GPR`. I believe the original author wanted to type `GCR`.

My proposed change doesn't use `GCR` as this acronym has been generally used by the Google Container Registry. If you guys still want to use acronyms, then I suggest using `GHCR` (from your ghcr.io domain) to avoid conflicts with any existing accepted acronyms in the community.

I didn't touch other parts of the code that use `GCR` as an acronym but I can certainly do if you believe it's the right direction.

### Author/Contributor

### Reviewing Maintainer 

### DIFF
--- a/.github/workflows/deploy-RELEASE.yml
+++ b/.github/workflows/deploy-RELEASE.yml
@@ -54,9 +54,9 @@ jobs:
         shell: bash
         run: .automation/upload-docker.sh
 
-      #############################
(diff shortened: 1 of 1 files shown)


f0rmiga:
The PR number is 666 but I guarantee there's no evil in it! :joy:

PR #675, opened 2020-09-03
Making MULTI_STATUS honor DISABLE_ERRORS setting

dmerrick:
Fixes unexpected behavior when `DISABLE_ERRORS=true`

If a user sets `DISABLE_ERRORS=true`, they are trying to make all linting checks succeed, such that PRs can be merged even with linter failures. Unfortunately `MULTI_STATUS` doesn't respect this setting, causing builds to go red even if `DISABLE_ERRORS=true`. (To make matters worse, MULTI_STATUS is enabled by default, so right now a user will have to explicitly disable MULTI_STATUS in order for DISABLE_ERRORS to work)

This adds a bit of logic into the `MULTI_STATUS` feature which will prevent the individual statuses from failing the build when `DISABLE_ERRORS=true`.

### Author/Contributor

### Reviewing Maintainer 

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -988,17 +988,18 @@ CallStatusAPI() {
   else
     # Failure
     MESSAGE="${FAIL_MSG}"
-
-    # make sure we honor DISABLE_ERRORS
-    if [ "${DISABLE_ERRORS}" == "true" ]; then
-      STATUS="success"
-    fi
   fi
 
   ##########################################################
   # Check to see if were enabled for multi Status mesaages #
   ##########################################################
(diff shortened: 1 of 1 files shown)


admiralAwkbar (approved):
I agree with the logic!

dmerrick:
Ty for approving, @admiralAwkbar!

Any idea when we can expect the next release to be cut?

admiralAwkbar:
Hey! I can cut one Monday if that's ok
Trying to get another small fix in as well

dmerrick:
Understood, and that'd be great. Ty for being responsive!

Issue #672, opened 2020-09-03
Failed to find [dotnet-format] in system


v1v:
**Describe the bug**

```
[2020-09-03T15:26:37.627Z] 2020-09-03 15:26:37 [INFO  ]   ----------------------------------------------
[2020-09-03T15:26:37.627Z] 2020-09-03 15:26:37 [ERROR ]   Failed to find [dotnet-format] in system!
[2020-09-03T15:26:37.627Z] 2020-09-03 15:26:37 [FATAL ]   []
script returned exit code 1
```

**To Reproduce**

1. Go to https://apm-ci.elastic.co/blue/rest/organizations/jenkins/pipelines/apm-shared/pipelines/apm-pipeline-library-mbp/branches/master/runs/1023/nodes/231/steps/243/log/?start=0
2. Scroll down to the very end
3. See error

command:

```
docker run -e RUN_LOCAL=true \
            -e DISABLE_ERRORS=true \
            -e OUTPUT_FORMAT=tap \
            -e OUTPUT_DETAILS=detailed \
            -e OUTPUT_FOLDER=.super-linter \
            -v ${PWD}:/tmp/lint \
            github/super-linter:latest
```

**Screenshots**

![image](https://user-images.githubusercontent.com/2871786/92137378-f66f7180-ee04-11ea-9b0b-9d98f7dd5c4e.png)

**Additional context**

Might be related to https://github.com/github/super-linter/commit/6c25dc24f0a58fd2973d77aa677c7a7e191b1e83

admiralAwkbar:
@v1v 

Super odd..

I just downloaded the latest and tried:
```bash
~/github/super-linter on  master ⌚ 12:53:05
$ docker run -it --entrypoint /bin/bash github/super-linter:latest                                                                                                                                 2.5.3
Unable to find image 'github/super-linter:latest' locally
latest: Pulling from github/super-linter
df20fa9351a1: Already exists
36b3adc4ff6f: Already exists
3e7ef1bb9eba: Already exists
78538f72d6a9: Already exists
1692f1a2776b: Already exists
aa495c978777: Pull complete
1bc58411f080: Pull complete
7977756abeee: Pull complete
bcc33850e438: Pull complete
57372e2f7735: Pull complete
e13c7e768328: Pull complete
53dd6c007e52: Pull complete
fc672017caa9: Pull complete
a3737bb17dcb: Pull complete
fcd863b0c2c4: Pull complete
1b0679b62096: Pull complete
[...]

PR #677, opened 2020-09-04
Print version, date and revision

v1v:
Expose docker build arguments as environment variables to expose them when printing the header, the idea is to help to debug what's the version that has been used in case something bad happens (see https://github.com/github/super-linter/issues/672).

It has been added the git sha commit as a docker label metadata and also in the header

### Author/Contributor
- ~[ ] If documentation is needed for this change, has that been included in this pull request~

### Reviewing Maintainer 


admiralAwkbar:
@v1v Amazing work!
Just added a few small tweaks but again, super cool stuff

admiralAwkbar (approved):
so dope

PR #679, opened 2020-09-04
Just filter the .git folder

Goryudyuma:
## Proposed Changes

1. Linter doesn't look under .git/ as before.
Unlike before, linter looks under .github/ .
(.github/workflows/ etc.)

### Author/Contributor

### Reviewing Maintainer 


admiralAwkbar:
@Goryudyuma Good catch on this!

admiralAwkbar (approved):
Such fix... much wow...

Goryudyuma:
https://github.com/github/super-linter/pull/679/checks?check_run_id=1073302462#step:8:261

This is what I get instead of `${FILE} == *".git/"*` but `${FILE} == *".git" || ${FILE} == *".git/"*` That's why

Goryudyuma:
I don't know why, but .git is the target.

Goryudyuma:
So, please revert [`828de13` (#679)](https://github.com/github/super-linter/pull/679/commits/828de134a98547086dc702267c11893e90d408ed) or fix it
@admiralAwkbar

Goryudyuma:
Hmm..
I only want to exclude .git and .git/.., anything else is fine.
It's okay to have only .github in it, but will we be able to respond whenever we find something that starts with .git?

admiralAwkbar:
@Goryudyuma I thin this last update is right.
Now it excludes the folder for `.git` and not `.github`
Also updated the logic to look only at the file path, and exclude file names for any possible name collisions

Goryudyuma:
Last but not least, do you realize that if we start adding more .git folders in the future, we need to respond?
(.gitfoo/ ,　.gitbar/ ...)

I don't want to support cases where .git folders contain things that aren't related to git. I think it's evil.

Goryudyuma:
The test is failing...

Goryudyuma (commented):
I suggest reverting to the first change.
`${FILE} == *".git" || ${FILE} == *".git/"*`

The reason why the first change failed the test has already been fixed.
https://github.com/github/super-linter/pull/681

When I revert to the first change, linter don't see .git/, but .github/ and other .gitfoo/, .gitbar/ etc. that start with .git are watched.
Perhaps the test will pass.

how is it?

lib/worker.sh
```diff
@@ -160,19 +160,19 @@ function LintCodebase() {
...
-      if [[ ${FILE} == *"node_modules"* ]]; then
+      if [[ ${DIR_NAME} == *"node_modules"* ]]; then
         # This is a node modules file
         continue
-      elif [[ ${FILE} == *"${TEST_CASE_FOLDER}"* ]]; then
+      elif [[ ${DIR_NAME} == *"${TEST_CASE_FOLDER}"* ]]; then
         # This is the test cases, we should always skip
         continue
-      elif [[ ${FILE} == *".git"* ]]; then
+      elif [[ ${DIR_NAME} == *".git"* ]] || [[ ${DIR_NAME} == *".git/"* ]]; then
```
Goryudyuma, line 169:
```suggestion
      elif [[ ${DIR_NAME} == *".git" ]] || [[ ${DIR_NAME} == *".git/"* ]]; then
```

`*".git"*` is include .github/...

Goryudyuma:
https://github.com/github/super-linter/pull/697
DIR_NAME -> FILE

The test passed.

admiralAwkbar (approved):
better

Goryudyuma:
Thank you @admiralAwkbar !

PR #695, opened 2020-09-08
added import-error to the pylint disable rules

### LINKED ISSUE #157: How to install Python dependencies
In https://github.com/yhoiseth/python-prediction-scorer/runs/786096970, pylint is complaining about missing imports. They are already installed in a previous step, but it doesn’t seem like they are “kept”. How can I run e.g. `python -m pip install --upgrade pip pip-tools setuptools wheel && pip-sync` so that it works with Super-Linter?

### LINKED ISSUE #645: Pylint should ignore import errors
Pylint regularly fails for import errors on modules. This isn't really a linting problem (IE something wrong with your code) so I feel that it's safe to ignore. 

In https://github.com/github/super-linter/blob/master/TEMPLATES/.python-lint we can add `import-error` to the `disable=` section.

We currently have to add this line to every module and it would be easier to globally ignore these errors.

**Additional context**
I can add this feature if this is approved.

Thanks

primetheus:
Fixes #157 
Fixes #645

## Proposed Changes

1. ignore `pylint import` errors

### Author/Contributor

### Reviewing Maintainer 

### DIFF
--- a/TEMPLATES/.python-lint
+++ b/TEMPLATES/.python-lint
@@ -127,7 +127,8 @@ disable=print-statement,
         next-method-defined,
         dict-items-not-iterating,
         dict-keys-not-iterating,
-        dict-values-not-iterating
+        dict-values-not-iterating,
+        import-error
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option


admiralAwkbar (approved):
dank

PR #706, opened 2020-09-10
Set GROOVY Linter failure level to warning

nvuillam:
## Proposed Changes

1. Set GROOVY Linter failure level to warning
I noticed that most of linters fail once there is a 'info' level error. I think it's not relevant to fail for an indentation rule fail, but instead of current **error** level, **warning** seems more appropriate

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
agree we need to fix these issues, not shy away from them

PR #707, opened 2020-09-10
Fix list files function typos

### LINKED ISSUE #704: Super-Linter has started linting node_modules as part of v3.10.0
**Describe the bug**
Just got a dependenabot note about v3.10.0 being released. Linting failed in that pull request due to Super-linter now linting the `node_modules` folders and many npm developers not following our HTML conventions.

Note we do not check in `node_modules`to our repo, however we do generate HTML as part of our build process (which requires `npm install` to have run), so run an HTML Lint on that (see [our workflow here](https://github.com/HTTPArchive/almanac.httparchive.org/actions/runs/247486234/workflow)) and this threw the error, which did not happen on v3.9.4

**To Reproduce**

1. Use a repo that uses node
2. Kick off a full HTML linting (not just changed files) using v3.10.0 but also running `npm install` in the action before.
3. Notice from the log files that all `node_module` HTML files are linted - and many probably failed.

**Expected behavior**
I would expect only my source code to lint - like they did prior to v3.10.0

**Screenshots**
See example log file here: https://github.com/HTTPArchive/almanac.httparchive.org/pull/1282/checks?check_run_id=1094735802

**Additional context**
I presume this is due to #696 and could be fixed with the recently merged #684 ? If so can we have a v3.10.1 or v3.11.0 release with that?

ferrarimarco:
Fixes #704

## Proposed Changes

1. Fix typos in the `find` query when building the list of files to lint.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -773,8 +773,8 @@ function PopulateShellScriptsList() {
   mapfile -t LIST_FILES < <(find "${GITHUB_WORKSPACE}" \
     -path "*/node_modules" -prune -o \
     -path "*/.git" -prune -o \
-    -path "*/.venv"-prune -o \
-    -path "*/.rbenv"-prune -o \
+    -path "*/.venv" -prune -o \
+    -path "*/.rbenv" -prune -o \
     -type f 2>&1)
   for FILE in "${LIST_FILES[@]}"; do
     if IsValidShellScript "${FILE}"; then
--- a/lib/worker.sh
+++ b/lib/worker.sh
@@ -180,19 +180,19 @@ function LintCodebase() {
       ######################################################
       # Make sure we don't lint node modules or test cases #
       ######################################################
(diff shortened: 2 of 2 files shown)


bazzadp:
Looks like this was broken as part of https://github.com/github/super-linter/commit/6040030c8ac8aa95ceb76b1df4b808f61cabccd8 so should we also fix `lib/buildFileList.sh` as part of this as looks like it had a similar typo in that commit?

ferrarimarco:
I don't see that typo in the latest master in `buildFileList.sh`

admiralAwkbar (dismissed):
sorry

bazzadp:
Note this doesn't fix #704 as discussed in this. But should commit this anyway!

admiralAwkbar:
@bazzadp It does now :)

admiralAwkbar (approved):

PR #713, opened 2020-09-11
Fix file detection issues of Snakemake linter/formatter

kpj:
This PR makes the Snakemake linter/formatter work by fixing issues with the file detection.

In particular, it does the following:
* fix bug in regex
* initialize `FILE_ARRAY_SNAKEMAKE` properly

Please let me know if there are any other things to repair 😄


admiralAwkbar (dismissed):
very nice catches

admiralAwkbar:
@kpj Interesting, it didn't seem to exit with error code...
It showed the error found, but I'm guessing it still exited `0`

admiralAwkbar:
@kpj issue was we were locked on 0.1.4 of snakefmt with that error, updated to 0.1.5

admiralAwkbar:
@kpj so it seems the root of the issue:
- We used to be able to use `--diff` and `--check`
- That would give us a proper error code when checking the file.
Now both return a `0` error code
We need to reach out to snakefmt to see if they can give proper error code when a file is changed, or allow for both flags to be executed again

```bash
bash-5.0# snakefmt  --check --diff --config action/lib/.automation/.snakefmt.toml bad.smk
[WARNING] Both --check and --diff/--compact-diff given. Only running --check...
[INFO] 1 file(s) would be changed 😬
bash-5.0# echo $?
1
bash-5.0# snakefmt  --diff --config action/lib/.automation/.snakefmt.toml bad.smk
=====> Diff for bad.smk <=====

  rule all:
      input:
-         file1='result.txt',
?               ^          ^
+         file1="result.txt",
?               ^          ^
+

  rule simulation:
      output:
-         file1="result.txt"
+         file1="result.txt",
?                           +
      shell:
          """
          touch {output}
          """
-

[INFO] All done 🎉
bash-5.0# echo $?
0
bash-5.0#
```

kpj:
> @kpj issue was we were locked on 0.1.4 of snakefmt with that error, updated to 0.1.5

Yes, I reported the issue (https://github.com/snakemake/snakefmt/issues/65) and now everything should be fine. Thanks for updating the version 😃 


> * We used to be able to use `--diff` and `--check`
> * That would give us a proper error code when checking the file.
>   Now both return a `0` error code
>   We need to reach out to snakefmt to see if they can give proper error code when a file is changed, or allow for both flags to be executed again

You are right, I have submitted another issue: https://github.com/snakemake/snakefmt/issues/68

kpj:
Now that `snakefmt` has been updated to version 0.2.0, formatting works nicely.

Another issue I noticed is that `snakemake --lint` sometimes crashes because it is not executed in the expected working directory and cannot e.g. read configuration files. However, this seems like something which should be fixed on Snakemake's side, so we cannot do much about it here. What do you think?

admiralAwkbar (approved):
solid

lib/linter.sh
```diff
@@ -1944,8 +1945,8 @@ if [ "${VALIDATE_SNAKEMAKE_SNAKEFMT}" == "true" ]; then
   ################################
   # Lint the files with snakefmt #
   ################################
-  # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILE_ARRAY" "FILTER_REGEX_INCLUDE" "FILTER_REGEX_EXCLUDE"
-  LintCodebase "SNAKEMAKE_SNAKEFMT" "snakefmt" "snakefmt --config ${SNAKEMAKE_SNAKEFMT_LINTER_RULES} --check --compact-diff" ".*\(Snakefile\|\.smk\)\$" "${FILE_ARRAY_SNAKEMAKE[@]}" "${FILTER_REGEX_INCLUDE}" "${FILTER_REGEX_EXCLUDE}"
+  # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILTER_REGEX_INCLUDE" "FILTER_REGEX_EXCLUDE" "FILE_ARRAY"
+  LintCodebase "SNAKEMAKE_SNAKEFMT" "snakefmt" "snakefmt --config ${SNAKEMAKE_SNAKEFMT_LINTER_RULES} --check --compact-diff" "\(Snakefile|.*\.smk\)\$" "${FILTER_REGEX_INCLUDE}" "${FILTER_REGEX_EXCLUDE}" "${FILE_ARRAY_SNAKEMAKE[@]}"
```
kpj, line 1949: Doesn't `"\(Snakefile|.*\.smk\)\$"` break the file detection?

PR #714, opened 2020-09-11
Fixed a typo in linter.sh code comments

josephmarino:
Hi folks :wave:

I was reading linter.sh to learn more about how super-linter works, and I found a typo in the code comments for the ``HTML_LINTER_RULES`` variable. The code comments referred to the ``HTML_LINTER_RULES`` variable as being the path to the CSS lint rules rather than being the path to the HTML lint rules, so I went ahead and fixed that typo.

## Proposed Changes

Replaced the term "CSS" with "HTML" in code comments for the HTML_LINTER_RULES variable.

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
Thanks for the catch!

PR #738, opened 2020-09-17
Use YAML syntax in FILTER_REGEX examples

theodore-s-beers:
## Proposed Changes

I think it probably makes more sense for the examples in this README to follow YAML syntax, assuming that people will be running super-linter as a GitHub Action. There's a separate doc covering the process for running the linter locally, in which case, of course, environment variables would be set with an equals sign. It's not a big deal (and no worries about closing the PR on grounds of triviality), but I thought I would make the suggestion.

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
Thank you for helping to keep the documentation consistent! much appreciated

PR #739, opened 2020-09-17
Cleanup Logic for loops

admiralAwkbar:
So this is a crazy idea on how we get into the basic loops...
But the main idea is that we create 2 sets of data:
- All files changed, or modified files changed
We then use that data to see if we need to even go into the specific linter loops.
This would help stop going into loops where the user has no code. Hopefully eliminating many random failures for languages users are not using.

Its a fun idea, and might take some more cleanup

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -258,7 +258,7 @@ DEFAULT_IFS="${IFS}"                                # Get the Default IFS for up
 DEFAULT_DISABLE_ERRORS='false'                          # Default to enabling errors
 export DEFAULT_DISABLE_ERRORS                           # Workaround SC2034
 ERROR_ON_MISSING_EXEC_BIT='false'                       # Default to report a warning if a shell script doesn't have the executable bit set to 1
-export ERROR_ON_MISSING_EXEC_BIT
+export ERROR_ON_MISSING_EXEC_BIT                        # Workaround SC2034
 RAW_FILE_ARRAY=()                                       # Array of all files that were changed
 export RAW_FILE_ARRAY                                   # Workaround SC2034
 READ_ONLY_CHANGE_FLAG=0                                 # Flag set to 1 if files changed are not txt or md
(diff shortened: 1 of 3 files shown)


nvuillam:
I hate to be a buzz killer .... but aren't we reaching the limits of BASH scripts, there are so many copy/paste & redundancies, and every generic behaviour update forces to update all linter-specific code ...
Maybe you have an intern at GitHub who could rewrite all that in typescript or js ? 😈

admiralAwkbar:
@nvuillam @GaboFDC
So this was an interesting idea:
- Now we create the list of files for the respective arrays...
then we use that to make sure we don't go into a linters loop unless there are files for it.
Should help prevent issues about random linters failing... but would love your opinion

admiralAwkbar:
@nvuillam I think I agree...
This was a super fun simple script in the beginning and only handled a few languages and was super simple.
As we have added more and more features and additional languages, that all require their small changes, this is getting to a critical point.
Unfortunately, I built this is my spare time and GitHub does not pay me to work on it, or maintain it, so we open-sourced it.

It may be an interesting idea to get some help for `v4` and bring it into a more high-level language...

I had always hoped we would get some component action steps and be able to split this up to be more reasonable, but that point has not been reached yet :)

nvuillam:
@admiralAwkbar > I never learned to code in Python... maybe it could be an occasion to spend a week-end doing that, I'll think about it ... 😄 

For my "high level" point of view, we could keep all repo content except the /lib folder who would need to be rewritten, do you confirm ?

nvuillam:
@admiralAwkbar I was too tempted to resist .... :D 

cf: https://github.com/github/super-linter/pull/745/files

lib/linter.sh
```diff
@@ -1505,7 +1510,7 @@ fi
 ########################
 # EDITORCONFIG LINTING #
 ########################
-if [ "${VALIDATE_EDITORCONFIG}" == "true" ]; then
+if [ "${VALIDATE_EDITORCONFIG}" == "true" ] && [ ${#FILE_ARRAY_ENV[@]} -ne 0 ]; then
```
GaboFDC, line 1513: Not sure if this is the array we should use for this case

GaboFDC (approved):
This approach looks good to me

lib/linter.sh
admiralAwkbar: Yeah I think editorconfig only uses env

admiralAwkbar:
I dont think i like this approach anymore and need to look at this more

PR #745, opened 2020-09-19
POC: Refactoring in python (draft)

nvuillam:
I made a little POC in python (which is also my hello world in python), it has never been runned once and is probably full of syntax errors , but it can give you a global idea of an architecture we could use :)

- SuperLinter.py class: Runner for all linters
- LinterTemplate.py class: root for any linter to implement
   - GroovyLinter.py: embed call to npm-groovy-lint
   - JavascriptEsLinter.py : embed call to eslint
etc ...

The global idea is that once SuperLinter & LinterTemplates.py architecture is completed , a new linter integration could be easily added by creating a MyLanguageLinter.py in /linters folder :)

The goal is also to have ascending compatibility with Super-Linter V3 capabilities and configuration, so that upgrading to v4 would be transparent for users

Can I have some opinions to see if it looks interesting for a Super-Linter V4 ? :) ( @admiralAwkbar ? )

### Author/Contributor

### Reviewing Maintainer


nvuillam:
Little refactoring perfomed, now it is almost running  :) 
- Load config variables
- List linters 
- List all files
- Execute linters on all matching files

To be continued .... :)

GaboFDC:
Only gave it a broad look, but seems really promising. One thing that comes to mind is that maybe when having multiple linters for a language, instead of having one sub-class from `LinterTemplate` for each, we could have a `LanguajeLinter` sub-class i.e (JavascriptLinter) and then have a sub-class form there for each linter `JavascriptEsLinter`

I think designing it like this, could bring some improvements in how this is managed and allow it to perform better?

Just a thought anyway.

Nice work, would try to follow along to help as much as I can.

nvuillam:
@GaboFDC thanks for your feedback :) 

We could use intermediate sub-classes for Linters for each language , good idea :)

for example : 

**JavascriptLinterRoot** (with only language and file_extensions locally defined)
- **JavascriptEs** (linter_name, config_file_name and build_lint_command locally defined)
- **JavaScriptStandardLinter** (linter_name, config_file_name and build_lint_command locally defined)

But for the moment i'm quite stucked with dynamic imports ... i succeed to run the script with PyCharm right click -> Run , but when called from entrypoint.sh, it doesn't find modules ....

I'd love to get rid of lib.superlinter and replace them by superlinter in code, but for goes knows why , it absolutely wants to consider lib as a package ... any idea ?

nvuillam:
After some struggle I'm satisfied of what i have implemented yet :) 

Working now : 
- Core architecture 
  - root classes when a language has several linters
  - disable linter
  - disable language (new in V4)
  - include and exclude regex filters for languages and linters (new in V4) 
  - groovy, python and javascript linter classes
  - test classes for main class and linter classes (not all linter test classes ok yet)
  - run locally and from docker image

Remaining to be ISO v3
  - add all linters and their special cases
  - `VALIDATE_ALL_CODEBASE = false` management ( git diff etc... )
  - github status management
  - TAP stuff

Ideas of new stuff
   - python superlinter package code coverage

@GaboFDC @admiralAwkbar 
- Any items to add to these lists
- Would it be possible to create a NextVersion branch where I could target my PR, so others could also participate to V4 ?

I start to love Python, that's a very fun HelloWorld to do 🥳

nvuillam:
Fixes #760

.gitignore
```diff
@@ -66,3 +66,17 @@ typings/
 
 # default output report
 super-linter.report
+
+superlinter/__pycache__/
```
GaboFDC, line 70: I think leaving only:
`__pycache__/`
would suffice

superlinter/linters/DockerfileDockerfileLintLinter.py
```diff
@@ -0,0 +1,24 @@
...
+Use dockerfilelint to lint Dockerfile
+https://github.com/replicatedhq/dockerfilelint
+@author: Nicolas Vuillamy
+"""
+import os
+
+from superlinter.linters.DockerfileLinterRoot import DockerfileLinterRoot
+
+
+class DockerfileDockerfileLintLinter(DockerfileLinterRoot):
```
GaboFDC, line 12:
```suggestion
class DockerfileDockerfilelintLinter(DockerfileLinterRoot):
```

GaboFDC (commented):
I have been busy on my work, but gave another broad look, and seems still going in a good direction

superlinter/linters/DockerfileDockerfileLintLinter.py
nvuillam: done :)

.gitignore
nvuillam: done, thanks :)

JustinGrote:
Recommend a look at my super-duper-linter for a reimplementation in Powershell, modularizing things, using configs to define linters, running linters in parallel, etc. :)
https://github.com/JustinGrote/Super-Duper-Linter

nvuillam:
@JustinGrote  this is another product, with no compatibility with Super-Linter, so ... kind of out of the topic here ^^

nvuillam:
Reopen a cleaner PR

PR #754, opened 2020-09-21
Fix ERROR_ON_MISSING_EXEC_BIT initialization

ferrarimarco:
Fixes N/A

## Proposed Changes

1. Fix the initialization of the `ERROR_ON_MISSING_EXEC_BIT` variable. We were not picking an eventually set value from the environment.
2. Add a debug statement to print the values of `ERROR_CODE`, `FILE_TYPE`, and `ERROR_ON_MISSING_EXEC_BIT`

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
good catch!

PR #756, opened 2020-09-21
Update linter.sh

### LINKED ISSUE #755: Protocol Buffer Linter not using GetLinterRules
It seems like the same thing that has happened in #524 with PHP is happening to the Protocol Buffers linter. After adding our `.protolint.yml` config to `.github/linters` as described [in the README](https://github.com/github/super-linter#template-rules-files), the workflow logs still show that its referencing the default location. I'm not sure, but I think it should be added [here](https://github.com/github/super-linter/blob/master/lib/linter.sh#L1272).

#### Logs
```
2020-09-18 22:33:44 [ERROR ]   Linter CMD:[protolint lint --config_path /action/lib/.automation/.protolintrc.yml lms/assignments.proto]
```
#### `.github/linters/.protolintrc.yml`
```
# Lint directives.
lint:
  # Linter rules options.
  rules_option:
    max_line_length:
      max_chars: 90
```
#### `.github/workflows/main.yml`
```
jobs:
  Lint:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout Source
        uses: actions/checkout@v2

      - name: Lint Protos
        uses: github/super-linter@v3
        env:
          VALIDATE_ALL_CODEBASE: false
          VALIDATE_PROTOBUF: true
```

Am I doing it wrong?

admiralAwkbar:
This closes #755 
- We never looked for the file to set its default

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -1270,6 +1270,8 @@ GetLinterRules "PHP_PHPSTAN"
 GetLinterRules "PHP_PSALM"
 # Get PowerShell rules
 GetLinterRules "POWERSHELL"
+# Get Protobuff linter rules
+GetLinterRules "PROTOBUFF"
 # Get Python pylint rules
 GetLinterRules "PYTHON_PYLINT"
 # Get Python flake8 rules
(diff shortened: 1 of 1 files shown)


admiralAwkbar (commented):
just needed to go look for it :)

PR #766, opened 2020-09-24
Lint all changed files with editorconfig-checker

mshimokura:
## Proposed Changes

1. Lint _all_ changed files with editorconfig-checker, not just files that were detected to have the "ENV" type. Using the FILE_ARRAY_ENV seemed like a mistake honestly, I'm not sure why those files specifically should be checked against the .editorconfig. For example, previously editorconfig-checker would not be run against Javascript files if the VALIDATE_ALL_CODEBASE variable was set to false, but it would it if was set to true.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -1542,7 +1542,7 @@ if [ "${VALIDATE_EDITORCONFIG}" == "true" ]; then
   # Lint the files with editorconfig #
   ####################################
   # LintCodebase "FILE_TYPE" "LINTER_NAME" "LINTER_CMD" "FILE_TYPES_REGEX" "FILTER_REGEX_INCLUDE" "FILTER_REGEX_EXCLUDE" "FILE_ARRAY"
-  LintCodebase "EDITORCONFIG" "editorconfig-checker" "editorconfig-checker" "^.*$" "${FILTER_REGEX_INCLUDE}" "${FILTER_REGEX_EXCLUDE}" "${FILE_ARRAY_ENV[@]}"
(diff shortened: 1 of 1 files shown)


ferrarimarco (approved):

GaboFDC:
Is counter intuitive, but this is the file array it's been used since it was added on #221 
I think the related bug is coming from another change. Editor config is kind of an special case.
Did you tested this changes?
maybe @mstruebing can provide better insigths on this?

mshimokura:
Regrettably, I have not yet figured out how to test this PR. I can create a test-case repository if you like, and if someone can point me to instructions on how to run this change I'd be happy to do that as well.

ferrarimarco:
The `editorconfig-checker` test cases run correctly, but this is not enough as you pointed out. To test this against the whole code base, you should have a .editorconfig file. A test repo you can use is `https://github.com/ferrarimarco/dotfiles`, which contains one.

mshimokura:
okay, so I figured out how to build the docker container and configure my fork to use that built container as an action.
Here are the test cases:

https://github.com/mshimokura/super-linter-editorconfig-test/pull/1
This test case fails because the linter does not catch any errors, even though new.b has incorrect indentation.

https://github.com/mshimokura/super-linter-editorconfig-test/pull/3
This test case passes because the linter catches the indentation problems.

mstruebing:
Looks like not an issue with editorconfig-checker itself if I get it correctly.
unfortunately I'm not so much into the super-linter codebase and I think it was changed since editorconfig-checker was added.

However, you are able to lint your whole codebase with `editorconfig-checker` and specific files with `editorconfig-checker [FILE]`

mshimokura:
yeah, that's right. the issue we're seeing is that when we're trying to lint only changed files in our pull requests, those files don't get passed into editorconfig-checker (because its only passing in ".env" type files)
when we're linting the whole codebase though its fine

admiralAwkbar (approved):

PR #757, opened 2020-09-21
Lint with Kubeval and fixes

ferrarimarco:
Fixes #730
Fixes #272

## Proposed Changes

1. Lint Kubernetes descriptors with [kubeval](https://github.com/instrumenta/kubeval).
1. Fix the initialization of `DEFAULT_ANSIBLE_DIRECTORY` and `REPORT_OUTPUT_FOLDER`. Both variables are now initialized after `GITHUB_WORKSPACE` is initialized.
1. Fix the command that populates file listing in `LintAnsibleFiles`.
1. Disable read-only skipping of Ansible files introduced in #34. I suppose it was a leftover, and `ansible-lint` supports linting of read-only files. There's no need to disable linting in this case, at least AFAIK.
1. Add a couple of debug statements when validating `ANSIBLE_DIRECTORY`.
1. Fix a typo in a comment in `.github/workflows/deploy-DEV.yml`.

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [ ] Label as `breaking` if this is a large fundamental change
- [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #769, opened 2020-09-25
Remove READ_ONLY_CHANGE_FLAG flag and fix .editorconfig and shfmt tests

ferrarimarco:
## Proposed Changes

1. Remove all assignments of `READ_ONLY_CHANGE_FLAG`. It's not needed anymore after #757 is merged
1. Add the `root = true` to `.automation/test/editorconfig-checker/.editorconfig` directive so that it doesn't conflict with an eventual, future, .editorconfig file.
1. Fix an error in `.automation/test/shell_shfmt/.editorconfig`. It was looking for the wrong file extension (`.ext`) instead of the shell script default one (`.sh`).
1. Fix shfmt test files names. An extra `_` was fooling the test script.

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
Thank you for the fix! this is amazing

PR #770, opened 2020-09-25
Ignore .git and symlink .dockerignore

ferrarimarco:
## Proposed Changes

1. Create a symbolic link to `.gitignore` named `.dockerignore`, so that we don't have to repeat ourselves.
1. Ignore the `.git` directory, so that it's ignored by the Docker daemon when building the container, because of the symbolic link above. This saves about 15-20 seconds of build time (the .git directory is currently about 20 MB in size).

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ typings/
 
 # default output report
 super-linter.report
+
+# Git directory (useful for .dockerignore)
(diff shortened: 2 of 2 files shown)


ferrarimarco:
The build is failing due to a connection issue.

admiralAwkbar (approved):

admiralAwkbar:
@ferrarimarco makes sense! very nice thoughts on this one!

alauq (approved):

PR #771, opened 2020-09-25
Sort environment variables when printing debug info

ferrarimarco:
## Proposed Changes

1. Sort the environment variables when printing them with `printenv`. This makes debugging easier, because you don't have to manually scan the whole list looking for the variable you need to check.

### Author/Contributor

### Reviewing Maintainer


alauq (commented):
👾

admiralAwkbar (approved):
clean fix!

Issue #760, opened 2020-09-22
Simplify the code base


ferrarimarco:
**Is your feature request related to a problem? Please describe.**
Extending super-linter is hard, and there's a lot of duplicated code. This causes errors and issues, and hinders the maintainability of the project.

**Describe the solution you'd like**
Here are some things that come to mind, after an inspection of the code base:

1. Use a single array for languages and linters: having two makes it harder adding new linters.
1. Define VALIDATE_* variables using the single array above, instead of having to define one each time a new linter is added. Maybe with some `eval`s?
1. Increase the separation between linters by grouping them according to the language (shell, Java, JavaScript, Dockerfile...). super-linter is already headed this way, but the design doesn't currently enforce it.
1. Reuse where possible:
    1. Have a generalized mechanism, which should work in the majority of cases. Should support at least
        1. Lint all the files in the codebase (example: editorconfig-checker).
        1. Lint all the files in a specific directory (example: kubeval, ansible-lint).
        1. Lint all the files in a specific directory, including the ones that match a regex.
        1. Lint all the files in a specific directory, excluding the ones that match a regex.
    1. Make it possible to specialize the behavior for each language (e.g. Java, JavaScript...).
    1. Make it possible to specialize the behavior for each linter.
        1. For example, when a linter supports a rules file.
    1. Leverage `eval` a bit more. Ideally, the only things we should do to add a new linter, should be:
        1. Add the linter to the list of available linters.
        1. Install the linter in the environment (modifying the `Dockerfile`).
        1. Customize the behavior, only if needed.
1. Have a single point where we build file lists. There are multiple ways of building the file list to lint, considering the operational modes (validate the whole codebase, or only what changed). Sometimes, the logic to build such lists is not the same (for example, see #757 where I fixed a couple of these issues when building the file list for Ansible).
1. Remove unused code and environment variables.
1. Have a centralized "exclusion list", where we define things to globally ignore (node_modules, .git...).

NOTE: the above list will likely be updated as the discussion in this thread goes on.

I suspect that the above points will make super-linter way more manageable and extensible. Thoughts?

ferrarimarco:
Additionally, we might leverage a descriptor that we parse to build the environment. For example, a `JSON` that we parse with `jq` that describes the properties of each linter (name, how to get the version, default command switches, default rules file name...).

This scenario also supports the use case that sees a user personalizing this configuration. For example, super-linter now doesn't support any mechanism to personalize the options of linters that don't expose them via a configuration file.

[...]

PR #781, opened 2020-09-28
Remove unneeded variables and eval the rest

ferrarimarco:
Related to #760 

## Proposed Changes

1. Remove the assignment of `VALIDATE_XXXXX` variables. Those variables are not needed because they are set in `validation.sh` anyway. By removing those variables, we don't have to remember to add new ones when extending super-linter to add new languages, and we have to update the the `LANGUAGE_ARRAY` array only.
1. Define and export `ERRORS_XXXX` variables dynamically in `validation.sh`, based on the values in `LANGUAGE_ARRAY`, so that we don't have to define a new `ERRORS_XXXX` variable for each linter we add.
1. Move the initialization of debug and log variables at the top of `linter.sh`, so that we can use logging in sourced files and during the initialization of `linter.sh`.
1. Remove the initialization of such variables from the `Dockerfile` as well, because they are not needed during the build and might confuse the users.

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar:
@ferrarimarco This is amazing... i didnt know you could do that lol

admiralAwkbar:
@ferrarimarco It also seems to skip all runs... check out https://github.com/github/super-linter/pull/781/checks?check_run_id=1178767786

Seems to see no languages and skips all

ferrarimarco:
Will have a look ASAP!

ferrarimarco:
I think tests are running fine, judging from the execution time of that step. There must be something with log settings. I'm checking now.

ferrarimarco:
Should be fixed now. Let's see.

ferrarimarco:
@admiralAwkbar shall we take the chance to remove variables from the `Dockerfile` as well? Why do we need them there if we set defaults or load values from the runtime environment?

admiralAwkbar:
@ferrarimarco I'm trying to understand. I thought it was necessary to add them to the docker container. and it may be needed for users who are running this locally.
That's my fear of removing them from the docker container
but the rest seems to make sense.

admiralAwkbar:
@ferrarimarco It seems to be working as expected and this did seem to clean up a lot of the issues we normally see! pretty cool

ferrarimarco:
I suppose it might do the opposite, meaning: it might confuse users. The [ENV](https://docs.docker.com/engine/reference/builder/#env) directive does (quoting from the docs):

```
The ENV instruction sets the environment variable <key> to the value <value>. This value will be in the environment for all subsequent instructions in the build stage and can be replaced inline in many as well.

[...]

The environment variables set using ENV will persist when a container is run from the resulting image. You can view the values using docker inspect, and change them using docker run --env <key>=<value>.

Environment variable persistence can cause unexpected side effects. For example, setting ENV DEBIAN_FRONTEND=noninteractive changes the behavior of apt-get, and may confuse users of your image.

If an environment variable is only needed during build, and not in the final image, consider setting a value for a single command instead

[...]
```

So I guess we don't need those variables in the Dockerfile. Why would we want to set up during build time like that? We don't care about those values when building the image.

admiralAwkbar:
@ferrarimarco I think your right! i think some additional tests and we can knock this out

ferrarimarco:
I've pushed a new commit to remove those vars as well. Let's see what the testing suite has to say :)

admiralAwkbar (approved):
Send it

PR #802, opened 2020-10-02
Dynamically initialize FILE_ARRAY variables, add a missing export

ferrarimarco:
## Proposed Changes

1. Dynamically initialize `FILE_ARRAY` variables by iterating on the `LANGUAGES` array.
1. Remove the statically defined `FILE_ARRAY` variables.
1. Add a missing export of `LANGUAGE` variables.
1. Make each linter use its own file array.
1. Populate multiple file arrays where needed.

Note that I couldn't initialize the `FILE_ARRAY` variables in `validation.sh` because you cannot export arrays in Bash (yet).

This is a follow-up to #781 . Hopefully more PRs in this spirit are on the way!

### Author/Contributor
- [ ] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [ ] Label as `breaking` if this is a large fundamental change
- [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #803, opened 2020-10-03
Centralize file list building logic in buildFileList.sh

ferrarimarco:
## Proposed Changes

1. Centralize the file list building in `buildFileList.sh`. This simplifies the logic of super-linter, because we use the same functions to generate the list of files to lint, regardless of the value of `VALIDATE_ALL_CODEBASE`. The value of that variable is now used in `buildFileList.sh` to limit file listing to what's changed in the last commit, as it was before. While streamlining this logic, I found several inconsistencies in how we were building file lists, that I fixed. See below.
1. Remove the `KUBERNETES_DIRECTORY` variable and logic, because it's not needed now that we centeralized the file list building logic.
1. Changed the file listing detail log from `info` to `debug`. This was too verbose for a regular run, and useful to debug super-linter, and not too much for the user.
1. Fix the Dockerfile files building logic to exclude `dockerfilelintrc`.
1. Fix the Dockerfile files building logic to correctly populate `FILE_ARRAY_DOCKERFILE`. It was wrongly populating `FILE_ARRAY_DOCKER`.
1. Add missing extensions when building the list of PowerShell files. We were missing a couple of extensions that we were checking only when validating the whole code base.
1. Fix the Python file list building logic. We were not populating `FILE_ARRAY_PYTHON_BLACK` at all.
1. Fix `Snakefile` detection. We were looking for `Snakefile`, but it should be `snakefile`, because we lowercase `BASE_FILE`.
1. Remove the `PopulateShellScriptsList`, because we now rely on the centralize file list building logic.
1. Remove the `FILE_EXTENSIONS` parameter from the `LintCodebase` function, because we don't need to specify that parameter anymore, and it was ignored in some cases anyway.
1. Make the `grep` in `DetectAWSStatesFIle` quiet, otherwise it shows confusing matches in the output.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
Maybe it's easier to have a look at #802 first. I can fix conflicts later.

ferrarimarco:
@admiralAwkbar rebased against the latest commit in the main branch 👍

ferrarimarco:
The build is failing because of a connection issue.

admiralAwkbar (approved):

PR #808, opened 2020-10-06
Remove unnecessary --update for apk in Dockerfile

PeterDaveHello:
There are two `apk add` use both `--update` and `--no-cache` parameter,
which is unnecessary, usually should use only `--no-cache` to prevent
leaving cache files, just like the other `apk add` in line 249

## Proposed Changes

Remove unnecessary --update for apk in Dockerfile

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
Thank you!

PR #814, opened 2020-10-07
Fix documentation error with VALIDATE_MD being renamed

omusavi:
The Environment Variable VALIDATE_MD is no longer being used, but the documentation was never updated to show that.

Simple documentation only change to update environment variable

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
nice catch

PR #815, opened 2020-10-07
Dynamically load linter rules

ferrarimarco:
## Proposed Changes

1. Dynamically load linter rules, so that we don't have to remember adding the call to `GetLinterRules` when adding a new language.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -311,6 +311,14 @@ GetLinterRules() {
   LANGUAGE_LINTER_RULES="${LANGUAGE_NAME}_LINTER_RULES"
   debug "Variable names for language file name: ${LANGUAGE_FILE_NAME}, language linter rules: ${LANGUAGE_LINTER_RULES}"
 
+  #####################################################
+  # Check if the language rules variables are defined #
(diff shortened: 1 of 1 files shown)


lib/linter.sh
```diff
@@ -311,6 +311,14 @@ GetLinterRules() {
   LANGUAGE_LINTER_RULES="${LANGUAGE_NAME}_LINTER_RULES"
   debug "Variable names for language file name: ${LANGUAGE_FILE_NAME}, language linter rules: ${LANGUAGE_LINTER_RULES}"
 
+  #####################################################
+  # Check if the language rules variables are defined #
+  #####################################################
+  if [ -z "${!LANGUAGE_FILE_NAME+x}" ]; then
```
admiralAwkbar, line 317: This is a super interesting way to write that :)

admiralAwkbar (approved):

lib/linter.sh
ferrarimarco: Thanks!

PR #824, opened 2020-10-08
editorconfig-checker - Use .ecrc Files

mshimokura:
## Proposed Changes

1. Use the -config option for editorconfig-checker to specify an .ecrc file. This allows you to change the linter options, i.e. allowing spaces after tabs.
2. Add a variable EDITORCONFIG_FILE_NAME to specify the .ecrc file to use.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
Hi! Nice work.

What if you change the variable name from `EDITORCONFIG_FILE_NAME` to `EDITORCONFIG_CHECKER_FILE_NAME`?

`EDITORCONFIG_FILE_NAME` might lead to confusion with the `.editorconfig` file itself.

Thoughts?

mshimokura:
actually, so I tried that out, and it doesn't work because there's a naming convention for the variables that it should be `${LANGUAGE_NAME}_FILE_NAME`. 
https://github.com/github/super-linter/blob/a5cada70f874be6fd70f8f780672d3cc2c645ba8/lib/linter.sh#L313

So, changing the variable name ends up making it not pick up the config file.

lib/linter.sh
```diff
@@ -77,7 +77,7 @@ DOCKERFILE_LINTER_RULES="${DEFAULT_RULES_LOCATION}/${DOCKERFILE_FILE_NAME}" # Pa
 DOCKERFILE_HADOLINT_FILE_NAME="${DOCKERFILE_HADOLINT_FILE_NAME:-.hadolint.yml}"               # Name of the file
 DOCKERFILE_HADOLINT_LINTER_RULES="${DEFAULT_RULES_LOCATION}/${DOCKERFILE_HADOLINT_FILE_NAME}" # Path to the Docker lint rules
 # Editorconfig Vars
-EDITORCONFIG_FILE_NAME='.ecrc'                                                  # Name of the file
```
joejump: >

joejump (changes requested):
groovylintrc.json'

admiralAwkbar (approved):

PR #827, opened 2020-10-09
ignore kustomize file with kubeval

PascalBourdier:
currently, kubeval doesn't support kustomize file

so kubeval returns a false-positive about a missing `metadata` key:
```
2020-10-09 06:19:19 [ERROR ]   Found errors in [kubeval] linter!
2020-10-09 06:19:19 [ERROR ]   [ERR  - .../kustomization.yml: Missing 'metadata' key]
2020-10-09 06:19:19 [ERROR ]   Linter CMD:[kubeval --strict .../kustomization.yml]
```

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -559,12 +559,7 @@ DetectKubernetesFile() {
   FILE="${1}" # File that we need to validate
   debug "Checking if ${FILE} is a Kubernetes descriptor..."
 
-  if grep -q -E 'apiVersion:\s*kustomize.config.k8s.io' "${FILE}" >/dev/null; then
-    debug "${FILE} is NOT a Kubernetes descriptor (Kustomize only)"
-    return 1
-  fi
-
-  if grep -q -E '(apiVersion):' "${FILE}" >/dev/null; then
+  if grep -v 'kustomize.config.k8s.io' "${FILE}" | grep -q -E '(apiVersion):'; then
     debug "${FILE} is a Kubernetes descriptor"
     return 0
   fi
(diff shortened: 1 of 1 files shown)


joejump (commented):
teach me

lib/linter.sh
```diff
@@ -559,6 +559,11 @@ DetectKubernetesFile() {
   FILE="${1}" # File that we need to validate
   debug "Checking if ${FILE} is a Kubernetes descriptor..."
 
+  if grep -q -E 'apiVersion:\s*kustomize.config.k8s.io' "${FILE}" >/dev/null; then
```
ferrarimarco, line 562: Can you simplify the check by embedding the negation of this check in the existing `if`? So we don't have to repeat ourselves in the debug message.

@admiralAwkbar

PascalBourdier: fixed according your review

ferrarimarco: Nice!

lib/linter.sh
```diff
@@ -559,7 +559,7 @@ DetectKubernetesFile() {
   FILE="${1}" # File that we need to validate
   debug "Checking if ${FILE} is a Kubernetes descriptor..."
 
-  if grep -q -E '(apiVersion):' "${FILE}" >/dev/null; then
+  if grep -v 'kustomize.config.k8s.io' "${FILE}" | grep -q -E '(apiVersion):'; then
```
ferrarimarco, line 562: Do you need to enable the `-q` flag for the condition you added as well?

PascalBourdier: my condition require an output for the `|`
so `-q` flag must (and can) only be on the the second `grep` (or a redirection to `/dev/null` if you prefer)

admiralAwkbar (approved):
looks good

PR #853, opened 2020-10-15
Update linter.sh

### LINKED ISSUE #852: JavaScript ESLint Issue (And possible solution)
**Describe the bug**
Trying to now use a configuration file with ESLint causes issues. This is probably because the script tries to look for JAVASCRIPT_ES_LINTER_RULES (it's dynamically doing this per language) but in some places some things are still defined as JAVASCRIPT_LINTER_RULES. This causes an issue where it skips over custom rules and uses the `DEFAULT_RULES_LOCATION='/action/lib/.automation'` and says that no eslint custom configuration file is present.

**To Reproduce**
Set everything up like normal. Use JAVASCRIPT_ES_CONFIG_FILE. It will keep using the default one or error out because it cannot find the provided file name in the DEFAULT_RULES_LOCATION

**Expected behavior**
I expect it to load the configuration file correctly, but because of an old naming scheme, some of the javascript environment variables do not get updated.

**Screenshots (and possible solution)**
I feel like these vars should be JAVASCRIPT_ES_LINTER_RULES instead of this, because the default one does not get used anymore, so when a config file is provided, only JAVASCRIPT_ES_[var] will get dynamically updated and not JAVASCRIPT_[var].

JAVASCRIPT_ES is in the languages list, JAVASCRIPT is not.

Screenshot from 2020-10-15 01-41-05

**Additional context**
Screenshot from 2020-10-15 01-46-25

As you can see here it looks for JAVASCRIPT_ES_LINTER_RULES **not** JAVASCRIPT_LINTER_RULES
Screenshot from 2020-10-15 02-50-14

PhunStyle:
# Proposed Changes

Change environment variable JAVASCRIPT to JAVASCRIPT_ES where applicable. Might fix #852

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -92,8 +92,8 @@ HTML_LINTER_RULES="${DEFAULT_RULES_LOCATION}/${HTML_FILE_NAME}" # Path to the HT
 JAVA_FILE_NAME="sun_checks.xml"                                 # Name of the Java config file
 JAVA_LINTER_RULES="${DEFAULT_RULES_LOCATION}/${JAVA_FILE_NAME}" # Path to the Java lint rules
 # Javascript Vars
-JAVASCRIPT_FILE_NAME="${JAVASCRIPT_ES_CONFIG_FILE:-.eslintrc.yml}"          # Name of the file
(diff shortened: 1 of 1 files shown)


ferrarimarco:
Thanks for this attempt! I'm taking a more comprehensive approach to fix this, but you were on the right track! Let's leave this open for now.

ferrarimarco:
Closing this, in favor of #854, for a more generalized fix.

PR #854, opened 2020-10-15
Simplify the rules loading logic, plus related fixes

### LINKED ISSUE #852: JavaScript ESLint Issue (And possible solution)

ferrarimarco:
Fixes #852 

## Proposed Changes

1. Dynamically initialize `<LANGUAGE>_LINTER_RULES` variable in `GetLinterRules`, and removed the related static assignment from `linter.sh`. For this to work, I had to add a parameter to `GetLinterRules`, pointing to `DEFAULT_RULES_LOCATION`
1. Fix Javascript and Typescript linter rules variable initialization. Now each linter has it's own variable, so the dynamic initialization introduced in older PRs works.
1. Initialized variables in `GetLinterRules` to point to the paths of the primary and secondary language rules files, to avoid repetitions.
1. Added a check for the existence of the rules file in the cases where we expect it to be there (if the user specified one, or the default one).
1. Enable debug output during CI builds.

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar:
@ferrarimarco Very nice work!

ferrarimarco:
Thanks! There are still a couple of rough edges (tests failing), but I'll fix them :)

ferrarimarco:
@admiralAwkbar should be ok now!

admiralAwkbar (approved):
amazing work

PR #746, opened 2020-09-19
Move images to GHCR

assignUser:
Closes #735

### Author/Contributor
- [ ] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [ ] Label as `breaking` if this is a large fundamental change
- [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #857, opened 2020-10-15
Extract ENV vars to config file

IAmHughes:
## Proposed Changes

This PR introduces the concept of a `.super-linter-config.yml` that would reside in the `.github` directory on a repository, and contain all the values we currently require to be exported as [environment variables](https://github.com/github/super-linter#environment-variables).

These changes will be released in `v3`, and will display a deprecation warning to users so they know to create a `.super-linter-config.yml` file before switching to `v4` once released.
- *:warning:* **Note:** *:warning:* This will be mandatory in the next major release (`v4+`)
- *:warning:* **Note:** *:warning:* The configuration file (`.github-super-linter.yml`) will override any **Environment** variables that are passed

This will simplify the process of starting with the `super-linter` and allow for ease of editing a config, versus modifying environment variables, to control linter rules, enable/disable features, etc.

### Author/Contributor

### Reviewing Maintainer
- [X] Label as `breaking` if this is a large fundamental change
- [X] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`


IAmHughes:
@ferrarimarco I wanted to tag you to get your thoughts on this as you just changed various rules for the linters.

ferrarimarco:
Sure, I can have a look as soon as it's ready for review!

In the meantime, I'm baking another PR to get rid of TestCodeBase, merging it with LintCodeBase. 80% of duplicated code, and now that we've the dynamic linter invocation, we should use it for testing as well :)

IAmHughes:
cc @admiralAwkbar for [above](https://github.com/github/super-linter/pull/857#issuecomment-710152599) ☝️

.github/workflows/deploy-DEV.yml
```diff
@@ -82,4 +82,4 @@ jobs:
       ############################################
       - name: Run against all code base
         shell: bash
-        run: docker run -e RUN_LOCAL=true -e OUTPUT_DETAILS=detailed -v ${GITHUB_WORKSPACE}:/tmp/lint github/super-linter:${GITHUB_SHA}
+        run: docker run -e RUN_LOCAL=true -e OUTPUT_DETAILS=detailed -v ACTIONS_RUNNER_DEBUG=true -v ${GITHUB_WORKSPACE}:/tmp/lint github/super-linter:${GITHUB_SHA}
```
ferrarimarco, line 85: This should be `-e ACTIONS_RUNNER_DEBUG=true`, not -v. I think I introduced this error in a old PR :(

IAmHughes:
```suggestion
        run: docker run -e RUN_LOCAL=true -e OUTPUT_DETAILS=detailed -e ACTIONS_RUNNER_DEBUG=true ${GITHUB_WORKSPACE}:/tmp/lint github/super-linter:${GITHUB_SHA}
```

Like this @ferrarimarco

IAmHughes: or do I keep the `-v` after the `ACTIONS_RUNNER_DEBUG`

ferrarimarco: keep it :) just change -v ACTIONS_RUNNER_DEBUG to -e ACTIONS_RUNNER_DEBUG

IAmHughes:
```suggestion
        run: docker run -e RUN_LOCAL=true -e OUTPUT_DETAILS=detailed -e ACTIONS_RUNNER_DEBUG=true -v ${GITHUB_WORKSPACE}:/tmp/lint github/super-linter:${GITHUB_SHA}
```

joojoodamn:
My problem is that I can't even see #857 and that just hurts my feelings

ferrarimarco:
I like this direction (and I even suggested something similar in #760 !), so great work!

I'm wondering if we should include the full linter configuration, and not only the `VALIDATE_` boolean, such as the `LINTER_NAME` and the `LINTER_COMMAND`. The user will then have full customization support. Note that if we do this, we've to take into account "merging" the user configuration with our defaults.

This seems the right chance to do it, so that we can remove hard-coded values for languages, linter names, and linter commands.

IAmHughes:
> I'm wondering if we should include the full linter configuration, and not only the VALIDATE_ boolean, such as the LINTER_NAME and the LINTER_COMMAND.

@ferrarimarco great idea! In pairing with @admiralAwkbar, I think that could probably be the next iterative step once this is merged. This is a lot of change already, so I'd like to take an iterative approach, but the idea of removing hard-coded language values should definitely be implemented in the near future.

admiralAwkbar:
@ferrarimarco think you could lend a hand on this one? it's proving to be a monster lol

ferrarimarco:
I see! What's the current status of this?

admiralAwkbar:
@ferrarimarco so i think the logic flows as:
- linter.sh is envoked
- We go initally into to the config parser
- We set those pulled from the config file as env vars
- We then go down the current route of looking through the env vars and seeing if not set, then defaulting, etc...

I think I'm mentally hung up on how to set the config vars as env...

ferrarimarco:
What about a for loop that dynamically loads the variables? Much like we're doing now in linter.sh to initialize the errors count variables for example...

admiralAwkbar:
@ferrarimarco Thats the idea with the reading of the file and loading the variables... but I'm missing something still it seems. i cant get the vars to persist i think...

lib/loadConfig.sh
```diff
@@ -0,0 +1,140 @@
...
+    mapfile -t CLEANED_VAR_ARRAY < <(CleanVarArray "${GET_ALL_VARS_CMD[@]}")
+
+    ##################################
+    # Set the ENV with the variables #
+    ##################################
+    for PAIR in "${CLEANED_VAR_ARRAY[@]}"; do
+      KEY=$(echo "${PAIR}" | cut -f1 -d '|')
+      VALUE=$(echo "${PAIR}" | cut -f2 -d '|')
+      info "Setting ENV with [${KEY}]:[${VALUE}]"
+      export "${KEY}=${VALUE}"
```
ferrarimarco, line 94: I think you should use `eval` here, and split the definition and the export of the variable in two steps.

Additionally, I'd try without exporting first, because we call this function from `linter.sh` without spanning any subshell, so I suspect there's no need to export.

ferrarimarco:
@admiralAwkbar added a comment inline that might help! If the suggestion doesn't work, I'll have a deeper look :)

.github/.super-linter-config.yml
```diff
@@ -0,0 +1,329 @@
+---
```
ferrarimarco, line 1: We should leverage the default configuration in TEMPLATES/.super-linter-config.yml. This file is identical to that one. No need for a duplicate.

ferrarimarco, line 1: Why don't we use a `yaml` extension? `yml` is deprecated :) https://stackoverflow.com/questions/22268952/what-is-the-difference-between-yaml-and-yml-extension/22268973 and https://yaml.org/faq.html

lib/loadConfig.sh
admiralAwkbar: thank you!

admiralAwkbar:
@ferrarimarco 
- I updated all the configs from `yml` to `yaml` to go with the go forward path
- I removed all duplicate files from `.github/linters` and cleaned that up nicely...

ferrarimarco:
@admiralAwkbar I still see both `TEMPLATES/.super-linter-config.yaml` and `.github/.super-linter-config.yaml` tho...

admiralAwkbar:
@ferrarimarco So the file at `TEMPLATES` is for users to make a copy of, and place in their `.github/` folder as we move to using the config file.
We will explicitly tell them to copy that file and make any updates they see fit.

The file at `.github/super-linter-config.yaml` is our personal 1 used for running the super linter against itself. I assume in the near future we will need to update it a bit from the default values of the TEMPLATE file.

I hope that helps explain

ferrarimarco:
You know, this raises a question I had floating around: do we need stuff in `TEMPLATES` at all? AFAIK, we don't update it when new configuration options are added to new linter versions (if we don't remember to!), nor we remove deprecated options. What if we just linked to each linter doc that support a config file? So that we don't have to maintain a copy ourselves, that we keep here just for convenience.

If the majority of super-linter users run it as a GitHub action, they don't even see those templates that often :)

admiralAwkbar:
@ferrarimarco I believe the templates are pulled into place at run time if the user does not provide a template or rules file.
So we need to maintain the base copy of the file with basic rules at least i believe

admiralAwkbar:
@ferrarimarco also seems actions is down a bit right now... so this is taking forever... lol

MusaBack (dismissed):

admiralAwkbar:
@ferrarimarco So i think the code for updating all the vars is solid...
But the vars for getting the arrays seems to be trashed... I'm not sure whats really wrong there... any chance you can take a look at the calls:
```bash
# shellcheck disable=SC2125
      LANGUAGE_FILE_ARRAY_NAME="${FILE_ARRAY_VARIABLE_NAME}"[@]
      LANGUAGE_FILE_ARRAY=("${!LANGUAGE_FILE_ARRAY_NAME}")
      debug "${FILE_ARRAY_VARIABLE_NAME} file array contents: ${!LANGUAGE_FILE_ARRAY}"

      debug "Invoking ${LINTER_NAME} linter. TEST_CASE_RUN: ${TEST_CASE_RUN}"
      LintCodebase "${LANGUAGE}" "${LINTER_NAME}" "${LINTER_COMMAND}" "${FILTER_REGEX_INCLUDE}" "${FILTER_REGEX_EXCLUDE}" "${TEST_CASE_RUN}" "${!LANGUAGE_FILE_ARRAY}"
```

admiralAwkbar:
@ferrarimarco around line 1294 on `linter.sh`

ferrarimarco:
Sure! I can give it a shot (my) tomorrow!

IAmHughes:
> Sure! I can give it a shot (my) tomorrow!

@ferrarimarco thanks! It seems like maybe a pass by reference versus pass by value issue, but I'm not sure yet.

ferrarimarco:
I see some errors in the log, so the script is not called. May be the reason why variables are not persisted!

```shell
/action/lib/loadConfig.sh: line 16: info: command not found
/action/lib/loadConfig.sh: line 17: info: command not found
/action/lib/loadConfig.sh: line 24: warn: command not found
/action/lib/loadConfig.sh: line 25: warn: command not found
```

ferrarimarco:
After fixing the issue with logging, I see:

```shell
2020-10-24 09:42:17 [WARN  ]   Config file:[] not found in repository, will use ENV vars only!
2020-10-24 09:42:17 [WARN  ]   This will be deprecated in v4+!
```

ferrarimarco:
Added a safety check for the `SUPER_LINTER_CONFIG_FILE` variable, because I don't see it defined anywhere. In fact, now the script terminates with a fatal error (as expected).

ferrarimarco:
So, we obviously need to define `SUPER_LINTER_CONFIG_FILE`.

The file name part is easy: I defined `SUPER_LINTER_CONFIG_FILE_NAME`, pointing to `.super-linter-config.yaml`.

The path to that file is tricky, because of a chicken and egg problem. We want to look in `DEFAULT_WORKSPACE`, or in `GITHUB_WORKSPACE` if `DEFAULT_WORKSPACE` is not defined, also taking `LINTER_RULES_PATH` into account.

But those variables might be defined in the configuration file! But we need them to look for the configuration file!

Possible solutions:

- Expect the config file to be in a well-known location while still allowing the configuration of the default workspace, but this might complicate mounting files and volumes.
- Avoid the customization of the workspace, and expect it to be always in the same location (e.g. `/tmp/lint`).

In any case, if the user doesn't provide any super-linter configuration file, then we go with the default one we provide in `TEMPLATES`.

GaboFDC:
This may be a bit controversial but I have 1 main concern:
1) this is small and not really needed but maybe moving the change of yml to yaml to another pr is better? so code review and number of changes is more focused?
2) what is the benefit from moving this to a file? From what I have seen, all standard GitHub actions use the envs for config, also with this change the users have 2 files to maintain, the workflow, and the config, I think is easier to keep it in the same file, also the current default config is really simple and small, with this change the default config (template) is hundreds of lines long, I don't really see a benefit from moving this to a new file.

> This will simplify the process of starting with the super-linter
How I see it, is easier to start right now, as you just add the lint workflow, and edit as needed, with this we need to add workflow, then copy the config and edit it as needed.

> and allow for ease of editing a config, versus modifying environment variables, to control linter rules, enable/disable features, etc.
There is no practical difference between editing a line in the lint workflow and the config file, you just need to change/add/remove a line we are just changing the place where is done.

Not sure if I'm biased but this is how I see it, any feedback and clarification about things I may be overlooking is really appreciated.

admiralAwkbar:
@GaboFDC I agree with you on this...
As we started to move to a config file we can see:
- It's not so clean or easy
- It's just as big and bulky
- It causes some fun on getting things started...

I'm open to trashing this for another time when it's evident it needs it

joojoodamn:
#746

PR #905, opened 2020-10-24
bugfix: #900

### LINKED ISSUE #900: Push to branch: No files were found in the GITHUB_WORKSPACE
Hello, 

Below command will find 0 differences between HEAD and last commit id if push commit to DEFAULT_BRANCH
 
https://github.com/github/super-linter/blob/b058f4ac1e4dbd77f3b75dcdece5d879f082ae19/lib/buildFileList.sh#L65

In case of PUSH to branch it is better use 

`git diff-tree --no-commit-id --name-only -r ${GITHUB_SHA}`

or 

`git show --pretty="" --name-only ${GITHUB_SHA}`

Warning message:
image

IvanKuchin:
Fixes #900

## Proposed Changes

Problem:
There is no differentiation between push and PR event type in BuildFileList function. File list gets empty when git commit pushes to DEFAULT_BRANCH.

Solution:
Make difference between push and PR requests to craft file list depends on request type

### Author/Contributor

### Reviewing Maintainer
- [X] Label as `bug`

### DIFF
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -61,12 +61,12 @@ function BuildFileList() {
       # print header #
       ################
       debug "----------------------------------------------"
-      debug "Generating Diff with:[git diff-tree --no-commit-id --name-only -r ${GITHUB_SHA}]"
+      debug "Generating Diff with:[git diff-tree --no-commit-id --name-only -r \"${GITHUB_SHA}]\""
 
       #################################################
       # Get the Array of files changed in the commits #
       #################################################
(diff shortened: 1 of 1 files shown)


IvanKuchin:
I'm not sure how I can fix failure in action "Deploy Docker Image -DEV (pull_request)

image

admiralAwkbar (approved):

dec5e:
This behaviour works correctly only for pushes with a single commit. Please check #1167.

PR #927, opened 2020-10-30
Don't override the secondary rules file path, if available

ferrarimarco:
## Proposed Changes

1. Don't override the rules file path when it has a secondary file name to check. Example: `.hadolint.yml` vs `.hadolint.yaml`. Before this PR, we were missing an `else` statement. I took the chance to simplify the `if` as well, adding a check for the existence of a secondary file name in the "main" `if`, instead of nesting it in the "main" `else` block.
1. Change the default file name for hadolint config file from `.hadolint.yml` to `.hadolint.yaml`.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
The build is failing, but for an unrelated issue, I suppose.

ferrarimarco:
Ok, as soon as we've a working `latest`, we can try again with the Stack linter job

admiralAwkbar (approved):

PR #941, opened 2020-11-02
Use the already-defined linter names array to print linter versions

### LINKED ISSUE #879: Use linter commands and linter names arrays to build the versions file
The current implementation of `linterVersions.sh` relies on a duplicated list of linters, instead of relying on the newly implemented associative arrays.

Use the linter commands and linter names associative arrays defined in `linter.sh` instead of the one defined in `linterVersions.sh`. By doing so, we remove yet another data structure that we might forget to update.

N/A for now.

**Additional context**
This should be the natural follow-up of the work done in #781, #802, #803, #815, #854, and #867.

ferrarimarco:
Fixes #879 

## Proposed Changes

1. Use the already-define linter names array while building the file list. To do this, I introduced a new variable (`WRITE_LINTER_VERSIONS_FILE`). When that variable is set to `true`, `linter.sh` runs but exists immediately after building the linter versions file. I didn't document this variable in the README because it's mostly for internal use. We can use a similar approach when we want to reuse the logic in `linter.sh`.
1. Fail hard with a `fatal` if we can't find the version of a linter, instead of a `warn`. This is useful to fail fast and doesn't impact end users because this check runs only when building the image.
1. Correct command debug output text in `worker.sh`, because we `echo` the output of commands, not the commands.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -340,7 +340,7 @@ COPY TEMPLATES /action/lib/.automation
 ###################################
 # Run to build file with versions #
 ###################################
-RUN /action/lib/linterVersions.sh
+RUN ACTIONS_RUNNER_DEBUG=true WRITE_LINTER_VERSIONS_FILE=true /action/lib/linter.sh
 
 ##################################4
 # Run validations of built image #
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -324,6 +324,18 @@ GetLinterVersions() {
   debug "---------------------------------------------"
   debug "Linter Version Info:"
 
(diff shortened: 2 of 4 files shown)


GaboFDC (approved):

admiralAwkbar:
@ferrarimarco Can you fix up the conflicts on this? wasn't sure what was right lol

ferrarimarco:
@admiralAwkbar Done! Rebased against the latest master.

admiralAwkbar (approved):

PR #947, opened 2020-11-03
Fix action versioning

### LINKED ISSUE #708: Why does Versioning not work?
**Describe the bug**
There is [a bug in v3.10.0](https://github.com/github/super-linter/issues/704) so I have not upgraded to this version so am still running 3.9.4.

I still hit the bug.

**To Reproduce**

1. Setup an action with node_modules and lock the version: `uses: github/super-linter@v3.9.4`
2. Lint a repo with a `node_modules` folder. Note this folder gets linted when it previously didn't

**Expected behavior**
Versions should lock the code for stability. It's pointless using version numbers if they are ignored and latest is always run.

**Screenshots**
Example run: https://github.com/HTTPArchive/almanac.httparchive.org/pull/1062/checks?check_run_id=1095951259
As can be seen it's running v3.9.4 but is showing the bug from 3.10.0

**Additional context**
This has been going on a while but annoyed me enough to raise a separate issue for this.

dec5e:
With the current config action's releases are out of use.

Fixes #708. 
Hopefully fixes, I can't test it right now but proposed change matches the [documentation](https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-docker-actions).

## Proposed Changes

1. Remove hardcoded docker tag from action.yaml 

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -3,7 +3,7 @@ author: 'GitHub'
 description: 'It is a simple combination of various linters, written in bash, to help validate your source code.'
 runs:
   using: 'docker'
-  image: 'docker://ghcr.io/github/super-linter:v3'
+  image: 'docker://ghcr.io/github/super-linter'
 branding:
   icon: 'check-square' 
   color: 'white'


action.yml
```diff
@@ -3,7 +3,7 @@ author: 'GitHub'
 description: 'It is a simple combination of various linters, written in bash, to help validate your source code.'
 runs:
   using: 'docker'
-  image: 'docker://ghcr.io/github/super-linter:v3'
+  image: 'docker://ghcr.io/github/super-linter'
```
HorizonNet, line 6: Wouldn't this fall back to the latest version instead of the actual version for a release when you're not using the latest release?

dec5e: Not sure I understood the case you described, could you clarify? What is actual version?

HorizonNet: For example, if I'm on version 3.12.0, wouldn't this not also force me to the latest version?

dec5e: My assumption was that version will be added automatically, but now I'm not sure :thinking: Now it seems to me the version needs to be in the code, so the fix doesn't look that easy anymore

ahmadnassri: I have tested this today on my own actions (after running into #938)... and I can confirm that `action.yml` seem to need versioning the exact version as the released one for it to work, otherwise it will get the latest always.

ahmadnassri:
I really hope the `super-linter` team can talk to the `Actions` team internally at GitHub and figure out what the right syntax for docker action versioning is ... cuz the docs don't specify it, and seemingly event he super-linter team suffered from this lack of clarity too

action.yml
dec5e: Thanks @ahmadnassri. That's sad news

dec5e:
Closing this as proposed solution doesn't work. But agree, it would be nice if `super-linter` and `Actions` teams will join forces and find a good way to fix versioning for this action. And as well to simplify versioning for many other actions based on public docker images.

ahmadnassri:
@dec5e maybe re-open this ... so that the super-linter team doesn't miss all the commnets/info :D

dec5e:
> @dec5e maybe re-open this ... so that the super-linter team doesn't miss all the commnets/info :D

I commented in the original issue #708 instead. Hopefully that'd be enough to bring the issue and related discussion to the team's attention.

PR #959, opened 2020-11-04
fix find

### LINKED ISSUE #954: super-linter includes .git directory in the list of files to lint
**Describe the bug**

```shell
Step #0 - "super-linter": 2020-11-04 10:04:35 [DEBUG]   File:[/workspace/.git], File_type:[git], Base_file:[.git]
Step #0 - "super-linter": 2020-11-04 10:04:35 [WARN]   Failed to get filetype for:[/workspace/.git]!
```

Affects v3.13.5, maybe others.

**To Reproduce**

1. Run super-linter in a Git repository.

**Expected behavior**
super-linter should ignore the `.git` directory [as expected](https://github.com/github/super-linter/blob/master/lib/buildFileList.sh#L98).

admiralAwkbar:
This closes #954 

This is a better way to help remove the `.git` folder from showing up in the list

### DIFF
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -99,7 +99,7 @@ function BuildFileList() {
     -path "*/.venv" -prune -o \
     -path "*/.rbenv" -prune -o \
     -path "*/.terragrunt-cache" -prune -o \
-    -type f 2>&1 | sort )
+    -type f 2>&1 |  grep -v -w '\.git' | sort )
   fi
 
   #######################


admiralAwkbar:
https://stackoverflow.com/questions/10346816/using-grep-to-search-for-a-string-that-has-a-dot-in-it

jwiebalk (approved):

ferrarimarco:
It's interesting that we didn't notice this. What about if we escape the dot? Sounds like a cleaner alternative.

admiralAwkbar:
@ferrarimarco I didn't test that!
I will see if that works, if not, at least this is doing git :)

admiralAwkbar:
@ferrarimarco 
running the command:
```bash
find "/Users/admiralawkbar/github/super-linter" \
-path "*/node_modules" -prune -o \
-path "*/.venv" -prune -o \
-path "*/\.git" -prune -o \
-path "*/.rbenv" -prune -o \
-path "*/.terragrunt-cache" -prune -o \
-type f 2>&1 | sort
```
Still returns `.git` folder....
Testing more :)

ferrarimarco:
and the same should happen with the other dotfiles and dotdirs

admiralAwkbar:
@ferrarimarco I agree, i think we need to add that grep command for the others... seems find hates dots...

ferrarimarco:
I think the right syntax is: `-path ./.git -prune`

ferrarimarco:
Or better: `*/.git`

ferrarimarco:
So no grep needed :)

ferrarimarco:
example: `find . -path */.git -prune -o -print`

ferrarimarco:
I think the double quotes are preventing the shell expansion on the `*`

PR #964, opened 2020-11-05
Fix dotdirs exclusion

### LINKED ISSUE #954: super-linter includes .git directory in the list of files to lint

ferrarimarco:
Fixes #954

## Proposed Changes

1. Fix the exclusion of certain hidden directories that we don't want to lint from the file list.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/buildFileList.sh
+++ b/lib/buildFileList.sh
@@ -94,12 +94,14 @@ function BuildFileList() {
     debug "----------------------------------------------"
     debug "Populating the file list with all the files in the ${WORKSPACE_PATH} workspace"
     mapfile -t RAW_FILE_ARRAY < <(find "${WORKSPACE_PATH}" \
-    -path "*/node_modules" -prune -o \
-    -path "*/.git" -prune -o \
-    -path "*/.venv" -prune -o \
-    -path "*/.rbenv" -prune -o \
-    -path "*/.terragrunt-cache" -prune -o \
-    -type f 2>&1 |  grep -v -w '\.git' | sort )
+    -not \( -path '*/\.git' -prune \) \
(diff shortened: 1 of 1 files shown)


ferrarimarco:
@admiralAwkbar let's move the discussion here. There's something else I don't understand: we shouldn't be including the `.git` directory also because we're specifying `-type f`. There must be something else going on...

ferrarimarco:
I tried to create a `.venv` directory and that gets included as well. Let me think...

admiralAwkbar:
@ferrarimarco There seems to be a bug or caveat we don't get with the `find` command and `.*` dirs...
Thats why i used that grep to get around it... but i didn't like it...

ferrarimarco:
I think that what's happening here is: `find` defaults to the `-print` unless you add another action. In fact, we didn't specify other actions!

ferrarimarco:
Yes, I think we're good to go now :)

GaboFDC (approved):

admiralAwkbar (approved):
awesome

PR #977, opened 2020-11-09
Tsql

### LINKED ISSUE #972: SQL Server Lint Support
Hi, 

We use SQL Server in our codebase, the current sql-lint only supports MySQL and PostgreSQL. It will be really useful if you could add another SQL lint which could check the SQL server.

This one tsqllint works well for SQL Server. 
https://github.com/tsqllint/tsqllint
I'll really appreciate if you could add this lint into Super Lint

Thanks a lot!

admiralAwkbar:
This closes #972 

- Adding TSQL Support


admiralAwkbar:
The error seems to be from a shared lib that's causing other issues as well...
```bash
error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory. Error code: 127
```

admiralAwkbar:
if you open the container the file is there... so I'm a little lost

admiralAwkbar:
So this linter is over 2 years old and doesn't seem supported any longer. It also seems to be built on Windows and Ubuntu and having issues for Alpine based images. At this point, I need to close this as it's not supported

PR #980, opened 2020-11-10
cleanup file

### LINKED ISSUE #969: Why do we have the disable-linters document?
Having the `disable-linters.md` document (which might contain stale information) is a maintenance chore for us.

What if we get rid of that, and just refer to each linter docs?

For example, in that file we state about `klint`: `There is no top level _configuration file_ available at this time`, which is not true as we discovered in #966 (and then by reading klint docs).

Shall we just get rid of that, and update the README and the PR templates accordingly?

admiralAwkbar:
This closes #969

As the documentation grows and adapts, we need to point users to their sources, instead of trying to double the documentation...


IAmHughes (approved):

jwiebalk (approved):

ferrarimarco:
I think there are a couple of places to update as well. From the top of my
head: the pull request template and the readme.

PS: Not strictly related, but I think there's a unused issue/or template
(Enhancement request). Writing here as now I'm not near my workstation 😀

admiralAwkbar:
@ferrarimarco I will check it out and update thanks!

PR #982, opened 2020-11-10
Harden testing

### LINKED ISSUE #949: Add more pre-release testing
The GitHub super-linter has published a number of bad releases since go-live which have been left in place for hours and days. This is particularly a problem as this project does not follow versioning (see #708 ) so this affects even those repos that do not upgrade, and also affects all forks and branches meaning even if we try to [work around a bad release](https://github.com/github/super-linter/issues/938#issuecomment-720977157) it requires all those forks and branches to pick up the latest workaround. This is a big problem when all linting workflows across many projects suddenly break.

The latest example is #938. While that issue can be dedicated to fixing that immediate problem I would like to raise this separate issue to track how this was able to get through to a release. Consider this almost the RCA and follow up actions to try to prevent #938 happening again.

Add test cases to prevent the cause of issues like #938.

The super-linter should also:
- Fix versioning - #708 
- Add post release checks to immediately identify issues on a bad release - #948

**Additional context**
Issues that have broken the linter (not an exhaustive list):
- #704
- #938

ferrarimarco:
Fixes #949

(maybe we can add more tests/validation in this PR before actually closing #949)

## Proposed Changes

1. Fail if a TAP to compare against is not available.
1. Remove `WARNING_ARRAY_TEST` declaration, export, and iteration over because we now fail on each error, instead of printing warnings.
1. Set `ERROR_ON_MISSING_EXEC_BIT` to `true` in our builds.
1. Enable executable bit for scripts that didn't have it set to `1`.
1. Fix file paths while generating the test report: now paths don't include the full path, which might change according to where volumes are mounted.
1. Exclude `.tap` files from the list when validating Dockerfiles.
1. Use `fatal` instead of an explicit `exit 1` to terminate testing.
1. Add tests for:
    1. arm-ttk
    1. bash-exec
    1. Clojure
    1. C#
    1. Editorconfig-checker
    1. Gherkin
    1. Groovy
    1. Java
    1. JSON
    1. Kubeval
    1. Latex
    1. Lua
1. Fix tests for:
    1. bash
    1. cloudformation
    1. Coffeescript
    1. CSS
    1. Dart
    1. Dockerfilelint
    1. Docker hadolint
    1. dotenv-linter
    1. Go
    1. HTML
    1. javascript_es
    1. javascript_standard
    1. Kotlin
    1. Markdownlint
    1. Perl
    1. PHP (built-in)
    1. PHPCS
    1. PHP PSALM
    1. Powershell
    1. Protobuf
    1. Python
    1. Raku
    1. Ruby
    1. States
    1. Terraform
    1. Typescript Standard
    1. XML
    1. yamllint

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
@admiralAwkbar basically what happens is:

1. we have an "expected" TAP report in the test directory.
1. Generate a TAP on the fly for each language.
1. Diff the generated tap against the expected tap.

Example of an expected TAP:

```tap
TAP version 13
1..2
not ok 1 - shell_bad_1.sh
  ---
  message: \nIn /tmp/lint/.automation/test/bash/shell_bad_1.sh line 10 \nif [ $ERROR_CODE -ne 0]; then\n^-- SC1009  The mentioned syntax error was in this if expression.\n   ^-- SC1073  Couldn't parse this test expression. Fix to allow more checks.\n                       ^-- SC1020  You need a space before the ].\n                       ^-- SC1072  Missing space before ]. Fix any mentioned problems and try again.\n\nFor more information \n  https //www.shellcheck.net/wiki/SC1020 -- You need a space before the ].\n  https //www.shellcheck.net/wiki/SC1072 -- Missing space before ]. Fix any m...\n  https //www.shellcheck.net/wiki/SC1073 -- Couldn't parse this test expressi...\n
  ...
ok 2 - /tmp/lint/.automation/test/bash/shell_good_1.sh
```

1. `TAP version 13`: tap version, mandated by the spec
1. `1..2`: test plan. Tells how many tests will be run, or how many tests have run. In this example we see two tests: one for the `good` file, and one for the `bad` file in the `shell` test directory.
1. `not ok 1 - shell_bad_1.sh` and `ok 2 - /tmp/lint/.automation/test/bash/shell_good_1.sh`: test lines. Tell that the test `1` fails (the `bad` file), and the test `2` runs successfully (the `good` file).
1. The YAML block. This is not standardized in the spec, so we have complete freedom. In super-linter, someone made the choice to use the format:

    ```yaml
    ---
    message: ABCDEFG.....
    ...
    ```

    We then use this YAML to save the expected output message, or the actual output message.

There are issues in this approach when the output has dynamic parts. I suspect that's the reason why some languages (such as arm-ttk) have a TAP file, but with an `ignored` extension.

ferrarimarco:
@admiralAwkbar almost there. I'm missing a regex and fixing a weird formatting issue with the powershell test.

ferrarimarco:
@admiralAwkbar finally ready for review.

admiralAwkbar (approved):
Amazing work!@

bazzadp:
Great work @ferrarimarco ! Dunno what it all means to be honest (TAP?) but hopefully will help catch things in future and looks like a huge amount of work, so thanks!

ferrarimarco:
LOL. TAP = [Test Anything Protocol](https://testanything.org/). It's a semi-structured protocol for test reporting. We're using it in super-linter to check if the test results match our expectations.

PR #983, opened 2020-11-10
testing and fixes

### LINKED ISSUE #949: Add more pre-release testing

admiralAwkbar:
This closes #949 

- Adding more testing
- Set exec bit to true vs false
- Fix tap reporting


admiralAwkbar:
@ferrarimarco I dont understand tap all that well and would love some help getting those up to date again... its crazy///

ferrarimarco:
#983 :) We're working on the same stuff

ferrarimarco:
With the exec bit I went a different route to avoid breaking existing users. I enabled that behavior in our build configs

admiralAwkbar:
@ferrarimarco lol nice! ill throw this pr away! i have another fix i want to do :)

ferrarimarco:
Feel free to commit there! :)

PR #993, opened 2020-11-12
Try to build dev with self-hosted runner

admiralAwkbar:
This should greatly help with the build times...

- We will use a self-hosted runner to run the deploy dev
- This will allow caching of the image
- This should greatly reduce build times

### DIFF
--- a/.github/workflows/deploy-DEV.yml
+++ b/.github/workflows/deploy-DEV.yml
@@ -30,7 +30,7 @@ jobs:
     # Name the Job
     name: Deploy Docker Image - DEV
     # Set the agent to run on
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     # Prevent duplicate run from happening when a forked push is committed
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     ##################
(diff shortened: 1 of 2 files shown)


admiralAwkbar:
# Build with no-cache
- 12+ minutes

# Build with cache
- 8+ minutes

# Rebuild with cache
- ???

admiralAwkbar:
@ferrarimarco this should cut our build time in half for CI runs.. not a bad start...

admiralAwkbar:
@nemchik @GaboFDC any concerns with this approach?

ferrarimarco:
Nice!

GaboFDC:
This is nice for this repo, but the caching problem will still be up everywhere that super-linter is used.
Also, can you describe who maintains this runner, where is located, etc. Just to be aware of the details.

admiralAwkbar:
@GaboFDC Great questions.
So the use of this is only for the dev development, which helps us make pushes faster to the source code.

The downside:
- Machine is hosted in AWS under GitHub
  - So not a lot of control externally for this machine
- I would likely have to be the one to maintain the machine and keep it upgraded

## Overall
I think this might not be the best idea, as it makes it hard to maintain for everyone, and could eventually be problematic...
I think it was nice to prove we can speed this up, but we need to use some sort of docker cache to help us. This is just not the best way

I think i will close this up :)

PR #1015, opened 2020-11-20
Fixes Terrascan to perform checks for all supported Terraform cloud providers

### LINKED ISSUE #692: Terrascan runs AWS checks only
**Describe the bug**
Terrascan is invoked with the `-t aws` option. This causes Terrascan check Terraform code against a subset of rules, which is not ideal.

**To Reproduce**

1. Run super-linter against Terraform code
2. See that no violations are reported against Google Cloud- or Azure- specific checks.

**Expected behavior**
Terrascan should not be limited by an arbitrary set of rules or, at least, this should be configurable by the user.

**Screenshots**
See https://github.com/github/super-linter/blob/c58bab262756890004e5bbb3cf8744ed1e6174c4/lib/linter.sh#L1909

**Additional context**
The `-t aws` switch was introduced in #632

cesar-rodriguez:
Fixes #692

## Proposed Changes

This updates the Terrascan flags to explicitly use Terraform and scan all supported cloud providers. 

### Author/Contributor

### Reviewing Maintainer


Dockerfile
```diff
@@ -16,7 +16,7 @@ FROM koalaman/shellcheck:v0.7.1 as shellcheck
 FROM wata727/tflint:0.20.3 as tflint
 FROM alpine/terragrunt:0.13.5 as terragrunt
 FROM mvdan/shfmt:v3.2.0 as shfmt
-FROM accurics/terrascan:d182f1c as terrascan
+FROM accurics/terrascan:2d1374b as terrascan
```
ferrarimarco, line 19: Can you update this to `d37fb58`? Thanks!

ferrarimarco (changes requested):

Dockerfile
ferrarimarco: Nevermind. 3.2.0 maps to `2d1374b`

ferrarimarco:
@admiralAwkbar Dependabot is not proposing updates for terrascan because (I suppose), they use Git hashes only to tag their images, and we're using one of that. I've opened https://github.com/accurics/terrascan/issues/398 to ask for that. Let's see how it goes.

admiralAwkbar:
@ferrarimarco I'm at a loss on how to solve our fun docker issues...

ferrarimarco:
I think hadolint is [getting confused](https://github.com/hadolint/hadolint/blob/73081a9f21799906891c2133b8119cd0c456ded3/src/Hadolint/Rules.hs#L1125) because it looks for commands starting with `pip`, such as `pipenv`.

`--no-cache-dir` is not even a [valid pipenv option](https://pipenv.pypa.io/en/latest/cli/#pipenv)...

I guess we've to disable that rule on that line for the moment. I'll open an issue in the hadolint tracker

ferrarimarco:
Ok, apparently we're not alone: https://github.com/hadolint/hadolint/issues/366

My suggestion is to disable `DL3042` on that Dockerfile line, and add a comment explaining why.

I've subscribed to that hadolint issue to keep us updated.

admiralAwkbar:
@ferrarimarco i got it!
So seems we needed some underlying libs, then fixed the hadolint issue... Sorry for all the noise in here...

ferrarimarco:
Nice work!

admiralAwkbar (approved):

PR #1051, opened 2020-12-03
Try to build with caching to help build times

admiralAwkbar:
- Trying out action to use docker cache to help build dev times


admiralAwkbar:
So for now, were not seeing any major gains in speed, but I would assume once this is working in sync, we should see better build times as caching the image is a good idea...

Also, it takes us away from the custom bash scripts and more of an actions approach...

Not a bad idea...

jwiebalk (approved):

PR #1111, opened 2020-12-30
Improved ANSIBLE_DIRECTORY validation to support the repository root

### LINKED ISSUE #1110: ANSIBLE_DIRECTORY will always fail to match if playbook is at the top-level
## Describe the bug

If `ANSIBLE_DIRECTORY` is set to `/` or some other hacky value to force the container to resolve to `/tmp/lint/`, it will always fail to match `$(dirname /tmp/lint/site.yml)` to `/tmp/lint/` since `dirname` returns `/tmp/lint` (no trailing slash)

## Expected behavior

Because a top-level playbook file is fairly common, I should be able to set my `ANSIBLE_DIRECTORY` to something use a repository root path and have my files identified as Ansible-related. 

## Steps to Reproduce

1. Set `ANSIBLE_DIRECTORY` to `/`
1. Run super-linter either locally or in Actions. For example:
```
docker run \
  -e RUN_LOCAL=true \
  -e ANSIBLE_DIRECTORY="/" \
  -v $HOME/dotfiles:/tmp/lint \
  github/super-linter:latest
```
1. See the debug line in the logs:
```
[0m2020-12-30 21:05:56 [0;34m[DEBUG][0m   Checking if /tmp/lint/dotfiles.yml is an Ansible file. Ansible directory: /tmp/lint/...[0m
[0m2020-12-30 21:05:56 [0;34m[DEBUG][0m   /tmp/lint/dotfiles.yml is NOT an Ansible-related file.[0m
```

## Logs

I've attached logs from my most recent run, using the docker command above.

[super-linter.log](https://github.com/github/super-linter/files/5755729/super-linter.log)

## Additional context

I believe the culprit is this line here: https://github.com/github/super-linter/blob/7943ce1f541fc84bcc0789028dc73aee047f72b3/lib/functions/validation.sh#L143

And for good measure, here's my repository: https://github.com/aaronsky/dotfiles. I'm trying to get `ansible-lint` to pick up `dotfiles.yml`, which is my main playbook.

aaronsky:
Fixes #1110, which prevents linting playbooks at the root of the repository from being supported.

## Proposed Changes

1. Support `ANSIBLE_DIRECTORY` values `"."`, `"/"`, and `"/${{ github.workspace }}"`
2. Update Ansible lint command to change into the `$ANSIBLE_DIRECTORY`, rather than the hardcoded `${GITHUB_WORKSPACE}/ansible` location. 
3. Changes should be purely additive and backwards compatible.

I tested these changes by building my fork locally in Docker and running several commands against my repo. [Logs are attached](https://github.com/github/super-linter/files/5755845/super-linter.log) for each run, plus some others not listed below:

```
docker run -e RUN_LOCAL=true -e ANSIBLE_DIRECTORY="//tmp/lint" -v $HOME/dotfiles:/tmp/lint 296d67fdca4c
docker run -e RUN_LOCAL=true -e ANSIBLE_DIRECTORY="/" -v $HOME/dotfiles:/tmp/lint 0534dd43cb81
docker run -e RUN_LOCAL=true -e ANSIBLE_DIRECTORY="/horace" -v $HOME/dotfiles:/tmp/lint 0534dd43cb81
docker run -e RUN_LOCAL=true -e ANSIBLE_DIRECTORY="" -v $HOME/dotfiles:/tmp/lint 0534dd43cb81
docker run -e RUN_LOCAL=true -e ANSIBLE_DIRECTORY="." -v $HOME/dotfiles:/tmp/lint 0534dd43cb81
```

### Author/Contributor

### Reviewing Maintainer


Vowakula82 (approved):

Vowakula82 (dismissed):

Vowakula82 (approved):

Vowakula82 (approved):

lib/functions/worker.sh
```diff
@@ -195,7 +195,7 @@ function LintCodebase() {
         # Lint the file with the rules #
         ################################
         LINT_CMD=$(
-          cd "${WORKSPACE_PATH}/ansible" || exit
+          cd "${ANSIBLE_DIRECTORY}" || exit
```
ferrarimarco, line 198: Looks like a legit fix. Good catch!

lib/functions/validation.sh
```diff
@@ -139,8 +139,19 @@ function GetValidationInfo() {
       # Remove first char
       ANSIBLE_DIRECTORY="${ANSIBLE_DIRECTORY:1}"
     fi
+
+    # Check if user is explicitly referring to the workspace dir
+    if [[ ${ANSIBLE_DIRECTORY} == "." ]] || [[ ${ANSIBLE_DIRECTORY} == "${GITHUB_WORKSPACE}" ]]; then
```
ferrarimarco, line 144: The second condition implies that `ANSIBLE_DIRECTORY` might not be relative to `GITHUB_WORKSPACE`, but it is!

ferrarimarco (changes requested):

ferrarimarco:
I think this should be a matter of setting `ANSIBLE_DIRECTORY=.` and handling just this corner case.

lib/functions/validation.sh
aaronsky: At this point, it's still unsanitized input, so it isn't treated as relative yet. So if someone really wanted to explicitly pass their `GITHUB_WORKSPACE`, I added a condition to express that correctly. I can remove this if you'd prefer not to support that, since it could constitute undefined behavior.

Vowakula82 (approved):

lib/functions/validation.sh
aaronsky: I think I see your point better now. I've pushed a new commit, let me know if that sorts this out and if my accompanying docs change makes sense. Thanks!

lib/functions/validation.sh
```diff
@@ -139,8 +139,19 @@ function GetValidationInfo() {
       # Remove first char
       ANSIBLE_DIRECTORY="${ANSIBLE_DIRECTORY:1}"
     fi
+
+    # Check if user is explicitly referring to the workspace dir
+    if [[ ${ANSIBLE_DIRECTORY} == "." ]]; then
```
ferrarimarco, line 144: I'd rather not change the value of `ANSIBLE_DIRECTORY`. Can you just use a single `if` and set `TEMP_ANSIBLE_DIRECTORY` directly? Perhaps I'm missing something (late in my timezone LOL).

ferrarimarco (changes requested):

lib/functions/validation.sh
aaronsky: Sure, let me update this real quick. 👍🏼 

Also if it's late for you, we can leave it until tomorrow or to someone else 🙂 This is only blocking my use of super-linter on a personal project.

Vowakula82 (approved):

Vowakula82 (approved):

admiralAwkbar (approved):
very nice!

aaronsky:
Thanks for the approval @admiralAwkbar! @ferrarimarco Is there anything else you want changed before this can be merged?

PR #1118, opened 2021-01-04
Adding support for Prettier

### LINKED ISSUE #175: Prettier for JavaScript and TypeScript
IIUC, super-linter uses ESLint and Standard to format JavaScript and TypeScript code rather than Prettier. I think many developers will have an issue with this. Despite the name, Standard is not a standard. It's developers just gave it that name. For the past few years the trend has been to use ESLint only to check for issues and to use Prettier for formatting.

Most developers will be fine with most of the formatting that Standard performs. However, there is one controversial thing that it does. It enforces a space after function names in their definition. For example:

```js
function myName (param1, param2) {
  ...
)
```

I think you will find that almost no JavaScript developer does this. So using Standard will make this change across their code bases and they will not like it.

Please reconsider the use of Standard for JavaScript formatting and instead use Prettier, which is by far the more popular option.

admiralAwkbar:
This closes #175

Trying to add this and default to standard... should be interesting...


admiralAwkbar:
@jwiebalk we did it lol

jwiebalk (approved):

PR #1132, opened 2021-01-09
Fix golang linting to use directories instead of files

### LINKED ISSUE #143: Invoking golangci-lint run on individual files breaks linters in packages.

mikegfink:
Fixes #143 

## Proposed Changes
Preface: As far as I can tell, the go linter is non-functional at the moment. If you have multiple files per package then linting fails, I don't think I've seen a project with one go file per package. If you have a repository with multiple languages, linting fails because golangci-lint is run from the workspace directory and gets confused. Given that, take the downsides of the PR as they are, but note it might let folks actually use super-linter for go.

1. Lint from DIR_NAME without a FILE_NAME.
The linter for go did not work if any packages contained multiple files that referenced one another. 
So, when linting the whole codebase, we now lint from the directory of each go file.

2. When VALIDATE_ALL_CODEBASE is false, pass the --new-from-rev flag to golangci-lint.
When only linting the most recent change, we make use of golangci-lint's new-from-rev to only lint based off the changes in the GIT_SHA.

3. Use --skip-files to skip files in FILTER_REGEX_EXCLUDE.
Since we are linting via directories, we need to pass the exclude list along. A user might have problems if they were using FILTER_REGEX_INCLUDE because while those files will get linted, so will everything else in the directory. This certainly seems better than not being able to use it at all.

Note that this results in linting a directory more than once if there is more than one file in that directory.
Say you have the following structure:
```
module/
  go.mod
  main.go
  package1/
     package1.go
  package2/
    package2.go
    package3/
      package3.go
```

This PR will result in each directory of `module`, `package1`, `package2` and `package3` getting linted. Which means that package1 and package2 will be linted twice and package3 linted three times.

One should be able to prevent this by using `--skip-dirs` and including each directory as it is linted. And I've tried, but bash scripting is not my friend. Something like:
```
if [ ! ${FILE_DIRS_GO[$DIR_NAME]+_} ]; then
          CMD=""
          if [[ -v FILE_DIRS_GO[@] ]]; then
            for i in "${!FILE_DIRS_GO[@]}"
            do
              CMD="${CMD} --skip-dirs '$i(\$|/)'"
            done
          fi
          if [[ -n "$FILTER_REGEX_EXCLUDE" ]]; then
            CMD="${CMD} --skip-files ${FILTER_REGEX_EXCLUDE}"
          fi
          CMD="${LINTER_COMMAND}${CMD}"
          LINT_CMD=$(
            cd "${DIR_NAME}" || exit
            ${CMD} 2>&1
          )
          FILE_DIRS_GO[$DIR_NAME]=1
        else
            info "${FILE} already linted via ${DIR_NAME}"
        fi
```
In worker.sh. But not exactly that because that doesn't actually run the CMD for some reason that is beyond me.

### Author/Contributor

### Reviewing Maintainer


mikegfink:
The workflow check failure doesn't appear to have anything to do with this change.

GaboFDC:
> The workflow check failure doesn't appear to have anything to do with this change.

There are some linting errors on your change, check the log. you are missing some `fi`s and other stuff.

Aside from that, it seems like a viable change, seems like the go linter has a very specific behavior. even receiving the sha for went linting a change. Not sure if this logic will conflict with the general one we have in place, did you test this locally?

Yimsiyouna:
XVIDEO.COM Sent from my Galaxy
-------- Original message --------From: Gabriel Diaz <notifications@github.com> Date: 1/13/21  21:54  (GMT+07:00) To: github/super-linter <super-linter@noreply.github.com> Cc: Subscribed <subscribed@noreply.github.com> Subject: Re: [github/super-linter] Fix golang linting to use directories instead of files (#1132) 

The workflow check failure doesn't appear to have anything to do with this change.

There are some linting errors on your change, check the log. you are missing some fis and other stuff.
Aside from that, it seems like a viable change, seems like the go linter has a very specific behavior. even receiving the sha for went linting a change. Not sure if this logic will conflict with the general one we have in place, did you test this locally?

mikegfink:
> > The workflow check failure doesn't appear to have anything to do with this change.
> 
> There are some linting errors on your change, check the log. you are missing some `fi`s and other stuff.
> 
> Aside from that, it seems like a viable change, seems like the go linter has a very specific behavior. even receiving the sha for went linting a change. Not sure if this logic will conflict with the general one we have in place, did you test this locally?

Thanks, I didn't see the linter issues somehow, but running super-linter locally on the super-linter repo picked them up. I'm going to hold the fixes though because for some reason this PR isn't working locally anymore. The golang linter is seeing the issues but they are only getting printed to the super-linter.log and not causing the job to fail.

By the way, testing locally was really difficult to figure out. I ended up pulling the docker image, starting a container, manually making my changes in the docker container, then doing docker run via that same container to see if it worked. Then copying those changes out to the actual codebase. There must be a better way.

MiladHassani73 (approved):

MiladHassani73:
> Fixes #143
> 
> ## Proposed Changes
> Preface: As far as I can tell, the go linter is non-functional at the moment. If you have multiple files per package then linting fails, I don't think I've seen a project with one go file per package. If you have a repository with multiple languages, linting fails because golangci-lint is run from the workspace directory and gets confused. Given that, take the downsides of the PR as they are, but note it might let folks actually use super-linter for go.
> 
> 1. Lint from DIR_NAME without a FILE_NAME.
>    The linter for go did not work if any packages contained multiple files that referenced one another.
>    So, when linting the whole codebase, we now lint from the directory of each go file.
> 2. When VALIDATE_ALL_CODEBASE is false, pass the --new-from-rev flag to golangci-lint.
>    When only linting the most recent change, we make use of golangci-lint's new-from-rev to only lint based off the changes in the GIT_SHA.
> 3. Use --skip-files to skip files in FILTER_REGEX_EXCLUDE.
>    Since we are linting via directories, we need to pass the exclude list along. A user might have problems if they were using FILTER_REGEX_INCLUDE because while those files will get linted, so will everything else in the directory. This certainly seems better than not being able to use it at all.
> 
> Note that this results in linting a directory more than once if there is more than one file in that directory.
> Say you have the following structure:
> 
> ```
> module/
>   go.mod
>   main.go
>   package1/
>      package1.go
>   package2/
>     package2.go
>     package3/
>       package3.go
> ```
> 
> This PR will result in each directory of `module`, `package1`, `package2` and `package3` getting linted. Which means that package1 and package2 will be linted twice and package3 linted three times.
> 
> One should be able to prevent this by using `--skip-dirs` and including each directory as it is linted. And I've tried, but bash scripting is not my friend. Something like:
> 
> ```
> if [ ! ${FILE_DIRS_GO[$DIR_NAME]+_} ]; then
>           CMD=""
>           if [[ -v FILE_DIRS_GO[@] ]]; then
>             for i in "${!FILE_DIRS_GO[@]}"
>             do
>               CMD="${CMD} --skip-dirs '$i(\$|/)'"
>             done
>           fi
>           if [[ -n "$FILTER_REGEX_EXCLUDE" ]]; then
>             CMD="${CMD} --skip-files ${FILTER_REGEX_EXCLUDE}"
>           fi
>           CMD="${LINTER_COMMAND}${CMD}"
>           LINT_CMD=$(
>             cd "${DIR_NAME}" || exit
>             ${CMD} 2>&1
>           )
>           FILE_DIRS_GO[$DIR_NAME]=1
>         else
>             info "${FILE} already linted via ${DIR_NAME}"
>         fi
> ```
> 
> In worker.sh. But not exactly that because that doesn't actually run the CMD for some reason that is beyond me.
> 
> ## Readiness Checklist
> ### Author/Contributor
> * [x]  If documentation is needed for this change, has that been included in this pull request
> 
> ### Reviewing Maintainer
> * [ ]  Label as `breaking` if this is a large fundamental change
> * [ ]  Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #1179, opened 2021-02-02
Fix the default hadolint rule name in the documentation

JonZeolla:
Fixes the default hadolint rule name in the documentation.  `linter.sh` has a variable default of `.hadolint.yaml` ([link](https://github.com/github/super-linter/blob/ab8780a58f056cc470449cf96a41982c742dcaf8/lib/linter.sh#L91)) and that is the default for the `hadolint` project, so updated the docs.

### Author/Contributor

### Reviewing Maintainer
- [X] Label as `breaking` if this is a large fundamental change


JonZeolla:
/label `documentation`

(I don't have access to set it myself)

JonZeolla:
Looks like there is a secondary file lookup for `hadolint` which was how the file in `.github/linters/` was still getting picked up previously.

admiralAwkbar (approved):
Thanks for the catch!

Issue #786, opened 2020-09-30
Support ignore experience similar to .gitignore


migolfi:
**Is your feature request related to a problem? Please describe.**

I love that I can integrate "One linter to rule them all", and with that comes the file ignoring challenge.
Recently we've seen generated code added to our repo and the lint errors which come with it.
The generated code is absolutely needed for our tests, and so it needs to be maintained in the repo.
I added the `FILTER_REGEX_EXCLUDE` variable to ignore the files under that path,
although it leaves a lot to be desired to match the experience in ignoring files which is supported
by Git and several other tools.

**Describe the solution you'd like**

I would like have an experience similar to having a `.superlinterignore` file, which can be at
any directory level and allow that to determine which files are added or removed from the list
of files needing to be linted.

**Describe alternatives you've considered**

Currently, the `FILTER_REGEX_EXCLUDE` will work for single paths, and I can see it becoming very
hard to read and even maintain if several paths need to be ignored.

ferrarimarco:
And you might just symlink your `.gitignore` to point to `.superlinterignore` (much like we're doing with the `.dockerignore` in this repo), and fallback reading the `.gitignore` file directly...

PR #1185, opened 2021-02-04
Ignore files that are ignored by Git

ferrarimarco:
## Proposed Changes

1. Ignore files that Git ignores (related to #786)
2. Fix setting the Ansible directory path when running tests, so that super-linter can correctly build the file list.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
@admiralAwkbar do we want to make this customizable with an environment variable? My gut feeling is that we can enable this by default, because PR is useful only for non-pristine workspaces.

GaboFDC:
Just a thought, if the files are ignored by git, they will not end up in the repo, therefore when the action runs and checkouts the code, the files should not be present.  Or is there something I'm missing?

ferrarimarco:
That's why I added my previous comment :)

ferrarimarco:
The build failure is not related to this PR :(

admiralAwkbar:
@ferrarimarco @GaboFDC i like the simple idea that if git is going to ignore the file, so should we... I think we can go forward without a flag and keep this very simple... i would love to hear the argument for validating a file that git ignores :)

admiralAwkbar (dismissed):

nemchik:
Question:

With this implementation I believe if a file is ignored by .gitignore it will still output a debug message that the file was found and skipped. I'm not against this, but curious how useful this is? The reason I ask is that if we don't care about logging those files we could probably use `git -C "${GITHUB_WORKSPACE}" ls-tree -r HEAD | awk '{print $4}'` to create the list of files that should be included rather than worrying about what should be excluded. It would also be possible to create the include and exclude list separately and simply debug log the exclude list all at once and then process all of the include list.

admiralAwkbar:
@nemchik 
Thats a really good point.

Im up for either answer on this... and trying to think of more reasons this could be helpful to go the route you propose...

nemchik:
https://github.com/github/super-linter/pull/1192 is based on this branch and targeted to merge into this branch. Technically if I were to make that change to master instead (so the changes from this branch are not included) the changes from this branch would not be required. If https://github.com/github/super-linter/pull/1192 merges the logic here https://github.com/github/super-linter/pull/1185/files#diff-12d4345f470b2a2f77d8ab44d4522f6ca5b0ac1b17fbd1ce0b330b96fa097a11R260-R264 will never fire because the files it's looking for won't be included in the loop of `RAW_FILE_ARRAY` at https://github.com/github/super-linter/blob/90547ac39a99d8c3105c2bbbe59f71f0ce8f4aae/lib/functions/buildFileList.sh#L211

If https://github.com/github/super-linter/pull/1192 merges to this branch, we should consider adding an environment variable that would add `GIT_IGNORED_FILES` to `RAW_FILE_ARRAY` so that the current behaviour in the master branch (files ignored by `.gitignore ARE linted) can be achieved if a user requires it.

ferrarimarco:
Yeah, I think so as well. I recently came across a use case where I had to lint the sources of a website which are the results of an execution of a static site generator. These generated sources are ignored by Git.

Without an environment variable, there's no way of supporting this behavior.

ferrarimarco:
Ok, I added the variable, defaulting to `false`, so that it's a non-breaking change.

GaboFDC:
> 
> Yeah, I think so as well. I recently came across a use case where I had to lint the sources of a website which are the results of an execution of a static site generator. These generated sources are ignored by Git.
> 
> Without an environment variable, there's no way of supporting this behavior.

I still fail to see the usability, if you lint files that are automatically generated, then do u manually edit them?  it defeats the purpose of the files being automatically generated.

Also, we do support using the action locally, but still, the expected behavior should be meant to use as an action.
IMHO adding some extra logic and covering corner cases for local usage, doesn't seem like a good approach.

ferrarimarco:
Example: how do you ensure that a template output is correct HTML? You don't edit the output, but you can fix the template

admiralAwkbar (approved):

PR #569, opened 2020-08-16
Add lintr support to super-lint

assignUser:
Closes #262

Hello!
R support for super-linter would be great, so I gave it a try.  I have it (mostly) working using the only (afaik) R linter [lintr](https://github.com/jimhester/lintr) 👍 I followed the Wiki guide/example pr #185 
I have no prior experience with docker, so please let me know if I did not adhere to best practice etc..

As installing and compiling `lintr` and it's dependencies takes a very long time I created an image to use as a build stage: https://github.com/assignUser/lintr-lib/ 
Maybe this could/should be moved into the super-linter Dockerfile?

Linting is working but for some reason the output is formatted in a weird way and I am unsure how to fix it:
image

## Proposed Changes

1. added lintr support for R/Rmd Files
2. added test cases
3. added documentation

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer 
- [ ] Label as `breaking` if this is a large fundamental change
- [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #1233, opened 2021-02-18
only error on rstats lintr errors, not all lints

### LINKED ISSUE #1232: rstats lintr errors on all lints
## Describe the bug

the lintr linter for rstats errors on *all* lints...

## Expected behavior

... should error only on errors.

## Source

I'm skipping the reprex, because the code is pretty straightforward about the behavior:

https://github.com/github/super-linter/blob/b6bbc88e8e3ce21da52df376e0c3c080cb331c59/lib/functions/worker.sh#L221

`lintr::lint()` will always return all lints, not just errors, so this behavior might be a bit trigger happy.

I'm assuming that the general assumption for linters in super-linter is that they carry forward the error/warning/info etc. designations of the upstream linters.

maxheld83:
Fixes #1232.

paging @assignUser who added lintr support in the first place 🙏  in #569.

lintr does not have a native capability to filter by type, barring https://github.com/jimhester/lintr/issues/309, so we need some purrs action, which happily is included in the image.

## Proposed Changes

1. what it says on the tin; only errors on errors.

### Author/Contributor

No docs changes necessary.

### Reviewing Maintainer


admiralAwkbar:
@maxheld83 i like it
trying to get tap to pass... its been a mess lately

maxheld83:
the failing check is unrelated to my PR though, right?

admiralAwkbar:
@maxheld83 
the tap is failing with message:
```
message: Warning message \nIn options(stringsAsFactors = TRUE)  \n  'options(stringsAsFactors = TRUE)' is deprecated and will be disabled\nError in loadNamespace(name)   there is no package called purrr\nCalls     ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart\nExecution halted\n
```

admiralAwkbar:
@maxheld83 @assignUser 
It seems we pull in R from
`FROM ghcr.io/assignuser/lintr-lib:0.1.2 as lintr-lib`

Would we need to do a separate install of purr or pull from a diff image??

assignUser:
> Would we need to do a separate install of purr or pull from a diff image??

I will update the image with purrr when I have a moment.

maxheld83:
sorry, my bad. 

I was sloppy and had yesterday only tested:

`docker run ghcr.io/assignuser/lintr-lib:latest Rscript -e "purrr::map(c(1), sum)"` which worked alright.

So it seems that purrr was *already* in `latest` as of yesterday, @assignUser.

assignUser:
@maxheld83 up until now `latest`and `v0.1.2` where identical (at least I thought so^^). I updated the image to include purrr in v0.2.0. 
Update your pr to: `FROM ghcr.io/assignuser/lintr-lib:0.2.0 as lintr-lib` that should do it.

maxheld83:
argh, shoot, still have to update the TAP snapshots I guess ...

maxheld83:
yikes, finally all passing.

Here's what I was missing:

- the TAP snapshot had an unrelated diff, caused by the change of the R version, I think? (something something `stringsAsFactors`)
- previously, the lintr would error out on *every lint*, not just errors (changing that is the point of this PR).
     Now with the more generous behavior, the test failed because the `bad.R` actually had caused no *errors*, just lints.
     So now there's an error in there and the snapshot has also been changed.

admiralAwkbar (approved):

PR #1240, opened 2021-02-19
Fix Tap report spacing

admiralAwkbar:
It appears as the linters update, spacing changes, and causes failures...
If we store the tap and trim the spaces, and then trim at run time, then we should have more standardized returns

This should help stop some of the tap failures...


admiralAwkbar:
Basicaslly, we ran `sed -i 's/ \{1,\}/ /g' "$SOME_TAP_FILE"` against all tap reports
And at report generation, we do the same to the expected return value to make sure they are both trimmed...

ferrarimarco:
What if we avoid to delete new lines? It makes a bit hard to spot differences with the expected report.

jwiebalk (dismissed):

jwiebalk (approved):

PR #1241, opened 2021-02-20
added RUN for clean up ~700mb

clemenko:
Fixes #

Added a clean up line to the bottom of the Dockerfile. This will save roughly 700mb when used with the following build command. Next pass would be looking at a true multistage build to move only what is needed for the resulting image.

```bash
docker build -t superlinter --squash .
```

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
Doing that cleanup this way doesn't help because the data you're cleaning up will still get downloaded as part of other layers, unless you squeeze layers before pushing the image to the registry (as you pointed out). But then, we cannot benefit from the Docker cache.

I think this should not be merged, but @admiralAwkbar may want to have a look.

clemenko:
Agreed. But Since you are using about 10 different images and about 15 different layers there really is no benefit from cache. Plus saving 700mb is a big deal. Ideally a true multistage image should be built. But it will take some serious in-depth look at all the packages that are needed. Thus this is a first pass.

-
andy 410-212-3200
________________________________

admiralAwkbar:
@clemenko Hey Andy thanks for the look on this!

We definitely need some help getting this image under some control...

After adding all the linters and base libs, I think we sit around 4gb... Anything we can do to help speed up the build process, the size of the image would be greatly appreciated.

admiralAwkbar:
@clemenko I think the answer to this is a multistage build, but that gets into docker areas that I'm much less versed.
Would love to collaborate on this and see what we could do for this image and the community

thanks!

clemenko (commented):
LGTM

zkoppert:
@clemenko I reached out with a google calendar invite. Thanks for helping out with this!

clemenko:
joining!

zkoppert:
multi-stage build issue opened at https://github.com/github/super-linter/issues/1458

zkoppert:
Should we close this PR or leave it open? It's unclear to me @ferrarimarco and @admiralAwkbar if we want to merge this while we wait for the multistage build to get traction.

ferrarimarco:
I think this should be closed because it changes nothing in terms of size, unless you pair it with squeezing the final image, which is (IMHO) an anti-pattern in the vast majority of cases.

zkoppert:
I think that this is helpful and we should move forward with it. @admiralAwkbar or @ferrarimarco, any objections?

zkoppert (approved):

ferrarimarco:
The issue with this approach is that this change does NOTHING to reduce the image size, and makes it harder to understand the Dockerfile, because it might confuse people reading it, and actually thinking that it helps.

clemenko:
with the new approach on #1458 I am cool with closing this MR.

PR #1259, opened 2021-02-24
Add logic to check command output before maping the files

### LINKED ISSUE #1258: Failed to get diff of files for lint
## Describe the bug

There is an error when checking the changed files:
`File:{fatal: master...dad356fc1a8f8fe0ed036fb518762fd1f32a9c2b: no merge base} existed in commit data, but not found on file system, skipping...`

This means there are 2 bugs. 
The command is failing fatally, but the script continues to work as if it didn't fail.
And also, the command is broken in some way and not finding the merge base.

## Expected behavior

The linter should continue to work as expected. We have yet another broken release :/

## Steps to Reproduce

I think the problem is related to squash commits. But also the way it is being used, makes that if the command fails, it is not properly handled.

GaboFDC:
Fixes #1258

## Proposed Changes

1. Added logic to run the command and check for errors, before mapping the files to an array, to avoid silent failures of the commands.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/functions/buildFileList.sh
+++ b/lib/functions/buildFileList.sh
@@ -62,16 +62,39 @@ function BuildFileList() {
       ################
       # push event   #
       ################
+      DIFF_TREE_CMD="git diff-tree --no-commit-id --name-only -r ${GITHUB_SHA}"
+
       ################
       # print header #
       ################
       debug "----------------------------------------------"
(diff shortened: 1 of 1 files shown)


Sal430 (approved):

admiralAwkbar (approved):
Awesome

PR #1272, opened 2021-02-26
adding sha info

admiralAwkbar:
Getting sha info to help pass status check


admiralAwkbar:
Now that we get the commit sha that was created when the pr was created, I can call the status API and pass the status to the pr and provide success to the required status.

This will then allow for further autmotaion.

This is an issue because an action is calling another action and I need to move forward...

PR #1305, opened 2021-03-04
Fix get file diff on pr event

GaboFDC:
Fixes #

## Proposed Changes

1. Fix diff for or event.
2. Also we don't need to checkout the main branch. The diff commands work fine being in the same brnach.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -480,7 +480,7 @@ GetGitHubVars() {
     # Github sha on PR events is not the latest commit.
     # https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
     if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
-      GITHUB_SHA=$(cat $GITHUB_EVENT_PATH | jq -r .pull_request.head.sha)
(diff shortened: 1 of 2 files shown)


lib/linter.sh
```diff
@@ -474,6 +474,15 @@ GetGitHubVars() {
     ######################
     GITHUB_ORG=$(jq -r '.repository.owner.login' <"${GITHUB_EVENT_PATH}")
 
+    ########################
+    # Fix SHA for PR event #
+    ########################
+    # Github sha on PR events is not the latest commit.
+    # https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
+    if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
+      GITHUB_SHA=$(cat $GITHUB_EVENT_PATH | jq -r .pull_request.head.sha)
```
ferrarimarco, line 483: Didn't check the logs yet, but shellcheck should complain about missing quotes here.

ferrarimarco (changes requested):

lib/linter.sh
GaboFDC: Yes it did, I improved it just now

PR #1306, opened 2021-03-05
Support markdown custom rules

tpansino:
## Proposed Changes

1. Adds support for [custom rules](https://github.com/DavidAnson/markdownlint/blob/main/doc/CustomRules.md) in `markdownlint`.

    A bit of background: unlike most CLI tools, which allow for setting of options via either command line switches, a config file, or both, `markdownlint-cli` makes a distinction between "options" and "config". Options are _ONLY_ configurable via CLI switches, not via the config file, and this includes the option for specifying a list of paths to check for custom rules.

    This means that the only way to specify custom rule sets is to modify the command being executed by the Super Linter, so that's what I've attempted, though not as cleanly as I would have hoped.
    
    I also did not see a great way to add tests for this functionality based on current patterns in the automated test suite, and am open to suggestions.

### Author/Contributor

### Reviewing Maintainer


tpansino:
@admiralAwkbar @ferrarimarco @GaboFDC @Hanse00 @jwiebalk @nemchik @zkoppert, could one of you re-run the build on this PR please? 🙏 It's failed 3 different times on items that can't possibly be related to my changes. Seems like it's just build/test flakiness?

Also, I'd appreciate some feedback if you can make time, either on the change itself or how we might add some tests for it.

admiralAwkbar (approved):

PR #1356, opened 2021-03-11
Update deploy-PROD.yml

admiralAwkbar:
Fix Release tags

### DIFF
--- a/.github/workflows/deploy-PROD.yml
+++ b/.github/workflows/deploy-PROD.yml
@@ -96,9 +96,7 @@ jobs:
           push: true
           tags: |
             github/super-linter:latest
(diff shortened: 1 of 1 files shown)


ferrarimarco (approved):

ferrarimarco:
So only a `3.whatever` release is able to update the `v3` image tag, right? (at least it's my understanding after reading https://github.com/github/super-linter/blob/master/.github/workflows/deploy-RELEASE.yml)

admiralAwkbar:
@ferrarimarco YOu are correct, this seems to have been an oversight
Also I'm fixing the market place issue right now as well...
Seems we need to accept some terms...

PR #1466, opened 2021-04-22
Implement a test suite with InSpec

ferrarimarco:
## Proposed Changes

This PR implements an [InSpec](https://community.chef.io/tools/chef-inspec/) test suite that validates the container image. Testing super-linter code is out of the scope of this PR.

Other PRs, such as #1461 will benefit from this one because it's a baseline to build on.

List of notable changes:

- [ ] Implement an InSpec test suite to validate the image.
    - [ ] Installed APK packages
    - [ ] Installed gems
    - [ ] Installed pip packages
    - [ ] Installed npm packages
    - [ ] Installed phive packages
    - [x] Availability of linter commands
    - [x] Ensure linter version commands return successfully
- [x] Validate the InSpec test suite (see [inspec check](https://docs.chef.io/inspec/cli/#check)).
- [x] Run the test suite in our CI workflows.
- [ ] Remove the runtime validation steps that are covered by the test suite for quicker super-linter executions:
    - [x] Availability of linters and commands at runtime
- [ ] (TO EVALUATE) Dynamically load information in InSpec controls
    - [ ] APK packages
    - [ ] Gems
    - [ ] pip packages
    - [ ] npm packages
    - [ ] phive packages
    - [ ] Linter commands
    - [ ] Linter version commands
    - [ ] Linter names 

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar:
@ferrarimarco made ruby happy for you :)

test/inspec/super-linter/controls/super_linter.rb
```diff
@@ -0,0 +1,150 @@
...
+
+###########################################
+# Check to see all binaries are installed #
+###########################################
+control "super-linter-installed-binaries" do
+  impact 1
+  title "Super-Linter installed binaries check"
+  desc "Check that binaries that Super-Linter needs are installed."
+
+  packages = [
```
ferrarimarco, line 85: Maybe rename this to `commands`? or `binaries`?

test/inspec/super-linter/controls/super_linter.rb
```diff
@@ -0,0 +1,150 @@
...
+    "tekton-lint",
+    "terragrunt",
+    "terrascan",
+    "tflint",
+    "xmllint",
+    "yamllint"
+  ]
+
+  packages.each do |item|
+    describe command("command -v #{item}") do
```
ferrarimarco, line 146: Shall we test the for the path as well? Are we expecting the binaries to be in a particular directory?

ferrarimarco (commented):
@admiralAwkbar nice additions!

While we implement these checks with InSpec, let's consider removing the checks that we do in shell scripts that become obsolete, such as the [linter command availability](https://github.com/github/super-linter/blob/24e1042fe8e69a84b20768dd50ccb581fc4a8ce8/lib/linter.sh#L326-L340).

test/inspec/super-linter/controls/super_linter.rb
admiralAwkbar: yes commands...

test/inspec/super-linter/controls/super_linter.rb
admiralAwkbar: @ferrarimarco the paths are all over in /usr/lib, /usr/bin/ ... so path is a nightmare, but the `command -v` makes sure it can be found in the path, so I think were good

admiralAwkbar:
@ferrarimarco i removed the code block from `linter.sh` that does the check for `command -v` as we test in the Inspec now :)

test/inspec/super-linter/controls/super_linter_versions.rb
```diff
@@ -0,0 +1,125 @@
...
+##########################################################
+# Being able to run the command `linter --version` helps #
+# achieve that the linter is installed, pathed, and      #
+# has the libraries needed to be able to basically run   #
+##########################################################
+
+#################################################################
+# Check to see that version command works on installed commands #
+#################################################################
+control "super-linter-version-commands" do
```
ferrarimarco, line 12: @admiralAwkbar question: why don't we use a list of {linter, version_command, expected_return_code}? Then we can just loop over that.

test/inspec/super-linter/controls/super_linter_versions.rb
```diff
@@ -0,0 +1,125 @@
...
+    "standard",
+    "stylelint",
+    "tekton-lint",
+    "terragrunt",
+    "tflint",
+    "xmllint",
+    "yamllint"
+  ]
+
+  linters.each do |item|
```
ferrarimarco, line 68: @admiralAwkbar 

After having a look at this, I suggest putting this control in `super-linter.rb`, so we don't have to maintain two lists of commands.

ferrarimarco (commented):
Nice additions @admiralAwkbar 

I'm wondering how can we automate populating the packages and commands lists, so that we don't have to maintain the same info in different places.

Packages are listed in:

- Dockerfile
- Gemfile
- Pipfile
- package.json
- phive.xml

With the exception of the Dockerfile, those are all easily parsable (`ini`, `json`, and `xml`), so that's doable in Ruby. We might move the `apk` package list somewhere out of the Dockerfile, and load it from the Dockerfile and in the InSpec control as needed.

Additionally, we already defined the linters and the linter commands in super-linter, so we should try to use those lists as well, without redefining it. A viable strategy for this might be:

1. Extract the definition of those variables in dedicated shell files.
2. Source those shell files as needed as part of super-linter or before running the InSpec test suite, so that those variables are available in the environment.
3. Load data from those environment variable (easily doable in Ruby as well

test/inspec/super-linter/controls/super_linter_versions.rb
admiralAwkbar: yes this would simplify the process!

test/inspec/super-linter/controls/super_linter_versions.rb
admiralAwkbar: sounds good to me

admiralAwkbar:
@ferrarimarco i agree we shoudl streamline the tests! Im wondering in this PR if we should just get it to POC level, and then we can iterate in further PR's to streamline and make more compact :)

ferrarimarco:
> @ferrarimarco i agree we shoudl streamline the tests! Im wondering in this PR if we should just get it to POC level, and then we can iterate in further PR's to streamline and make more compact :)

Fair enough. I'll move the streamlining tasks to a separate PR

test/inspec/super-linter/controls/super_linter.rb
```diff
@@ -0,0 +1,267 @@
...
+  ]
+
+  commands.each do |item|
+    describe command("command -v #{item}") do
+      its("exit_status") { should eq 0 }
+    end
+  end
+end
+
+# frozen_string_literal: true
```
ferrarimarco, line 143: Remove this

admiralAwkbar:
@ferrarimarco no idea what this error means :)

```bash
Profile Summary: 3 successful controls, 0 control failures, 0 controls skipped
Test Summary: 157 successful, 0 failures, 0 skipped
Error response from daemon: page not found
make: *** [inspec] Error 1
```

ferrarimarco:
I'm having a look :)

ferrarimarco:
> @ferrarimarco no idea what this error means :)
> 
> ```shell
> Profile Summary: 3 successful controls, 0 control failures, 0 controls skipped
> Test Summary: 157 successful, 0 failures, 0 skipped
> Error response from daemon: page not found
> make: *** [inspec] Error 1
> ```

I think this is the docker daemon that tries to kill a container that's not there.

ferrarimarco:
@admiralAwkbar the build is now failing because of the TAPs not matching with the expected output. We're going to remove those, so not a big concern IMHO.

I feel this is ready for review!

admiralAwkbar:
@ferrarimarco amazing work on this! Well clean up the code base and remove tap after this is up

jwiebalk (approved):

Issue #1482, opened 2021-04-26
`git: 'remote-https' is not a git command` error with github/super-linter@v3 (v3.16.0)


jphuynh:
## Describe the bug

The [latest release v3.16.0](https://github.com/github/super-linter/releases/tag/v3.16.0) seems to error our GH actions that uses the `v3` tag. 

Pinning to version `v3.15.5` fixed the issue.

## Steps to Reproduce

Use the v3 of the github action.

## Logs

```
2021-04-26 16:20:53 [INFO]   --------------------------------------------
2021-04-26 16:20:53 [INFO]   Gathering user validation information...
2021-04-26 16:20:53 [INFO]   - Only validating [new], or [edited] files in code base...
git: 'remote-https' is not a git command. See 'git --help'.
```

ohashih:
@jphuynh
Thanks !
I was trying to report an issue in the same situation.

## Logs
```
git: 'remote-https' is not a git command. See 'git --help'.
2021-04-26 16:51:16 [INFO]   Failed to switch to main branch to get files changed!
2021-04-26 16:51:16 [FATAL]   [error: pathspec 'main' did not match any file(s) known to git]
```

PR #1483, opened 2021-04-26
fix: Install git in final docker image

jphuynh:
This is missing since the v3.16.0 release when refactoring the Dockerfile.

Fixes: https://github.com/github/super-linter/issues/1482

Fixes #

https://github.com/github/super-linter/issues/1482

## Proposed Changes

1. Install `git` and `git-lfs` in final docker image

### Author/Contributor

### Reviewing Maintainer


jphuynh:
Note: I couldn't test it easily but in my opinion the error seemed to indicate that and a quick check with the v3.15.5 showed that git was available on that image.

admiralAwkbar (approved):

admiralAwkbar:
@jphuynh good catch.. this would cause issues...

PR #1496, opened 2021-04-30
remove tap and tests

admiralAwkbar:
- Removing tap tests as we have moved to Inspec...


jwiebalk (approved):

zachpetersen:
@admiralAwkbar would this be why we no longer get TAP reports from output? These had been used previously to publish test results in Azure DevOps pipelines.

admiralAwkbar:
@zachpetersen Hey, 
We had moved away from Tap reports and checks on our side as it was constantly breaking the builds.
We may need to look to re-implement them for users, but they were so intertwined, they had to be pulled.

Sorry for any inconvenience, and if you want, you can start a PR to bring back that capability for Azure

zachpetersen:
@admiralAwkbar thank you for the reply.  Too bad about the difficulties it was causing. I think TAP was the only output format available other than to the console. Is there anything similar that is possible with the existing code base?

I had previously [referenced this article](https://adinermie.com/publishing-github-super-linter-terraform-quality-checks-to-azure-devops-pipelines/) to configure TAP to JUnit conversion which could then be published to an Azure DevOps Pipeline using the [Publish Test Results task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results).

I'll see if there is any reasonable chance I could successfully reimplement the TAP output, however, I am not sure thats in my wheelhouse.

@AErmie fyi

PR #1497, opened 2021-04-30
Inspec additional Tests

admiralAwkbar:
- Adding Inspec tests to the test suite


ferrarimarco:
@admiralAwkbar I'd try with the [InSpec shell](https://docs.chef.io/inspec/cli/#shell), and run the same check there and see if it keeps failing

ferrarimarco:
Plus, there are more leftovers to remove :)

See the [TEST_CASE_RUN variable](https://github.com/github/super-linter/search?q=TEST_CASE_RUN), and where it's called (including [a workflow step](https://github.com/github/super-linter/blob/master/.github/workflows/deploy-DEV.yml#L134))

admiralAwkbar:
added #1509 to track the issue with those npm packages

Dockerfile
```diff
@@ -214,7 +214,7 @@ COPY --from=dockerfile-lint /bin/hadolint /usr/bin/hadolint
 #################
 # Install lintr #
 #################
-COPY --from=lintr-lib /usr/lib/R/library/ /home/r-library
+COPY --from=lintr-lib /usr/lib/R/library/ /usr/lib/r-library
```
ferrarimarco, line 217: I'd add a test for this (and similar directives). See https://docs.chef.io/inspec/resources/file/

ferrarimarco (changes requested):

PR #1554, opened 2021-05-12
Allow ansible-lint with git diffs

DanEmord:
super-linter checks the path of the file in lib/functions/detectFiles.sh to determine whether the file is an ansible playbook. 

When VALIDATE_ALL_CODEBASE=true, a list of absolute paths is generated which matches the expected path.
When VALIDATE_ALL_CODEBASE=false, a list of relative paths is generated so no files are "detected" as ansible playbooks. 

This change outputs everything as an absolute path so diffs will also work.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/functions/buildFileList.sh
+++ b/lib/functions/buildFileList.sh
@@ -100,7 +100,7 @@ function BuildFileList() {
       ################
       # push event   #
       ################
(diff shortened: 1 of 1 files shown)


admiralAwkbar:
@DanEmord Very cool catch and simple fix

Will let automation take the wheel :)

admiralAwkbar:
@DanEmord

admiralAwkbar (approved):
Dope code

DanEmord:
Sweet, thanks for the help @admiralAwkbar :-)

PR #1562, opened 2021-05-19
fix gcc LTO

M1cha:


admiralAwkbar (approved):
Nice catch!

weikengchen:
Thanks a lot for this fix!

PR #1567, opened 2021-05-21
Create linter.yaml

DamonDoesStuff:


zkoppert:
https://github.com/github/super-linter/blob/master/.github/workflows/stack-linter.yml

PR #1595, opened 2021-05-25
adding the old way

admiralAwkbar:
- use the action vs the script
- Since the script wasn't working for the actions sections, back to simplicity


admiralAwkbar:
@ferrarimarco  @GaboFDC It looks like this came back to bite us :)

https://github.com/github/super-linter/pull/1595/checks?check_run_id=2669011203#step:8:1383

Any ideas on the best way to solve this so we don't break the logic of getting the diff?

admiralAwkbar:
guessing it was a fluke.... or out-of-date weird...

PR #1644, opened 2021-06-09
adding rules file

### LINKED ISSUE #1643: Support passing a custom .ini file to the PHP_BUILTIN linter (so I can enable short_open_tag)
I am trying to run the linter on files that use short open tags and it throws errors for valid files.

I would like to be able to override ini directives using a custom ini file

**Additional context**
I believe the ini file could be passed in on the command line by reusing the same approach you use for custom rules. e.g.
```
LINTER_COMMANDS_ARRAY['PHP_BUILTIN']="php -l -c ${PHP_PHP_BUILTIN_INI}"
```

admiralAwkbar:
this closes #1643 

This will add the new var:
- `PHP_CONFIG_FILE`
And we have a default template: `php.ini`

Also updated command to take in config file


admiralAwkbar:
@ferrarimarco So this PR is about adding PHP config, but we ran into space issues on the test. I updated the makefile to be a bit smarter :) that way we don't build the image again and take up the 14GB disk space on the runner :)

jwiebalk (approved):

PR #1647, opened 2021-06-10
Change default LINTER_RULES_PATH to / (#1634).

### LINKED ISSUE #1634: Change Default of LINTER_RULES_PATH to Root of Project
Super-Linter is a great tool for enforcing lint rules in CI, but running linters locally (a) is typically faster, (b) can be done before committing to keep each individual commit clean (and not just the entire pull request), (c) doesn't require an Internet connection, and (d) doesn't consume build minutes. Presently, if you have already configured linters to run locally, Super-Linter largely ignores these configuration files, which can be easily overlooked since it doesn't necessarily result in an error.

Default `LINTER_RULES_PATH` to "." rather than ".github/linters." Nearly all configurable linters look for their configuration files in the root of the project by default, so this would make Super-Linter play better out of the box with most of these linters.

At present, the alternative is that every project that desires to both use Super-Linter and run their linters locally needs to independently make this same discovery and then set `LINTER_RULES_PATH` to `.` themselves.

Kurt-von-Laven:
# Fixes #1634

Replaces the default `LINTER_RULES_PATH` of `.github/linters` with the root of the downstream repository.

## Proposed Changes

1. Match the location essentially all linters check for their rules file by default. The previous default, .github/linters, had the drawback of requiring the extra step of setting `LINTER_RULES_PATH` to `/` for those wishing to also run their linters locally.
2. Improve wording of `LINTER_RULES_PATH` docs.

### Author/Contributor

### Reviewing Maintainer


Kurt-von-Laven:
No worries; I understand. It was a very fast change to demonstrate in any case. Do you have any interest in the first commit alone? Thanks for #1649!

Issue #1566, opened 2021-05-20
Markdown Lint Ignore Files or Folders


tjperry07:
It's not clear how to ignore certain files when using Markdownlint. I have some markdown files that don't need to be checked and they return a false positive every time.

tjperry07:
PEBKAC. Forgot about adding`<!-- markdownlint-disable -->`  to files I wanted to ignore.

Issue #1456, opened 2021-04-16
Can superlinter be locally set to fix what it finds if possible when run on the command line


hepcat72:
E.g.: black & isort.

I understand if it may be problematic in an editor to have superlinter's run of black & isort have side-effects, but if run on the command line, is there like an environment variable that can be set to have it do what it can with linters that offer auto-fixes?

zkoppert:
👋🏻 Hi @hepcat72, personally I push my commits up to GitHub and use GitHub Actions to run the super linter but I see where you are coming from about running the linters sooner and wanting the auto fix capabilities where available. 

Take a look at how to run the super-linter locally [here](https://github.com/github/super-linter/blob/master/docs/run-linter-locally.md) to be able to see if there are any linter issues locally. Then in terms of autofix, I normally download those tools with auto fix myself but you might be able to modify the docker container to do this.

Hope that helps!

nvuillam:
You can also use github super-linter's little brother Mega-linter, who can apply fixes locally or even from CI with a new commit :)

Issue #1102, opened 2020-12-22
Support arguments for Kubeval


sheeeng:
**Is your feature request related to a problem? Please describe.**

As a user, I want super-linter to be able to support arguments for kubeval.

```bash
kubeval --ignore-missing-schemas fixtures/test_crd.yaml
```

See [CRDs](https://www.kubeval.com/#crds)

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Note**
If this is a request to add a new language, after submitting this issue check out [the wiki](https://github.com/github/super-linter/wiki/Adding-new-language-support) for more info on how to accomplish that.

ferrarimarco:
Duplicate #891

ferrarimarco:
Actually, not a duplicate. You can now use `KUBERNETES_KUBEVAL_OPTIONS`. Although, there's #891 that hurts a bit, still.

PR #1667, opened 2021-06-15
Support linter command initialization

### LINKED ISSUE #1852: Allow to specify command line argument of actionlint

[actionlint](https://github.com/rhysd/actionlint) provides `-ignore` option to ignore some errors. This is necessary to ignore some false positives or errors which a user is not interested in. However, super-linter currently does not provide a way to pass command line options to actionlint.

Related issue: https://github.com/rhysd/actionlint/issues/26

I'd like to have the way to ignore some errors from actionlint. It'd be nice to have some environment variable for it.

N/A

**Additional context**

N/A

ferrarimarco:
## Proposed Changes

There have been multiple requests to customize the linter invocation commands (#1643, #1566, #1456, #1102, #1853). This PR aims to support that feature. Changes:

1. `linterCommands.sh`: holds the functions to manage the linter commands. (+ test to check for its presence)
    1. `GetLinterOptions` function: checks if a user provided a customized value (`<XXXX>_LINTER_COMMAND_OPTIONS`) to configure the command line options of the `<XXXX>` linter.
2. Move default command options to `LINTER_DEFAULT_OPTIONS_ARRAY` 
3. ~Ensure backward compatibility with the linters for which we already allow some customization~ Exit with error if a user is using deprecated variables:
    - [x] Kubeval (`KUBERNETES_KUBEVAL_OPTIONS`)
    - [x] Markdownlint (`MARKDOWN_CUSTOM_RULE_GLOBS`)
4. Support the customization of the command line options for every linter.
    - [ ] Fix the issue with the customization of the ARM linter because PowerShell variables and subshells are interpreted by the linux shell, instead of being literals, regardless of the escaped `$`. 
5. (STILL TO DECIDE) Deprecate the linter configuration variables because now users can fully customize the command line options, so no need to duplicate that behavior.
6. Update the README:
     - [x] New variables
     - [ ] (STILL TO DECIDE) Deprecated variables (linter configuration variables)
     - [x] Deprecated variables (`KUBERNETES_KUBEVAL_OPTIONS`, `MARKDOWN_CUSTOM_RULE_GLOBS`)
     - [x] Add a missing newline when listing the Docker images.

### Author/Contributor

### Reviewing Maintainer

Closes #1852


ferrarimarco:
@admiralAwkbar at the moment, this PR supports customizing the invocation of linters for which we already provide some default options, besides:

- The location of the configuration files.
- Enabling a "check/lint only" mode that doesn't modify files

Those linters are, at the moment: ansiblelint, shellcheck, dotnet-format, dartanalyzer, npm-groovy-lint, kubeval, chktex, phpstan, rubocop, snakefmt, terrascan, terragrunt.

There are a couple of open questions:

1. Do we want to allow disabling the "check/lint only" mode for linters that support it? This would lead to a more flexible experience for users, but would then allow those linters to actually modify the source code they lint.
2. Do we want to support customizing the invocation of every linter that's currently supported? Ideally yes, but then we need to have a look at each linter, to check if it supports options in the first place. For the moment, I took the approach of supporting the customization of the linters for which we set default options (see above).

admiralAwkbar:
@ferrarimarco this is an interesting problem... as it leads to issues with the way we had originally structured the commands.

For most linters, we feed the command to the configuration file, and some other very basic commands. If we remove those and allow the users to pick and choose, that gives them greater flexibility, but also breaks our original strategy, and makes basic configuration obsolete, and makes this tool more complex for the users.

I would assume the right path would be, if the user gives commands for a specific linter, we have to remove our base commands for that linter, which would likely include our path to config file, and small configs.

That would seem to lead users down the best path as all previous functionality would work, and they would only have to point to the configuration file and base commands in their extra commands...

Im going to say documentation on this should be pretty good to prevent many issues...

ferrarimarco:
So, following up to our chat, here's what we have on the plate. I think there are two questions:

1. How much do we want to allow for the customization of the invocation of each linter?
2. Should we allow the customization of all the linters?

These are the alternatives I see, for any given linter:

- Drop all the options that super-linter would have set: If a user customizes the options for a linter, we drop all the options that super-linter would have set. Then we use what the user provided, as is.
    - Pros: (likely) simpler to implement. Least surprising for users (all or nothing). Might lead to non-backward compatible simplification to the code base (example: we don't care about finding the configuration files anymore, because users can set up their own).
    - Cons: users lose the guarantee that super-linter correctly picked the right configuration file, and that enabled a "check/lint only" mode where possible.
- Keep the configuration options: If a user customizes the options for a linter, we keep the directives to find the configuration files, and we drop all the other options that super-linter would have set. Then we add what to user provided, as is.
    - Pros: keep the work that super-linter does to validate the location and the usage of configuration files.
    - Cons: may surprise some default users because some options are kept at runtime, but not all.
- Keep the configuration + "check/lint only" options: Like the previous one, but we also keep the "check/lint only" option as well.
    - Pros and cons same as above, plus the fact that we guarantee the "check/lint only" mode.

After thinking about this, I suppose the first option makes the most sense from the complexity point of view, and from being the least surprising for users.

My suggestion then is:

1. Implement the first option above.
2. Remove the variables we intended to deprecate in this PR.
3. Remove support for specifying the configuration files with environment variables because users can use the customization of the linter invocation instead. This makes those variables (and the related logic) obsolate.

Thoughts?

@admiralAwkbar

admiralAwkbar:
@ferrarimarco I agree with option 1. I think it's the sanest and likely easiest for the user to follow.

We need to make it evident to the user that if they set additional configuration for a linter, they will also likely need to set the config path as well. We can trey to do some simple path test to catch some users who make path issues, but that should be about right

ferrarimarco:
Sounds like a plan!

ferrarimarco:
@admiralAwkbar I've implemented the customization for all linters. There are some open points:

1. Deprecation of the linter configuration file variables.
2. Update README.
3. Fix the issue with ARM customization (see the details in the PR description above).

What shall we do about point 1?

admiralAwkbar:
@ferrarimarco We could go to version 5.x and that would leave those users time to update...
I do think giving them the ability to point to the file using a custom string with other rules does lend to greater flexability.

ferrarimarco:
> @ferrarimarco We could go to version 5.x and that would leave those users time to update...
OK

> I do think giving them the ability to point to the file using a custom string with other rules does lend to greater flexability.

I think I don't get this comment LOL :)

ferrarimarco:
> I do think giving them the ability to point to the file using a custom string with other rules does lend to greater flexibility.

@admiralAwkbar what do you mean here? :)

ferrarimarco:
@admiralAwkbar ping regarding the above question :)

admiralAwkbar:
@ferrarimarco Hey! sorry, I have been slammed!
But I think this is a great idea, and we need to try to go forward. Hopefully, in the next weeks, I get some time to work towards the super-linter and less maint work...

admiralAwkbar:
@ferrarimarco Giving you a hand... had a few min :)

rhysd:
Can you add `Closes #1852` to description of this PR? It will close the issue when this PR is merged.

admiralAwkbar:
@ferrarimarco Do you think this is a `5.y.z` release or a bump in `4.y.z`?

ferrarimarco:
I'd bump the major version.

Also, there's an issue with the customization of the ARM command line options because of escapes not working for the `$` sign, so the shell gets confused and picks up variables that are for Powershell.

admiralAwkbar:
@ferrarimarco Thank you for all the help and great info on this!

I think were down to the `arm` issue and then were done... Ran out of time to work on this for the week. if you have time it would be great!

admiralAwkbar:
@ferrarimarco I tried to test and validate the arm but I'm not able to make it behave the same as when run...

```bash
#!/bin/bash -x


ARM_TTK_PSD1="/usr/lib/microsoft/arm-ttk-master/arm-ttk/arm-ttk.psd1"
ARM_LINTER_RULES="/action/lib/.automation/.arm-ttk.psd1"

LANGUAGE='ARM'
LINTER_NAME='ARM-TTK'
FILTER_REGEX_INCLUDE=''
FILTER_REGEX_EXCLUDE=''
TEST_CASE_RUN='false'
WORKSPACE_PATH='/test'
FILE='arm-good.json'

ARM_LINTER_COMMAND_OPTIONS="\${config} = \$(Import-PowerShellDataFile -Path ${ARM_LINTER_RULES}) ; Test-AzTemplate @config -TemplatePath"
LINTER_COMMAND="Import-Module ${ARM_TTK_PSD1} ; ${ARM_LINTER_COMMAND_OPTIONS}"

################################################################################
################################################################################
function LintCodebase()
{
  FILE_TYPE="${1}" && shift            # Pull the variable and remove from array path  (Example: JSON)
  LINTER_NAME="${1}" && shift          # Pull the variable and remove from array path  (Example: jsonlint)
  LINTER_COMMAND="${1}" && shift       # Pull the variable and remove from array path  (Example: jsonlint -c ConfigFile /path/to/file)
  FILTER_REGEX_INCLUDE="${1}" && shift # Pull the variable and remove from array path  (Example: */src/*,*/test/*)
  FILTER_REGEX_EXCLUDE="${1}" && shift # Pull the variable and remove from array path  (Example: */examples/*,*/test/*.test)
  TEST_CASE_RUN="${1}" && shift        # Flag for if running in test cases
  
  LINT_CMD=$(
    cd "${WORKSPACE_PATH}" || exit
    pwsh -NoProfile -NoLogo -Command "${LINTER_COMMAND} \"${FILE}\"; if (\${Error}.Count) { exit 1 }"
    exit $? 2>&1
  )
}
################################################################################
################################################################################


LintCodebase "${LANGUAGE}" "${LINTER_NAME}" "${LINTER_COMMAND}" "${FILTER_REGEX_INCLUDE}" "${FILTER_REGEX_EXCLUDE}" "${TEST_CASE_RUN}"

```

When ran I get the output:
```bash
+ shift
++ cd /test

++ pwsh -NoProfile -NoLogo -Command 'Import-Module /usr/lib/microsoft/arm-ttk-master/arm-ttk/arm-ttk.psd1 ; ${config} = $(Import-PowerShellDataFile -Path /action/lib/.automation/.arm-ttk.psd1) ; Test-AzTemplate @config -TemplatePath
"arm-good.json"; if (${Error}.Count) { exit 1 }'
++ exit 0
+ LINT_CMD='
```

But when run from the test suite I get the bad parsing... Any ideas?

rhysd:
I'm not sure, but the option string seems [evaled](https://github.com/github/super-linter/pull/1667/files#diff-0b89e6e14353371aacf69d4b9067d4481dd99c6994409b32e3b906c69d478c21R898) and unescaped twice.

Maybe we need to escape `$` twice? (`\\\$` -> `\$` -> `$`)

```sh
LINTER_DEFAULT_OPTIONS_ARRAY['ARM']="\\\${config} = \\\$(Import-PowerShellDataFile -Path ${ARM_LINTER_RULES}) ; Test-AzTemplate @config -TemplatePath"
```

rhysd:
@adamahata Hey, I have debugged this branch and found we need more backslashes because we use `eval` several times. Can you try the following patch?

```diff
diff --git a/lib/linter.sh b/lib/linter.sh
index 44a7a0d..174bde9 100755
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -812,7 +812,7 @@ GetStandardRules "typescript"
 declare -A LINTER_DEFAULT_OPTIONS_ARRAY
 LINTER_DEFAULT_OPTIONS_ARRAY['ANSIBLE']="-v -c ${ANSIBLE_LINTER_RULES}"
 # shellcheck disable=SC2016,SC2154  # Don't expand variables that PowerShell will handle
-LINTER_DEFAULT_OPTIONS_ARRAY['ARM']="\\${config} = \\\$(Import-PowerShellDataFile -Path ${ARM_LINTER_RULES}) ; Test-AzTemplate @config -TemplatePath"
+LINTER_DEFAULT_OPTIONS_ARRAY['ARM']="\\\\\\\${config} = \\\\\\\$(Import-PowerShellDataFile -Path ${ARM_LINTER_RULES}) ; Test-AzTemplate @config -TemplatePath"
 LINTER_DEFAULT_OPTIONS_ARRAY['BASH']="--color --external-sources"
 LINTER_DEFAULT_OPTIONS_ARRAY['CLANG_FORMAT']="--Werror --dry-run"
 LINTER_DEFAULT_OPTIONS_ARRAY['CLOJURE']="--config ${CLOJURE_LINTER_RULES} --lint"
```

Since `eval` is also used in `GetLinterOptions`, unescaping backslashes is done repeatedly.

I confirmed workflows passed in my fork's CI:

- slim: https://github.com/rhysd/super-linter/runs/3383801327?check_suite_focus=true
- standard: https://github.com/rhysd/super-linter/runs/3383801324?check_suite_focus=true

Mum2541:
gh pr checkout 1667

ferrarimarco:
@admiralAwkbar resuming work on this one, finally! :)

ferrarimarco:
After looking at this a bit more, I realized there are a lot of corner cases to cover. Not worth it. I advise adopting an ad-hoc approach where needed.

egberts:
wut?  No shellcheck CLI option and passing the much vaunted need of ‘-x‘ option (which is ability to read other include bash script files via ‘source‘ command?

ferrarimarco:
@egberts I believe this is tracked in https://github.com/github/super-linter/issues/1633

PR #1689, opened 2021-06-23
Ignore files marked with @generated marker

stepancheg:
`@generated` marker is used by certain tools to understand that the
file is generated, so it should be treated differently than a file
written by a human:
* these files do not need to be reformatted,
* diffs in these files are less important,
* and linters should not be invoked on these files.

This PR proposes builtin support for `@generated` marker (and
`@not-generated` marker to mark file as not generated when it
contains `@generated` marker, like `README.md`).

I have not found a standard for a generated file marker, but:
* Facebook [uses `@generated` marker](https://tinyurl.com/fb-generated)
* Phabricator tool which was spawned from Facebook internal tool
  [also understands `@generated` marker](https://git.io/JnVHa)
* Cargo inserts `@generated` marker into [generated Cargo.lock files](https://git.io/JnVHP)

Super-linter supports regex includes and excludes, but they are
harder to maintain (each repository needs to be configured) than
patching the tools which generate the files.

My personal story is that I maintain rust-protobuf crate, which
started emitting `@generated` markers [six years ago](https://git.io/JnV5h)
after a request of a Phabricator user.

## Test Plan:

Create a test file `test.sh`:

```
echo $a
```

Run:

```
docker run -e RUN_LOCAL=true -v $HOME/tmp/g:/tmp/lint super-linter-test
```

Result is:

```
In /tmp/lint/test.sh line 1:
echo $a
^-- SC2148: Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.
     ^-- SC2154: a is referenced but not assigned.
     ^-- SC2086: Double quote to prevent globbing and word splitting.
...
2021-06-22 23:46:16 [ERROR]   ERRORS FOUND in BASH:[1]
```

Now add `@generated` to the file and run again:

```
2021-06-22 23:47:13 [NOTICE]   All file(s) linted successfully with no errors detected
```

Additionally, add `@not-generated` in addition to `@generated`, and
linter error pops up again.

### Author/Contributor

### Reviewing Maintainer


stepancheg:
Perhaps this need to be labeled as "enhancement" but GitHub doesn't allow me to do that.

ferrarimarco:
@stepancheg thanks for this PR!

This looks interesting. I added a few comments

README.md
```diff
@@ -399,6 +399,10 @@ Examples:
 - Do not lint files inside test folder: `FILTER_REGEX_EXCLUDE: .*test/.*`
 - Do not lint javascript files inside test folder: `FILTER_REGEX_EXCLUDE: .*test/.*.js`
 
+<!-- @not-generated -->
+Additionally super-linter ignores any file with `@generated` marker in it
+unless the file also has `@not-generated` marker.
+
```
ferrarimarco, line 405: Could be useful to:

1. Add an example.
2. Add a link in the README to the [FB blog post](https://engineering.fb.com/2015/08/20/open-source/writing-code-that-writes-code-with-hack-codegen/) you mentioned to get some context.

lib/functions/buildFileList.sh
```diff
@@ -299,6 +299,15 @@ function BuildFileList() {
       continue
     fi
 
+
+    #################################################
+    # Filter files with at-generated marker         #
+    #################################################
+    if IsGenerated "$FILE"; then
```
ferrarimarco, line 306: I think we should make this check configurable with an environment variable that defaults to something that doesn't enable this check, so that we don't break existing functionality.

ferrarimarco (changes requested):

lib/functions/buildFileList.sh
stepancheg: I'll add an environment variable, but maybe use that environment variable to opt-out rather than opt-in?

super-linters breaks things all the time (when it adds more linters which are on by default), so another possible incompatibility is IMHO OK.

Another option is to start with out-in and eventually switch it to opt-out.

stepancheg:
Amended the PR:
* use `IGNORE_GENERATED_FILES` variable, `false` by default
* more documentation

README.md
```diff
@@ -282,6 +282,7 @@ But if you wish to select or exclude specific linters, we give you full control
 | **FILTER_REGEX_INCLUDE**           | `all`                           | Regular expression defining which files will be processed by linters (ex: `.*src/.*`)                                                                                                                                |
 | **GITHUB_DOMAIN**                  | `github.com`                    | Specify a custom Github domain in case Github Enterprise is used: e.g. `github.myenterprise.com`                                                                                                                     |
 | **GITHUB_CUSTOM_API_URL**          | `api.github.com`                | Specify a custom Github API URL in case Github Enterprise is used: e.g. `https://github.myenterprise.com/api/v3/`                                                                                                    |
+| **IGNORE_GENERATED_FILES**         | `true`                          | If set to `true`, super-linter will ignore all the files with `@generated` marker but without `@not-generated` marker.                                                                                               |
```
ferrarimarco, line 285: I think this should state `false` :)

ferrarimarco (changes requested):

ferrarimarco:
Nice work!

README.md
stepancheg: Always had trouble with double negations. Thanks!

stepancheg:
Updated the readme.

stepancheg:
I have created a little website to promote `@generated` marker: https://generated.at/

ferrarimarco:
@admiralAwkbar want to have a look as well?

admiralAwkbar (commented):
This looks really good!
I need to get a fix in for the phar that's breaking all current builds... Shoudl have that done today and we can merge master into this branch and see where it goes!

Thanks for the cool work!

admiralAwkbar (approved):
looks good

admiralAwkbar:
@ferrarimarco all good to you?

admiralAwkbar:
So the fail was from `Raku` as when you run the version command it returns non ascii chars... its super dumb... So I had to update the command to pull those out and stop making the `makefile` angry...

PR #1753, opened 2021-07-13
adding tflint config

### LINKED ISSUE #1706: Use our own terraform config file .tflint.hcl
While using super-linter, i want to replace default terraform config file .tflint.hcl. How to do that? in the terraform directory i have kept config file .tflint.hcl, but it is not considering that

admiralAwkbar:
This closes #1706 

- Adding the ability for the user to supply `.tflint.hcl` instead of ours only :)


admiralAwkbar:
@prithakundu It seems we had it set to only take our file in and not allow users to pull in their own.

Once this is merged and released, you should be able to pass the env var to supply your own file, or put the file in the `.github/linters/.tflint.hcl` location and have it used at run time.

admiralAwkbar:
Example: 

```yml
      # Run Linter against code base #
      - name: Lint Code Base
        uses: github/super-linter@v4
        env:
          VALIDATE_ALL_CODEBASE: false
          DEFAULT_BRANCH: master
          TERRAFORM_CONFIG_FILE: some/dir/.tflint.hcl
```

PR #1762, opened 2021-07-14
adding clang

### LINKED ISSUE #1744: clang-format
If there's `.clang-format`, validate all C/C++ files with it.

I see there are linters like [clang-format-lint](https://github.com/marketplace/actions/clang-format-lint), would be nice if super-linter supported it out of the box.

Note I've created a similar [feature request in cpplint tracker](https://github.com/cpplint/cpplint/issues/180), but I'm not sure that cpplint wants it.

admiralAwkbar:
This closes #1744 

- Adding `Clang` as a linter for C/C++ files
- Need some help with some of the format info


admiralAwkbar:
@stepancheg

I need some help with the `clint-format` base file

We will need a default one to add to the `Templates` folder to default back towards for users, 

This will work for all files with the extensions:
```bash
elif [ "${FILE_TYPE}" == "cpp" ] || [ "${FILE_TYPE}" == "h" ] ||
      [ "${FILE_TYPE}" == "cc" ] || [ "${FILE_TYPE}" == "hpp" ] ||
      [ "${FILE_TYPE}" == "cxx" ] || [ "${FILE_TYPE}" == "cu" ] ||
      [ "${FILE_TYPE}" == "hxx" ] || [ "${FILE_TYPE}" == "c++" ] ||
      [ "${FILE_TYPE}" == "hh" ] || [ "${FILE_TYPE}" == "h++" ] ||
      [ "${FILE_TYPE}" == "cuh" ] || [ "${FILE_TYPE}" == "c" ]; then
      # Append the file to the array #
      FILE_ARRAY_CPP+=("${FILE}")
      FILE_ARRAY_CLANG+=("${FILE}")
```

Currently the command being run is:
```bash
LINTER_COMMANDS_ARRAY['CLANG']="clang-format --Werror --dry-run"
```
But I think when we push this it needs to be something like
```bash
LINTER_COMMANDS_ARRAY['CLANG']="clang-format --Werror --dry-run --style=/path/to/clang_format"
```

Can you help me get that file?

admiralAwkbar:
@stepancheg

Also if you could help with the tests, need a good file that lints successfully, and one that fails :) That way we can add to `.automation/test/clang/`

README.md
```diff
@@ -272,6 +272,7 @@ But if you wish to select or exclude specific linters, we give you full control
 | **ACTIONS_RUNNER_DEBUG**           | `false`                         | Flag to enable additional information about the linter, versions, and additional output.                                                                                                                             |
 | **ANSIBLE_CONFIG_FILE**            | `.ansible-lint.yml`             | Filename for [Ansible-lint configuration](https://ansible-lint.readthedocs.io/en/latest/configuring.html#configuration-file) (ex: `.ansible-lint`, `.ansible-lint.yml`)                                              |
 | **ANSIBLE_DIRECTORY**              | `/ansible`                      | Flag to set the root directory for Ansible file location(s), relative to `DEFAULT_WORKSPACE`. Set to `.` to use the top-level of the `DEFAULT_WORKSPACE`.                                                            |
+| **CLANG_CONFIG_FILE**              | `.clang-format`                 | Filename for [clang-format configuration](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) (ex: `.clang-format`                                                                                             |
```
stepancheg, line 275: `CLANG_FORMAT_CONFIG_FILE` probably.

clang-format shares C++ parser with clang, but it is a separate project within LLVM.

README.md
```diff
@@ -317,6 +318,7 @@ But if you wish to select or exclude specific linters, we give you full control
 | **VALIDATE_BASH**                  | `true`                          | Flag to enable or disable the linting process of the Bash language.                                                                                                                                                  |
 | **VALIDATE_BASH_EXEC**             | `true`                          | Flag to enable or disable the linting process of the Bash language to validate if file is stored as executable.                                                                                                      |
 | **VALIDATE_CPP**                   | `true`                          | Flag to enable or disable the linting process of the C++ language.                                                                                                                                                   |
+| **VALIDATE_CLANG**                 | `true`                          | Flag to enable or disable the linting process of the C++/C language with clang.                                                                                                                                      |
```
stepancheg, line 321: Same here, it should probably be `VALIDATE_CLANG_FORMAT` — this thing only validate formatting, no higher level things.

lib/linter.sh
```diff
@@ -210,6 +215,7 @@ LINTER_NAMES_ARRAY['ANSIBLE']="ansible-lint"
 LINTER_NAMES_ARRAY['ARM']="arm-ttk"
 LINTER_NAMES_ARRAY['BASH']="shellcheck"
 LINTER_NAMES_ARRAY['BASH_EXEC']="bash-exec"
+LINTER_NAMES_ARRAY['CLANG']="clang-format"
```
stepancheg, line 218: I would call it `CLANG_FORMAT`, but up to you of course.

test/inspec/super-linter/controls/super_linter.rb
```diff
@@ -94,6 +95,7 @@
     { linter_name: "asl-validator"},
     { linter_name: "bash-exec", expected_exit_status: 1}, # expect a return code = 1 because this linter doesn't support a "get linter version" command
     { linter_name: "black"},
+    { linter_name: "clang"},
```
stepancheg, line 98: And here.

stepancheg (commented):
Inline comments.

lib/linter.sh
admiralAwkbar: Done deal :)

test/inspec/super-linter/controls/super_linter.rb
admiralAwkbar: good catch

admiralAwkbar:
@stepancheg Thanks for all the collaboration on this! Should be almost down to the testing and were off and running :)

stepancheg:
> Can you help me get that file?

`--style` argument accepts style name, not file name, e. g. `--style=Microsoft` or `-style="{BasedOnStyle: llvm, IndentWidth: 8}"`. Perhaps it's better to avoid this option and rely solely on `.clang-format` in the repository root if provided by user.

And moreover, perhaps do not enable this linter by default if there's no `.clang-format` file.

AFAIU, `.clang-format` file name is hardcoded in `clang-format` binary. `--style=file` simply means "use .clang-format".

admiralAwkbar:
@stepancheg

> > Can you help me get that file?
> 
> `--style` argument accepts style name, not file name, e. g. `--style=Microsoft` or `-style="{BasedOnStyle: llvm, IndentWidth: 8}"`. Perhaps it's better to avoid this option and rely solely on `.clang-format` in the repository root if provided by user.
> 
> And moreover, perhaps do not enable this linter by default if there's no `.clang-format` file.
> 
> AFAIU, `.clang-format` file name is hardcoded in `clang-format` binary. `--style=file` simply means "use .clang-format".

So when the Super linter runs, it usually runs from the root directory and calls the file at its full absolute path. Would it be able to find the `.clang-format` file naturally like this? Or would we need to change into the directory and hope it traverses backward?

I misunderstood the `--style` and will remove those options :)

stepancheg:
As for good files and bad files, example good file with default style:

```
int main() { return 0; }
```

Example bad file:

```
void * * foo() {}
```

which is reformatted with default settings as:

```
void **foo() {}
```

or with with option `--style='{"PointerAlignment": "Left"}'` as:

```
void** foo() {}
```

stepancheg:
> Would it be able to find the .clang-format file naturally like this? 

Yes, `clang-format` looks for `.clang-format` file in all the parent directories of the source file (or current directory for stdin).

stepancheg:
Basically, developers just drop `.clang-format` in the project root, and clang-format "just works".

stepancheg:
One think I need to note. [clang-format-lint](https://github.com/marketplace/actions/clang-format-lint) project allows to specify `clang-format` version — 5 through current 12 — because `clang-format` behavior might change slightly between versions, and some projects might rely on specific version of `clang-format` in their projects.

I think we don't need that much flexibility in super-linter, and we should just always use the latest version. But that's something to keep in mind.

stepancheg:
Perhaps instead of running `--Werror --dry-run` it would be more convenient to run

```
clang-format some_file.c | diff -u some_file.c -
if [ $? -ne 0 ]; then
  echo "file need to be reformatted" >&2
fi
```

Because if a developer does not have a `clang-format` locally (or has a slight different version of clang-format), it would be convenient to know what linter expects from from them.

tojon:
> Basically, developers just drop .clang-format in the project root, and clang-format "just works".

@stepancheg @admiralAwkbar So, I should be able to drop a .clang-format file in my repo root like this:

```
# Use the Google style in this repo
BasedOnStyle: Google
```

..and super-linter will run with that Google-style? How can I know with certainty that it ran with that style^ in the super-linter GitHub Action output logs? I don't seem to see anything in those logs that states the style choice used during the run. Thanks :)

PR #1765, opened 2021-07-14
clang-format: print diff when code need to be reformatted

stepancheg:
Follow-up to #1762.

When ran locally, it prints now:

```
2021-07-14 22:59:56 [INFO]   File:[/tmp/lint/cc.cpp]
2021-07-14 22:59:56 [ERROR]   Found errors in [clang-format] linter!
2021-07-14 22:59:56 [ERROR]   Error code: 1. Command output:
------
--- cc.cpp
+++ cc.cpp.expected
@@ -1,5 +1,5 @@
-#include <iostream>
 #include <fstream>
+#include <iostream>
 using namespace std;

 int main() {
@@ -10,5 +10,5 @@
   MyFile << "Files can be tricky, but it is fun enough!";

   // Close the file
-  //MyFile.close();
+  // MyFile.close();
 }
------
```

Before this PR it printed:

```
2021-07-14 22:58:15 [INFO]   File:[/tmp/lint/cc.cpp]
2021-07-14 22:58:15 [ERROR]   Found errors in [clang-format] linter!
2021-07-14 22:58:15 [ERROR]   Error code: 1. Command output:
------
/tmp/lint/cc.cpp:1:1: error: code should be clang-formatted [-Wclang-format-violations]
#include <iostream>
^
/tmp/lint/cc.cpp:13:5: error: code should be clang-formatted [-Wclang-format-violations]
  //MyFile.close();
    ^
------
```

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar:
@stepancheg Really nice! I want to follow the flow we have with others that require some additional work when run. I hope this code makes more sense and the flow stays the same for the other linters.

stepancheg:
Might be somewhat harder to debug it this way, but you are the boss.

And on a related note, isn't it time to start rewriting super-linter in Python? I found these shell script quite fragile.

admiralAwkbar:
@stepancheg  Hey thanks! I will admit as it scaled out, it did become more so...
We have had other contributors try to rewrite in python, powershell, perl... etc

For the most case, the codebase is in a good spot, and we have tried to invert in more testing and hardening of the codebase, but since this is purely opensource, I don't get to spend the time or effort I would like :)

If anything, this would become Ruby or JS :)

stepancheg:
Last commit doesn't look right:
* `clang-format "${FILE_NAME}";` now seems to spill the reformatted file to output
* `diff` returns 1 on mismatch, not sure why the change was needed

stepancheg:
> If anything, this would become Ruby or JS :)

Ruby would do. As for node, I'm afraid it does not provide enough low-level flexibility to implement linting framework.

Although, perhaps Python is more suitable language to accept contributions from the outside world due to its popularity (although I dislike Python myself — if I did it, I would be Rust).

lib/functions/worker.sh
```diff
@@ -254,7 +254,7 @@ function LintCodebase() {
       elif [[ ${FILE_TYPE} == "CLANG_FORMAT" ]]; then
         LINT_CMD=$(
           LABEL=$(realpath --relative-to . "${FILE_NAME}")
-          if ! clang-format "${FILE_NAME}"; then
+          if ! clang-format --Werror --dry-run "${FILE_NAME}"; then
             diff -u --label "${LABEL}" "${FILE_NAME}" --label "${LABEL}.expected" -
```
stepancheg, line 258: Need to supply `clang-format` to stdin here.

admiralAwkbar: Im not exactly sure what you mean?

stepancheg:
```
clang-format "${FILE_NAME}" | diff -u --label "${LABEL}" "${FILE_NAME}" --label "${LABEL}.expected" -
```

stepancheg: This is still not fixed.

admiralAwkbar:
@stepancheg The error is it keeps saying it cannot find the test

will try to debug more later

MRGRAY0000 (approved):

lib/functions/worker.sh
```diff
@@ -248,6 +248,21 @@ function LintCodebase() {
...
           ${LINTER_COMMAND} "${FILE_NAME}" 2>&1
         )
+      ################################################################
+      # Corner case for CLANG_FORMAT as it needs additional commands #
+      ################################################################
+      elif [[ ${FILE_TYPE} == "CLANG_FORMAT" ]]; then
+        LINT_CMD=$(
+          LABEL=$(realpath --relative-to . "${FILE_NAME}")
+          if
+            ! clang-format --Werror --dry-run "${FILE_NAME}" |
```
ferrarimarco, line 258: This command should come from the linters command array in linter.sh to be consistent with how we configure such comands.

PR #1775, opened 2021-07-16
Adding actionlint

### LINKED ISSUE #1771: Add actionlint
[actionlint](https://github.com/rhysd/actionlint) is a static checker for GitHub Actions workflow files.
If we add it, we can find mistakes in GitHub Action workflow files and shell scripts in the files.

admiralAwkbar:
This closes #1771 

- Adding `actionlint` to lint **GitHub Actions**


Dockerfile
```diff
@@ -134,6 +134,13 @@ RUN pip3 install --no-cache-dir pipenv \
 # Installs Perl dependencies #
 ##############################
 RUN curl --retry 5 --retry-delay 5 -sL https://cpanmin.us/ | perl - -nq --no-wget Perl::Critic \
+#######################
+# Installs ActionLint #
+#######################
+    && curl --retry 5 --retry-delay 5 -sLO https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash \
+    && chmod +x download-actionlint.bash \
+    && ./download-actionlint.bash \
```
massongit, line 142:
```suggestion
    && ./download-actionlint.bash \
    && rm download-actionlint.bash \
```

You can remove unnecessary scripts to reduce the image size.

lib/linter.sh
```diff
@@ -193,7 +195,7 @@ fi
 ##################
 # Language array #
 ##################
-LANGUAGE_ARRAY=('ANSIBLE' 'ARM' 'BASH' 'BASH_EXEC' 'CLANG_FORMAT'
+LANGUAGE_ARRAY=('ACTIONS' 'ANSIBLE' 'ARM' 'BASH' 'BASH_EXEC' 'CLANG_FORMAT'
```
massongit, line 198: I think that `GITHUB_ACTIONS` is clearer than `ACTIONS`.

admiralAwkbar: @massongit I agree think I will update it around

.automation/update-actions-version.sh
```diff
@@ -29,7 +29,7 @@ GetReleaseVersion() {
   echo "Getting the latest Release version from GitHub ..."
 
   # Get the latest release on the Repository
-  GET_VERSION_CMD=$(echo "${RELEASE_NAME}" | grep -E -o "v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+" 2>&1)
+  GET_VERSION_CMD="$(echo \"${RELEASE_NAME}\" | grep -E -o \"v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+\" 2>&1)"
```
massongit, line 32:
```suggestion
  GET_VERSION_CMD="$(echo "${RELEASE_NAME}" | grep -E -o "v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+" 2>&1)"
```

You may not need to escape the double quotes in `$()`.

PR #1777, opened 2021-07-17
Switch from wata727/tflint to ghcr.io/terraform-linters/tflint-bundle

wata727:
This PR switches TFLint image from `wata727/tflint` to `ghcr.io/terraform-linters/tflint-bundle`.

TFLint recently migrated the image registry from Docker Hub to GitHub Container Registry. See https://github.com/terraform-linters/tflint/pull/1155
Images hosted on Docker Hub will no longer be updated.

At the same time, the image to be referenced is changed from `tflint` to `tflint-bundle` here. In the near future, We plan to deprecate the bundled plugin, and if you still need to lint your AWS resources, you'll need to install the plugin with `tflint --init`. See https://github.com/terraform-linters/tflint/pull/1160

The `tflint-bundle` image includes the latest plugins for AWS, Azure, and Google as an option to address this deprecation.
https://github.com/terraform-linters/tflint-bundle

In my understanding, the Super-Linter aims to provide better defaults, and doesn't want to let users choose plugins and run `tflint --init` each time. I believe the `tflint-bundle` image is suitable for the Super-Linter use cases.

## Proposed Changes

1. Switch TFLint image from `wata727/tflint` to `ghcr.io/terraform-linters/tflint-bundle`
2. Add `COPY` instruction to copy pre-installed plugins
3. Add `plugin` declaration to default `.tflint.hcl`. This will only enable the pre-installed AWS plugin

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ FROM mstruebing/editorconfig-checker:2.3.5 as editorconfig-checker
 FROM yoheimuta/protolint:v0.32.0 as protolint
 FROM golangci/golangci-lint:v1.41.1 as golangci-lint
 FROM koalaman/shellcheck:v0.7.2 as shellcheck
-FROM wata727/tflint:0.29.1 as tflint
+FROM ghcr.io/terraform-linters/tflint-bundle:v0.30.0 as tflint
 FROM alpine/terragrunt:1.0.1 as terragrunt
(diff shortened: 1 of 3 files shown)


admiralAwkbar:
@wata727  thanks for the help on this!

PR #1830, opened 2021-08-03
adding black binary

### LINKED ISSUE #1828: Separate black and snakefmt depedencies

Currently, `black` is pinned to a fairly _old_ version due to `snakefmt`. This effectively prevents `super-linter` from upgrading `black` as it can cause conflicts and break `snakefmt`. While `snakefmt` is actively maintained, it may choose not to upgrade `black` for any number of reasons.

It would be great if the version of `black` used by `super-linter` was not dependent on `snakefmt` dependency requirements for users that don't use `snakefmt`. Even better would be the version of `black` being made configurable to match the requirements of the project using the linter.

I have opened up a PR at snakemake/snakefmt#116, but that does not seem like a long-term solution. The other alternative seems to be to not use `super-linter` and use `black` directly.

**Additional context**

This issue was brought up before when upgrading to `black` 20.0 in #1212, #807, #840, and #762. Would be good to solve this for good.

`black` officially releases self-contained binaries now, which could help. However, this may mean manually updating `black` instead of relying on dependabot.

admiralAwkbar:
- Using the black binary, vs using `pip install`
- This should allow for updated versions and stop causing issues with pip

This closes #1828

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,8 +129,12 @@ RUN pip3 install --no-cache-dir pipenv \
     && wget --tries=5 -q -O dotnet-install.sh https://dot.net/v1/dotnet-install.sh \
     && chmod +x dotnet-install.sh \
     && ./dotnet-install.sh --install-dir /usr/share/dotnet -channel Current -version latest \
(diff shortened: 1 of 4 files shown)


jalaziz:
That was quick, thanks @admiralAwkbar!!

PR #1834, opened 2021-08-05
Sqlfluff

### LINKED ISSUE #1812: How would other linters eg sqlfluff be added to this tool?
I like the elegance of running all linters the same way. But the tool isn't compatible with all linters. If a user wanted to add another linter (eg. sqlfluff), is that even possible?

Instructions on how to extend compatibility with other linters.

Alternative is to not use super-linter for some use cases. But then it is no longer "one linter to rule them all".

**Additional context**
None today. 

**Note**
If this is a request to add a new language, after submitting this issue check out [the wiki](https://github.com/github/super-linter/wiki/Adding-new-language-support) for more info on how to accomplish that.

Why did you bury the above link down here? You could have put it in the readme. I wonder if it answers my question.

admiralAwkbar:
This closes #1812 

- Adding `sqlfluff` to help lint SQL files


johnsandersnl:
@admiralAwkbar you have added sqlfluff (beta) but did not update the documentation. (Or I cannot find it.)
Our linting is now extremely slow and builds break.
I cannot disable any rules for sqlfluff since it is not clear which filename to use. I am now experimenting with different names for template files. Last resort would be to disable super-linter completely.

johnsandersnl:
Solved for me. 
I have just excluded sqlfluff locally (docker run -e RUN_LOCAL=true -e VALIDATE_SQLFLUFF=false)
Will add this to our build pipelines later.

PR #1837, opened 2021-08-05
Fix black binary download

jalaziz:
## Proposed Changes

1. Fix wget output option (uppercase `-O`)
2. Overwrtie the version of `black` installed by pip as it comes first in
     PATH

Unfotunately, this is not enough. Turns out the `black` binary is not fully self-contained and does not run on Alpine by default. It requires `glibc` which is installed, but even with that we see `/usr/local/bin/black: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory`. Still investigating this.

### Author/Contributor

### Reviewing Maintainer


jalaziz:
@admiralAwkbar heads up

admiralAwkbar:
@jalaziz thanks for the look on this!

jalaziz:
@admiralAwkbar I believe it's fixed now. I've tested this locally and it seems to finally work.

Installing zlib was required. Installed from archlinux seems pretty standard. See sgerrand/alpine-pkg-glibc#75 and https://github.com/AdoptOpenJDK/openjdk-docker/blob/master/dockerfile_functions.sh#L84.

PR #1838, opened 2021-08-06
move black back to pip

admiralAwkbar:
- move black back to pip as we have a lib issue


admiralAwkbar:
This is to help solve #1837 as its a bug for now

PR #1862, opened 2021-08-15
Use unprivileged user by default

### LINKED ISSUE #1861: Add non-root user to image
When one uses super-linter as VS Code dev container, VS Code tries to run the container as the `USER` specified in the upstream Dockerfile, otherwise it defaults to `root`. Your Dockerfiles currently don't contain `USER` and don't add a non-system user in the first place. So, VS Code ends up running the container as root. Now, when some directory is bind-mounted from the host system into a super-linter container (e.g., the workspace dir with `.git/`), this results in file ownership being 0/`root`.

Add a non-system user to the Dockerfiles, as suggested in the problem description.

Adding a such a user myself in a an extra layer. This means other super-linter users will continue running it needlessly as superuser (e.g., under CI).

sanmai-NL:
Fixes #1861.

## Proposed Changes

1. Add a non-system, unprivileged user to the container images.
2. Set this user as default.

### Author/Contributor
Since the current behavior hasn't been documented, the change doesn't have to either. It would help little to do document it now, since it is a design detail and immediately obvious when running the container (from the prompt).

### Reviewing Maintainer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -458,11 +458,14 @@ COPY TEMPLATES /action/lib/.automation
 ################################################
 RUN ACTIONS_RUNNER_DEBUG=true WRITE_LINTER_VERSIONS_FILE=true /action/lib/linter.sh
 
+############
+# Set user #
+############
+RUN addgroup -g 1000 superlinter && \
+    adduser -u 1000 -D -G superlinter superlinter
+USER superlinter
+
 ######################
 # Set the entrypoint #
 ######################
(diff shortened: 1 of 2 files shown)


admiralAwkbar:
@sanmai-NL Thanks for the push on this.
I have no issues with adding a non-root user to help. I believe its also considered a best practice as well :)

admiralAwkbar (approved):

PR #1909, opened 2021-08-30
Update black binary to 21.9b0

jalaziz:
[Release Notes](https://github.com/psf/black/releases/tag/21.9b0)

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -133,7 +133,7 @@ RUN pip3 install --no-cache-dir pipenv \
 ########################
 # Install Python Black #
 ########################
-    && wget --tries=5 -q -O /usr/local/bin/black https://github.com/psf/black/releases/download/21.7b0/black_linux \
+    && wget --tries=5 -q -O /usr/local/bin/black https://github.com/psf/black/releases/download/21.9b0/black_linux \
     && chmod +x /usr/local/bin/black
 ##############################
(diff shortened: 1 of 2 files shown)


jalaziz:
This is currently broken as it seems like the latest black binary is throwing a new error.

admiralAwkbar:
@jalaziz  Looks like this resolved itself and since we have redeployed were up to date?

jalaziz:
> @jalaziz Looks like this resolved itself and since we have redeployed were up to date?

The issue was resolved, but `black` hasn't published new binaries as far as I can tell. I'm watching out for that to update this PR once it's been deployed.

That being said, it looks like `snakefmt` recently merged in the change that would allow `black` to be updated through `pip`. That being said, the binary version of `black` is probably still useful long term.

I don't believe super-linter is up-to-date though as the binary is still pointing to 21.7b0, unless i missed something? The 21.8b0 binary is certainly broken and it doesn't seem like black has republished it yet, so we might have to wait until the next release.

jalaziz:
@admiralAwkbar This PR should be good now.

admiralAwkbar (approved):

admiralAwkbar:
@jalaziz Thanks for keeping up with this one!

PR #1949, opened 2021-09-09
Match AWS States file using "States" JSON key

thibaultdelor:
## Proposed Changes

Matching only on `"Resource": "arn` is too wide and will match also [aws json policy files](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_resource.html) for example.
Given the keys States and StartAt are required in any states file it makes sense to use one of them to narrow it down (why not even both...)

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/functions/detectFiles.sh
+++ b/lib/functions/detectFiles.sh
@@ -203,8 +203,8 @@ DetectAWSStatesFIle() {
   ###############################
   # check if file has resources #
   ###############################
-  if grep -q '"Resource": *"arn"*' "${FILE}"; then
-    # Found it
+  if grep -q '"Resource": *"arn' "${FILE}" &&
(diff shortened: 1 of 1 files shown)


lib/functions/detectFiles.sh
```diff
@@ -203,8 +203,8 @@ DetectAWSStatesFIle() {
   ###############################
   # check if file has resources #
   ###############################
-  if grep -q '"Resource": *"arn"*' "${FILE}"; then
```
thibaultdelor, line 206: `"*` at the end of the pattern seems unnecessary.
As a general rule `x*`, no matter x, at the beginning or the end of a grep pattern brings no value. That should probably be a lint rule in somewhere! :)

admiralAwkbar (approved):

PR #1952, opened 2021-09-11
Correct expressions in GitHub Actions workflows

### LINKED ISSUE #1918: Annoying email alert when commenting on issues
## Describe the bug

Keep getting an email with the when commenting on issues:

77E295E7-3FDE-4243-98D4-A09BC0DE00CB

## Steps to Reproduce

Comment on an issue.
Look at your email.

Not sure if specific to non-maintainers?

tunetheweb:
Fixes #1918 

## Proposed Changes

The expression syntax used by a number of workflows is incorrect.

If you use this syntax:
```yml
if: ${{ github.event_name }} == 'schedule'
```
this will ALWAYS evaluate to true as the evaluation happens in the `${{..}}` part, and the part after the `==` is ignored.

You should instead evaluate the whole expression within the `${{..}}` parts:
```yml
if: ${{ github.event_name == 'schedule' }}
```

Or alternatively the `${{..}}` part is actually optional within `if:` statements so this works too:
```yml
if: github.event_name == 'schedule'
```

This means both parts of the stale job are run (the scheduled part which adds stale label to issue, and also removes it necessary), and the second part (which removes the stale label from the commented on issue only). This means the stale label is attempted to be removed twice, hence why the job fails and you get the email.

I originally noticed this for the stale job after raising #1918 but it actually is wrong in other workflows too, meaning:
- deploys are incorrectly running (and failing) on forks.
- permissions to release are not being checked - I was originally worried when I saw this, but don't _think_ it is really an issue since only maintainers have access to release anyway which is when this workflow is triggered. Still it should be fixed.

### Author/Contributor

No change to documentation required.

### Reviewing Maintainer

### DIFF
--- a/.github/workflows/deploy-DEV-slim.yml
+++ b/.github/workflows/deploy-DEV-slim.yml
@@ -30,8 +30,8 @@ jobs:
     # Set the agent to run on
     runs-on: ubuntu-latest
     # Prevent duplicate run from happening when a forked push is committed
-    if: ${{ github.event_name }} == 'push' ||
-        ${{ github.event.pull_request.head.repo.full_name }} != ${{ github.repository }}
+    if: ${{ github.event_name == 'push' }} ||
+        ${{ github.event.pull_request.head.repo.full_name != github.repository }}
     ##################
     # Load all steps #
     ##################
--- a/.github/workflows/deploy-DEV-slim.yml
+++ b/.github/workflows/deploy-DEV-slim.yml
@@ -30,8 +30,8 @@ jobs:
     # Set the agent to run on
     runs-on: ubuntu-latest
     # Prevent duplicate run from happening when a forked push is committed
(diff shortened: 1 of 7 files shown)


admiralAwkbar:
@tunetheweb Amazing fix!
How did you track this down?

admiralAwkbar (approved):

tunetheweb:
> @tunetheweb Amazing fix!
> How did you track this down?

It took a lot of trial an error on my own fork to figure this out!! But it was annoying me 😁

I actually think this is very bad for GitHub Actions to allow this and silently ignore it. I raised it with the GitHub Security team before opening this PR but they didn't see it as a security issue. So opened this PR and raised this: https://github.community/t/github-actions-if-statements-can-easily-be-coded-incorrectly/200621

If you have any contacts in GitHub Actions team it would be good to point them at this!

PR #1987, opened 2021-09-20
Support Containerfile as alternative to Dockerfile

ioanrogers:
## Proposed Changes

Please add support for `Containerfile` as an alternative to `Dockerfile`, as suggested in containers/buildah#1853
It's not a very popular alternative as docker is to containers as Xerox is to photocopiers, but it is supported by other tools, e.g.
* ekalinin/Dockerfile.vim#73
* microsoft/vscode#93184

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar:
@ioanrogers very cool!

PR #1991, opened 2021-09-23
docs: correct GITHUB_CUSTOM_API_URL format

patrickhuy:
Fixes #

## Proposed Changes

In the README.md file the example value of `GITHUB_CUSTOM_API_URL` for usage on GHES ends in a `api/v3/` this is incorrect, instead `api/v3` needs to be used. Also it states that the defautl value is `api.github.com`, the true default value is `https://api.github.com`, this again is confusing whether or not the `https` needs to be included or not (it does need to be included)

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
thanks for the catch!

admiralAwkbar (approved):

PR #2007, opened 2021-09-29
update clang-format

admiralAwkbar:
This will pull in `clag-format` as a layer instead of building it every time.

The build for `clang-format` is usually between `15-20` minutes... so this is nice


admiralAwkbar:
Screen Shot 2021-09-29 at 12 05 47 PM

This is insane.. its there...

Issue #1810, opened 2021-07-27
Add textlint


massongit:
[textlint](https://textlint.github.io/) is a text linter.
If we add it, we can improve the quality of the documents in our repositories.

massongit:
re-open

PR #2006, opened 2021-09-29
Add textlint

massongit:
Fixes https://github.com/github/super-linter/issues/1810

## Proposed Changes
[textlint](https://textlint.github.io/) is a text linter.
If we add it, we can improve the quality of the documents in our repositories.
Only the rule [textlint-rule-no-dead-link](textlint-rule-no-dead-link) that make sure the URL is alive are applied here.

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [ ] Label as `breaking` if this is a large fundamental change
- [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #2021, opened 2021-10-02
Add textlint

massongit:
Fixes https://github.com/github/super-linter/issues/1810

## Proposed Changes
(This PR is a modification of a rule that apply based on https://github.com/github/super-linter/pull/2006 )
[textlint](https://textlint.github.io/) is a text linter.
If we add it, we can improve the quality of the documents in our repositories.
The rule [textlint-rule-terminology](https://github.com/sapegin/textlint-rule-terminology) and the filter [textlint-filter-rule-comments](https://github.com/textlint/textlint-filter-rule-comments) are applied here.

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar:
@massongit 
Im not a big fan of the `dead-link` as it seems to time out and hit rate limits without a way to resolve.
I noticed when we had added this it was failing every time with rate limits... Im not sure we want to add this if its going to constantly fail and no way to fix that

massongit:
textlint-rule-no-dead-link has the problem you seid.
Therefore, I have changed to apply textlint-rule-terminology.

admiralAwkbar:
@massongit I like this rule a lot better... If you can clean up the package-lock I think I'm good to get this merged up

massongit:
I merged `main`.

PR #2052, opened 2021-10-15
Upgrade SQLFluff to 0.8.1 and add new SQLFluff DBT Templator dependency

### LINKED ISSUE #2051: Add optional DBT package for Super-Linter for SQLFluff from v0.7.0
Hi, I'm one of the maintainers of SQLFluff.

From our next release (0.7.0) we are splitting our some of the functionality into a separate, optional plugin that users can install only if they need that functionality.

The reason it's being split out is not all users need this, and it allows a separate release cadence, and to clean up our code, and to reduce the need for users who don't need it to install this code. There is no harm to install it even if it's not used.

Therefore I think the GitHub Super-Linter should include this by default. It won't be used unless referred to in the user's SQLFluff config file, but would need to be included in your Docker builds to allow them to continue using this functionality (which they may well have been happily using in the current version of the Super-Linter ). 

I've raised https://github.com/sqlfluff/sqlfluff/issues/1600 in the SQLFluff repo, as a reference for this to track adding this here so feel to ask any questions here or there.

The plugin is called `sqlfluff-templater-dbt` is currently part of the main SQLFluff repo, but is a separate pip package (the package is already available now as an alpha release (0.7.0a4): https://pypi.org/project/sqlfluff-templater-dbt/).

As it's an optional dependency it's not included in our Python `requirements.txt` dependence file so won't be installed by default based on how SQLFluff is currently pulled into the Super-Linter. How can we get this plugin added in as a dependency to the Super-Linter  at the same time as SQLFluff to ensure users can continue to avail of that functionality if they want it? As I said above, those who do not use DBT will not have any issues with this package being installed.

Happy to open a PR for this if you could give me some pointers of where to add it. Note this should only be added when the Super-Linter  upgrades to SQLFluff 0.7.0 which we expect to send live in next few days.

tunetheweb:
Fixes #2051 

## Proposed Changes

SQLFluff 0.7.0 moves one of it's features to an optional package. This change adds that optional package so it works in Super-Linter same as it did for previous versions.

### Author/Contributor

No documentation changes needed.

### Reviewing Maintainer


tunetheweb:
Updated to 0.8.1

admiralAwkbar (approved):

admiralAwkbar:
@tunetheweb Hey!
Was trying to get this running but hitting issue with missing `dbt`
I have tried to install with pip on my side, but it seems to hang badly...
Not sure if you have seen that on your side

admiralAwkbar:
@tunetheweb I would have assumed adding it to the pip file would have done it, but seems a not well-connected dep?

tunetheweb:
Just set up a fresh VirtualEnv and just did:

```
pip install sqlfluff-templater-dbt
```

And worked fine for me.

Any more info on what you're seeing?

Vendetta26:
gh pr out-of-date

tunetheweb:
Closing this. It appears Super Linter users do not use this functionality since heard no complaints. Can revisit if there is a desire to use this.

PR #2116, opened 2021-11-11
Fix calling GitHub API (github.api.~ -> github.rest.~)

massongit:
## Proposed Changes
https://github.com/github/super-linter/runs/4169311424?check_suite_focus=true#step:11:27

```
Error: Unhandled error: TypeError: Cannot read property 'issues' of undefined
```

I replace `github.api.~` in the scripts using `actions/github-script@v5` with `github.rest.~` to fix the above error.

https://github.com/actions/github-script#breaking-changes-in-v5
>Version 5 of this action includes the version 5 of `@actions/github` and `@octokit/plugin-rest-endpoint-methods`. As part of this update, the Octokit context available via `github` no longer has REST methods directly. These methods are available via `github.rest.*` - https://github.com/octokit/plugin-rest-endpoint-methods.js/releases/tag/v5.0.0
>
>For example, `github.issues.createComment` in V4 becomes `github.rest.issues.createComment` in V5
### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.github/workflows/deploy-PROD-slim.yml
+++ b/.github/workflows/deploy-PROD-slim.yml
@@ -133,7 +133,7 @@ jobs:
           # https://octokit.github.io/rest.js/v18#issues-create
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
-            const create = await github.api.issues.create({
+            const create = await github.rest.issues.create({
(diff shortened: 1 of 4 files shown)


admiralAwkbar (approved):
nice fix!

PR #2160, opened 2021-11-17
Update .python-black template with example config

MariusStorhaug:
Added a config example for python black linter

Fixes #

Update .python-black template with example config

## Proposed Changes

1. Update .python-black template with example config

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar:
@MariusStorhaug Thank you for the commit :)

Can you add some linkage to documentation to the template so users can see where they may need to update file

Also, do we need to pass a flag to use the config or does it scan and find it?

just trying to understand better

thanks!

MariusStorhaug:
For me I used the following article as the reference.
[Black - Configuration Format](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-format)

I could work to include some additional doc or update some links to point to this article, but that might be better done in the specific linter repo. What do you think makes more sense?

There is no flag needed to use this, its just an update to an existing file in the repo (using the default [template file](https://github.com/github/super-linter/blob/main/TEMPLATES/.python-black)). With the [default settings](https://github.com/github/super-linter#environment-variables) this file would need to be put under the .github/linter folder in the consumers repo, and the super-linter pick that up.

Vendetta26:
Add .patch or .diff to the end of URLs for Git’s plaintext views

MariusStorhaug:
Any comment on this PR? Its a simple starting template for black.

MariusStorhaug:
Taking too long, and doing some cleanup on my repos so deleting my fork and closing the issue.

PR #2178, opened 2021-11-25
Reduce duplication using a single Dockerfile

ferrarimarco:
## Proposed Changes

1. Reduce code duplication by using a single Dockerfile for both the standard and slim flavors of the super-linter image. To stop at the right point, we use target build stages.
1. Don't check if the versions file exist before building it. Setting the `WRITE_LINTER_VERSIONS_FILE` to `true` would suffice. This is necessary because we build the versions file twice (slim image, standard image) because we add more linters in the standard image, and need to rebuild the file to include those additional linters.
1. Fix setting BUILD_DATE in all the workflows. It was empty before this fix.

PS: I know that there's a lot of duplication in GitHub Actions workflow definitions, but it's a job for another day.

### Author/Contributor

### Reviewing Maintainer

## Requisites

- [x] #2195


admiralAwkbar (approved):
@ferrarimarco This is amazing

PR #2180, opened 2021-11-26
Lint the whole code base with JSCPD

### LINKED ISSUE #1619: Feed jscpd All Files At Once
## Describe the bug

jscpd's purpose is the detection of duplicated code, so invoking it separately on each individual file defeats its purpose by limiting it to only detecting duplicate code within each individual file.

## Expected behavior

jscpd is called one time on the entire code base, so it catches duplication between files, and it's threshold setting works properly. It doesn't make much sense to run it only on changed files either in my opinion, because then it will miss all the cases where code was copied from an unmodified file to a file in the pull request. Perhaps the ideal thing would be to only report duplication involving the modified files, but that seems like way too much work for one issue.

## Steps to Reproduce

1. Create a .github/linters/.jscpd.json.
2. Create some duplicate files that lack duplication within them.
3. Open a pull request, or take whatever alternative action is needed to trigger the action.
4. No error is reported, regardless of whether the entire code base is linted or only the modified files.

ferrarimarco:
Fixes #1619 

## Proposed Changes

This PR adds support to run JSCPD on the codebase as a whole if `VALIDATE_JSCPD_ALL_CODEBASE` is set to `true`.

### Author/Contributor

### Reviewing Maintainer

## Requisites

- [x] #2199
- [x] #2215


ferrarimarco:
We could have a separate flag. It's easy to add, and could be flagged as non-breaking that way. Maybe it's worth adding?

admiralAwkbar:
@ferrarimarco  I like the idea, its more of a feature add, vs changing the way it works...

ferrarimarco:
@admiralAwkbar done, defined a new variable as requested. The thing is that we need to check our own duplications :) (or tune JSCPD until we're ready to deal with that).

ferrarimarco:
@admiralAwkbar all green! Merging the latest changes from main...

PR #2224, opened 2021-12-13
Update black binary to 21.12b0

pdonorio:
Release reference:
https://github.com/psf/black/releases/tag/21.12b0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -108,7 +108,7 @@ RUN curl --retry 5 --retry-delay 5 -sL https://cpanmin.us/ | perl - -nq --no-wge
     ########################
     # Install Python Black #
     ########################
-    && wget --tries=5 -q -O /usr/local/bin/black https://github.com/psf/black/releases/download/21.11b1/black_linux \
+    && wget --tries=5 -q -O /usr/local/bin/black https://github.com/psf/black/releases/download/21.12b0/black_linux \
     && chmod +x /usr/local/bin/black \
     #######################
     # Installs ActionLint #


ferrarimarco:
@pdonorio thanks for this PR.

@admiralAwkbar why aren't we installing black via pip? I remember some dependency conflicts? Maybe it's time to retry?

jalaziz:
> @pdonorio thanks for this PR.
> 
> @admiralAwkbar why aren't we installing black via pip? I remember some dependency conflicts? Maybe it's time to retry?

@ferrarimarco This was changed because `snakefmt` has a dependency on `black` that prevented upgrading to 21.x releases. While the `snakefmt` dependency on `black` has been upgraded, we'll run into the same issues with 22.x releases and so forth.

The best solution at the time seemed to be to use the "binary" version of `black` to allow independent upgrades.

pdonorio:
Hello, any reason why this is not being merged?

admiralAwkbar:
@pdonorio As you can see from the information above by @ferrarimarco , This seems to break the current model and will need a better way to install or we have broken upgrades every time.

jalaziz:
> @pdonorio As you can see from the information above by @ferrarimarco , This seems to break the current model and will need a better way to install or we have broken upgrades every time.

@ferrarimarco @admiralAwkbar To be honest, the current way of installing using one `requirements.txt` has been broken for a while. The interdependencies of the different formatters prevent independent upgrades.

I've been thinking about it more and the best solution might be independently managed virtual environments (through pipx or another tool). The challenge would then be having to have a separate `requirements.txt` for each tool if you want depdendabot to handle updates.

admiralAwkbar:
@jalaziz I think that might be the best way to achieve this... With so many tools and deps, this has become a burden and that sounds like the right answer to getting this more modular, and supportable.
Just need to put in the work to get to that format

admiralAwkbar:
This was solved in the python deps pr #2294

PR #2298, opened 2022-01-06
Remove .rustup after compilation

lindluni:
I profess to not be a rust expert (or having ever used it) but after cargo/clippy/rustfmt are compiled, why do we need to leave the .rustup directory behind, its over 1GB of space and the single largest directory.

This moves the compiled binaries out of `/usr/lib/.rustup` after compilation, and removes the .rustup directory. Whats the effect of removing the .rustup directory

@yanganto it looks like you added Rust, what are the implications of removing the .rustup directory after the binaries have already been built?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -445,17 +445,18 @@ RUN ln -s /usr/bin/rustup-init /usr/bin/rustup \
     && rustup component add rustfmt --toolchain=stable-x86_64-unknown-linux-musl \
     && rustup component add clippy --toolchain=stable-x86_64-unknown-linux-musl \
     && mv /root/.rustup /usr/lib/.rustup \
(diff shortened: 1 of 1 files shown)


yanganto:
@lindluni 
Rust has two kinds of linter, one is `rustfmt` and the other is `clippy`.

As far as I know, `clippy` needs to compile the rust project to do much detail linter more than a syntax deep, so most files in `.rustup` are needed.  

As you can see, the test for `clippy` did not pass when `.rustup` removing.
https://github.com/github/super-linter/runs/4723943242?check_suite_focus=true#step:15:12672

There may be some files under `.rustup` that can be removed, but not all of them.  
Thanks for take care the linters with Rust.

PR #2322, opened 2022-01-11
Update Dependabot for python virtual envs

jalaziz:
## Proposed Changes

1. Simplify virtualenv installation to use multiple requirements files in
   a single directory. This eliminates the packages.txt file and provides
   an easier path to adding new tools in the future. It also allows us to
   simplify the dependenabot configuration as all requirements files are
   in a single directory.
2. Update dependabot.yml to point to the new Python dependencies folder.
### Author/Contributor

### Reviewing Maintainer


jalaziz:
@lindluni I did some testing with Dependabot and it seems like Dependabot is not smart enough to handle conflicting dependencies (seems like it expects a build failure to catch that), but it is smart enough to handle multiple files in the same directory.

What this gets us is a slightly simplified Dependabot configuration. Since it recognizes any `.txt` file as a python requirements file, we can one per tool. Also, since Dependabot doesn't actually try to figure out if the versions conflict across files, it will go ahead and put up PRs to update any dependency across all files that list it. The only downside here is it may eagerly update dependencies with versions that don't meet the transitive requirements, but that would fail on installation (which should fail the PR checks).

admiralAwkbar (approved):
Seems like a solid update

admiralAwkbar:
@lindluni Any issues?

lindluni:
LGTM: I see now, dependabot doesn't support recursive wildcards for the directory to scan 😢

lindluni:
LGTM: I see now, dependabot doesn't support recursive wildcards for the directory to scan 😢

jalaziz:
> LGTM: I see now, dependabot doesn't support recursive wildcards for the directory to scan 😢

Yeah, sorry, I could've clarified that more on the original PR. One of the more annoying dependabot limitations.

PR #2362, opened 2022-01-20
docs: Alphabetize lists

rasa:


admiralAwkbar (approved):
Thank you for the housekeeping! It drives me nuts as well

PR #2363, opened 2022-01-20
fix: Output gitleaks version

rasa:
See https://github.com/zricethezav/gitleaks#usage and
https://github.com/rasa/scoop-directory/runs/4883841409?check_suite_focus=true#step:4:139


admiralAwkbar:
@rasa thanks for the fix!

PR #2399, opened 2022-01-27
adding lind command output to success

admiralAwkbar:
This is an interesting addition...

This would now start showing the output of all linter commands on success...

It has come to the attention that some linters, such as yamllint, show warnings, but users may never be aware of them as its masked by the `0` return code and never printed.

Most outputs seem to have no real output but a `0` return code.
Some such as `GitLeaks` have returns like:
Screen Shot 2022-01-27 at 3 21 35 PM

So this would cause some additional lines in the output log...

@lindluni @zkoppert @nemchik @Hanse00 @GaboFDC @ferrarimarco
Does anyone have any for or against this small but likely impactful fix?

### DIFF
--- a/lib/functions/worker.sh
+++ b/lib/functions/worker.sh
@@ -309,6 +309,7 @@ function LintCodebase() {
           # Success #
           ###########
           info " - File:${F[W]}[${FILE_NAME}]${F[B]} was linted with ${F[W]}[${LINTER_NAME}]${F[B]} successfully"
+          info "   - Command output:${NC}\n------\n${LINT_CMD}\n------"
         fi
       else
         #######################################


ferrarimarco (approved):

zkoppert (approved):
👍🏻

ferrarimarco:
@admiralAwkbar I know this is closed, but wouldn't it be useful to check it the `LINT_CMD` size is `>0` before printing the output line? We might spare some space.

Example of the issue:

```
2022-02-23 19:31:16 [INFO]      - Command output:
------

------
```

admiralAwkbar:
@ferrarimarco great idea!

PR #2454, opened 2022-02-09
feat: add SSH key support

colwynlegitscript:
In order to properly run the module mode in `tflint` we need to be able to pull in the modules that we depend on. That works great for public modules, but if someone maintains their own private GitHub repos that they use as modules then their builds will fail. This PR adds the ability to use an SSH key that has access to those repos.

Additionally, the last PR I put up (https://github.com/github/super-linter/pull/2297) adds `terraform get` to the `tflint` workflow which introduces the same requirement even if module mode is disabled - my bad, I didn't fully think through the consequences of adding `terraform get`. We could walk that back by removing `terraform get` and explicitly stating that Super Linter doesn't support module mode using private modules in `tflint`. This change to add `SSH_KEY` might be turning this tool into too much of a build tool, this does feel potentially a little overreaching to me but I'm happy to have the conversation.

One last thing, if we do go ahead with adding the `SSH_KEY` parameter, should we also add the keys for other popular Git management platforms? I'm thinking about BitBucket and GitLab mainly. They both publish their SSH key fingerprints ([BitBucket](https://support.atlassian.com/bitbucket-cloud/docs/configure-ssh-and-two-step-verification/), [GitLab](https://docs.gitlab.com/ee/user/gitlab_com/)) so it would be trivial to add them.

## Proposed Changes

1. Adds the `SSH_KEY` environment variable which starts an `ssh-agent` and adds the key. It also verifies and adds the github.com SSH key to `known_hosts`
2. Adds the `SSH_SETUP_GITHUB` environment variable with adds the github.com SSH key to `known_hosts`. NOTE: This option is ignored if `SSH_KEY` is provided and github.com is always added to `known_hosts`
3. Adds the `SSH_INSECURE_NO_VERIFY_GITHUB_KEY` environment variable. This is primarily in case github.com changes their fingerprint for whatever reason and people need to get their builds passing before we update the fingerprint in Super Linter

### Author/Contributor

### Reviewing Maintainer


colwynlegitscript:
Looks like the code coverage report is having issues in the production build. The tests themselves pass so I _think_ this is good to look at. Sorry in advance if I broke something in the build and I'm just not reading that output correctly.

admiralAwkbar:
@colwynlegitscript This is great!

Thank you for the hard work on it. I added a little bit of our style to it and were off and to the races...

admiralAwkbar:
@colwynlegitscript The build failures are because you don't have access to the secrets for the values... its frustrating :)

colwynlegitscript:
Excellent! I read through some of my docs again and figured I could make them a little more clear. I'm actually done committing now unless y'all have some feedback. Thanks for taking a look at this!

colwynlegitscript:
> The build failures are because you don't have access to the secrets for the values... its frustrating :)

Ahh that makes sense. Welp, my bad. My commit is gonna fail again because of that.

PR #2459, opened 2022-02-09
fix: tflint should clean up after itself

### LINKED ISSUE #2433: Permission error on file after TFLINT
## Describe the bug

Once the super-linter (latest) runs a TF scan and we try to run terraform init, we get the following error:
```
Unable to write the module manifest file: open
.terraform/modules/modules.json: permission denied
```
If we use v4.8.5 we do not have the issue. It looks like the latest release is causing this (v4.8.7)

## Expected behavior

When linter runs VALIDATE_TERRAFORM_TFLINT, it should complete and shouldn't cause errors when running terraform init.

## Steps to Reproduce

Workflow steps:

```yml
    steps:
      - name: Checkout
        uses: actions/checkout@v2

      - name: Lint Code Base
        uses: github/super-linter@main
        env:
          VALIDATE_TERRAFORM_TFLINT: true

      - uses: hashicorp/setup-terraform@main
        with:
          terraform_version: latest

      - name: Terraform Init
        id: init
        run: terraform init -backend=false
```

## Logs

```
Unable to write the module manifest file: open
.terraform/modules/modules.json: permission denied
```

## Additional context

The error sounds like the file is left open after lint, causing this error (lock file).

colwynlegitscript:
Fixes #2433

Restores or cleans up the `.terraform` folder after `tflint` runs.

This adds a caching layer to the `TERRAFORM_TFLINT` worker that keeps track of the first time a directory was seen. If an existing `.terraform` folder is that directory, then that folder will be saved to `/tmp` and restored after the `TERRAFORM_TFLINT` worker finishes. If there is no `.terraform` folder, then the `.terraform` folder that was created during linting is simply removed. This should unblock workflows that either:

* Expect no `.terraform` folder to be created by Super Linter
* Expect their existing `.terraform` folder to stay the same after Super Linter is run

This should support all the same workflows that existed before `terraform get` was added in https://github.com/github/super-linter/pull/2297

NOTE: The caching layer does add some complexity, but is meant to deal with the edge case where someone may have many `.tf` files in a particular directory. The cache allows Super Linter to only run `terraform get` a single time per directory. Without the cache, `terraform get` would be run once for every since file in a directory, potentially causing significant runtime increases for users with many `.tf` files.

## Proposed Changes

1. `tflint` cleans up after itself, restoring the `.terraform` folder to its previous state. If there was no `.terraform` folder before Super Linter ran, no `.terraform` folder will exist after it runs.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/functions/worker.sh
+++ b/lib/functions/worker.sh
@@ -28,6 +28,11 @@ function LintCodebase() {
   ##########################
   LIST_FILES=()
 
+  ###################################################
+  # Array to track directories where tflint was run #
+  ###################################################
+  declare -A TFLINT_SEEN_DIRS
+
   ################
   # Set the flag #
   ################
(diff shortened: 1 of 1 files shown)


admiralAwkbar (approved):
Very nicely done.
Seems to be straight forward and clean
thank you for all the improvments

colwynlegitscript:
Thank you for maintaining this project! It's been really helpful for us. Updating it has been pretty straightforward since it's well organized too.

PR #2576, opened 2022-03-01
Lint the whole Ansible directory

### LINKED ISSUE #2080: Lint the whole ANSIBLE_DIRECTORY instead of individual files

The way the super-linter works, is that is sends to the linter each individual file to be linted. 
Unfortunately this conflicts with the way ansible-lint works which is aware of ansible structure and can determine the purpose of files and directories based on their name and parse everything in one go

This not only causes a ton of false-positives (as reported in #1671 but it also causes issues with role files etc) but it also makes ansible linting take a massive amount of time. I can lint my entire collection in a few seconds, compared to a couple of minutes through super-linter.

I understand that the approach of this tool is to be able to lint only the changed files, but the ansible-linter approach needs to be taken properly into account.

A middle-of-the-road approach would be to provide an env variable to be used to mark that ansible should use the single-pass approach (e.g. ANSIBLE_SINGLE_PASS) which, when true, will run the ansible-lint just once against the whole folder.

ferrarimarco:
Fixes #2080 

## Proposed Changes

1. Pass `ANSIBLE_DIRECTORY` to ansible-lint, instead of passing file by file.
2. Introduce a `ANSIBLE_ROLES_PATH` variable to point to the directory where roles are.
3. Remove the function to detect Ansible files because we don't call it anymore.
4. Increase the verbosity setting of ansible-lint so we get the list of linted files.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
This still needs some tuning on selecting the tests to run.

admiralAwkbar (approved):
Amazing

PR #1937, opened 2021-09-07
Install type stubs at mypy runtime

JonZeolla:
# Proposed Changes

1. This changes `mypy` to install missing type hints at runtime.  Prior to `mypy` 0.9* this was the default.  For additional context see python/mypy#10669 and python/mypy#10632

### Author/Contributor
- [X] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [ ] Label as `breaking` if this is a large fundamental change
- [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #2607, opened 2022-03-07
Copy over Python headers during staged build

HippocampusGirl:
- I am using the Mypy type checker as part of the super linter. Mypy requires type information for packages imported by my code. For some packages it can get these by installing type stubs, which super linter supports nicely (https://github.com/github/super-linter/pull/1937), but for other packages such as Numpy and Scipy, the package itself needs to be installed. 

- I would like to be able to run `pip install numpy` inside the container, but sadly this fails, because the Python headers are missing from the container.

- One commenter https://github.com/github/super-linter/issues/157#issuecomment-648850330 suggests some ideas for how to bind external packages into the super linter, but unfortunately that will not work for Mypy, which does not check `PYTHONPATH` or other environment variables for external directories containing packages. Instead, everything needs to be installed directly into the virtual environment from which Mypy is run. Proposals to change this have not yet been merged, see for example https://github.com/python/mypy/issues/5701.

## Proposed Changes

1. Copy over the Python headers in `/usr/local/include` from the base image during the container build. This will add approximately one megabyte to the container size.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -371,6 +371,7 @@ COPY --from=base_image /usr/bin/ /usr/bin/
 COPY --from=base_image /usr/local/bin/ /usr/local/bin/
 COPY --from=base_image /usr/local/lib/ /usr/local/lib/
 COPY --from=base_image /usr/local/share/ /usr/local/share/
+COPY --from=base_image /usr/local/include/ /usr/local/include/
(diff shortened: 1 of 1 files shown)


HippocampusGirl:
Let's hope that rebasing on the main branch will fix the vulnerability scan failure :-)

PR #2670, opened 2022-03-24
Add Android support for Kotlin

### LINKED ISSUE #2666: Add android support for ktlint
Please add Android support for ktlint 🙏 

ktlint command accepts a `--android` parameter to use Kotlin codestyles more suitable for Android.

We could achieve that by adding the following changes to `/lib/linter.sh`:
```bash
KOTLIN_ANDROID="${KOTLIN_ANDROID:-false}"

if [ "${KOTLIN_ANDROID}" == 'false' ]; then
  LINTER_COMMANDS_ARRAY['KOTLIN']="ktlint"
else
  LINTER_COMMANDS_ARRAY['KOTLIN']="ktlint --android"
fi
```

Then, we will be able to specify which linter command to use in our `linter.yml`:
```yaml
VALIDATE_KOTLIN: true
KOTLIN_ANDROID: true
```

alexandrefavre4:
Close #2666

## Proposed Changes

1. Add Android support for Kotlin

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
Hi @alexandrefavre4 , thanks for the PR.

That's a nice start. I think we need a couple of things:

- Is there a way we can identify the files that we should lint this way? We often look at their contents, path, name, or extension.
- After we know this, you need to add these files in in the FILE_ARRAY_KOTLIN_ANDROID array. [Here is an example of how this is done for Kotlin](https://github.com/github/super-linter/blob/c5e5046b8b0dc3fd997b7bc947cc1813a6640a0a/lib/functions/buildFileList.sh#L560).
- We need tests of files to lint, both "good" ones, and the ones for which we expect the linting to fail. [Example for Kotlin](https://github.com/github/super-linter/tree/main/.automation/test/kotlin).

Thanks!

alexandrefavre4:
> Hi @alexandrefavre4 , thanks for the PR.
> 
> That's a nice start. I think we need a couple of things:
> 
> * Is there a way we can identify the files that we should lint this way? We often look at their contents, path, name, or extension.
> * After we know this, you need to add these files in in the FILE_ARRAY_KOTLIN_ANDROID array. [Here is an example of how this is done for Kotlin](https://github.com/github/super-linter/blob/c5e5046b8b0dc3fd997b7bc947cc1813a6640a0a/lib/functions/buildFileList.sh#L560).
> * We need tests of files to lint, both "good" ones, and the ones for which we expect the linting to fail. [Example for Kotlin](https://github.com/github/super-linter/tree/main/.automation/test/kotlin).
> 
> Thanks!

Hey @ferrarimarco, there is no difference between standard Kotlin and Android Kotlin files. Both validators should run on `.kt` and `.kts` files. I added the files to `FILE_ARRAY_KOTLIN_ANDROID`, created the test cases for this new validator and fixed the existing good Kotlin test case (incorrect filename). Are we ready for the review?

ferrarimarco:
I think we're almost there. Let's wait for the CI jobs.

alexandrefavre4:
@ferrarimarco do you have any idea when we can expect a new release?

ferrarimarco:
@admiralAwkbar may know :)

admiralAwkbar:
@alexandrefavre4 @ferrarimarco I need to cut one...
Probably today :)

PR #2674, opened 2022-03-24
trying to create diagram

admiralAwkbar:
- Creating repo diagram for fun


ferrarimarco:
@admiralAwkbar I don't know if you noticed, but the diagram is not displayed in the README. :(

admiralAwkbar:
I just looked and it was working for me... Not sure if I have some auth...

ferrarimarco:
Also @admiralAwkbar I think we need to limit when the repo visualization workload runs because it's causing some builds to fail

admiralAwkbar:
Hey! Lol I forgot it was only going to be weekly and was testing it... Then actions was messing up and I forgot... Will update

ferrarimarco:
btw I get a 403 when trying to display the image :)

PR #2679, opened 2022-03-25
weekly

admiralAwkbar:
- run the visualization weekly
@ferrarimarco This was the original goal and got sidetracked with Actions failures due to GitHub DB issues :)


ferrarimarco:
Thanks!

PR #2786, opened 2022-04-12
SQLFluff 0.12.0 upgrade

### LINKED ISSUE #2785: SQLFluff 0.12.0 now requires the dialect to be specified
Hi, I'm one of the maintainers of SQLFluff and want to make you aware that SQLFluff 0.12.0 now requires the SQL dialect to be specified, either as a command line argument, or in the config file in the folder used.

The default dialect was our `ansi` dialect, which is more a generic base dialect rather than one that should really be used. There is no ANSI database, but various variants of this (Postgres, Oracle, T-SQL...etc.) each with slight differences from ANSI.

Although SQL looks the same, the subtle differences really do matter. ANSI SQL is not an implemented language - you cannot use an ANSI SQL server anywhere because it doesn't exist. EVERY SQL server is using significant variants on it. Be that additional (reserved) keywords, additional syntax, or unimplemented ANSI grammar. So ANSI is a poor choice for a default. But the SQL world is so fragmented that any other choice has its own issues too. We really do need the user to tell us what SQL dialect they are using for us to properly be able to lint their SQL.

Therefore, it's very unlikely users really want to use the `ansi` dialect, and there's likely a much better dialect that should be used instead (or `ansi` can be explicitly selected if they really want it). So it wasn't a great default in all honesty.

We were getting a number of support queries from, for example, Postgres users stating their Postgres SQL wasn't working in SQLFluff, when it would be - if they just knew to set the dialect. We therefore took the decision to remove the default dialect and insist that users set this, either in the command line (using the `--dialect ansi` option for example), or via a `.sqlfluff` config file.

We also hope getting users use to configuring SQLFluff will allow them to look at setting up the other config to their needs, rather than just accepting the defaults.

This is causing the tests to fail in #2779. I can provide a PR to fix that.

But you may also receive reports from users on your next upgrade so opening this issue to make you aware of that.

tunetheweb:
Fixes #2785 
Supersedes #2779

## Proposed Changes

SQLFluff 0.12.0 now requires a dialect to be set and no longer provides a default

This change:

1. Sets a default .sqlfluff config file in the `.automation/test/sqlfluff/` folder so tests can pass.
2. Updates the README in that folder to explain this
3. Updates the example config file to the latest set of options available (unrelated to above, but it was a bit out of date so included that in here).

Note Super Linter users who use SQLFluff will similarly need to configure SQLFluff now, after this upgrade is rolled out. The reasons for this breaking change are given in #2785 

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.automation/test/sqlfluff/.sqlfluff
+++ b/.automation/test/sqlfluff/.sqlfluff
@@ -1,0 +1,2 @@
+[sqlfluff]
+dialect = ansi
--- a/.automation/test/sqlfluff/README.md
+++ b/.automation/test/sqlfluff/README.md
@@ -4,7 +4,9 @@ This folder holds the test cases for **SQL**.
 
 ## Additional Docs
 
-No Additional information is needed for this test case.
+From version 0.12.0 SQLFluff requires a dialect to be set, and no longer sets a default. This can be provided as a command line argument, or a `.sqlfluff` config file (either in the usualy place for SQLFluff config files, or within the folder containg the SQL).
+
+For SQLFluff we have added a default `.sqlfluff` config file in this test directory.
 
 ## Good Test Cases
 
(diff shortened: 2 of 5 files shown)


admiralAwkbar:
@tunetheweb Thanks for the commit!
That dumb URL failed as always... re-running

PR #2873, opened 2022-05-09
Update NPM deps

lindluni:
Update NPM deps


lindluni:
I will work on fixing this tomorrow.

We need to remove support for `ts-standard` it is no longer maintained as the updates required to move to the `es-lint@v8` engine require a rewrite of the entire `ts-standard` core packages. The maintainers of `ts-standard` have stated they do not intend to do the work, and would prefer if the team who created `standard` itself instead took over the project: https://github.com/standard/ts-standard/issues/171#issuecomment-921521437

Users can use eslint directly as we already include support for `@typescript-eslint` in Super Linter. Users who want to continue to use `ts-standard` can use the last version of Super Linter to include `ts-standard`.

If in the future support changes, we can bring `ts-standard` back, but we can't hold back all of our other deps (which include security updates) because of one unmaintained package.

lindluni:
FYI: @admiralAwkbar ^^^

admiralAwkbar:
@lindluni make it so....

PR #2959, opened 2022-05-25
adding log level

### LINKED ISSUE #2954: tflint failing in slim-v4.9.3 "Plugin `aws` not found in /github/home/.tflint.d/plugins"
## Describe the bug

We're running super-linter on our codebase that is largely made up of terraform files. it has just failed with error:

```shell
2022-05-24 20:26:03 [INFO]   File:[/github/workspace/repo/path/to/file.tf]
2022-05-24 20:26:03 [ERROR]   Found errors in [tflint] linter!
2022-05-24 20:26:03 [ERROR]   Error code: 1. Command output:
------
Failed to initialize plugins; Plugin `aws` not found in /github/home/.tflint.d/plugins
```
Note that we don't use AWS, we don't use any AWS terraform providers.

I note that this is running on v4.9.3:

> Status: Downloaded newer image for ghcr.io/github/super-linter:slim-v4.9.3

I have compared to a successful run of the super-linter 6 horus ago which also used tflint to lint some terraform code, it is using an earlier version of the super-linter:

> Status: Image is up to date for ghcr.io/github/super-linter:slim-v4.9.2

## Expected behavior

This error should not occur. We haven't done anything to cause it and it does look as though its been caused by the bump to v4.9.3.

admiralAwkbar:
This closes #2954 

- Adding log level to output when we run `ACTIONS_RUNNER_DEBUG=true`
- more?

### DIFF
--- a/lib/functions/detectFiles.sh
+++ b/lib/functions/detectFiles.sh
@@ -514,7 +514,7 @@ function RunAdditionalInstalls() {
     #########################
     BUILD_CMD=$(
       cd "${WORKSPACE_PATH}" || exit 0
-      "tflint --init --loglevel=${TF_LOG_LEVEL}" 2>&1
+      tflint --init --loglevel="${TF_LOG_LEVEL}"  -c "${TERRAFORM_TFLINT_LINTER_RULES}" 2>&1
     )
 
     ##############
--- a/lib/functions/detectFiles.sh
+++ b/lib/functions/detectFiles.sh
@@ -514,7 +514,7 @@ function RunAdditionalInstalls() {
     #########################
     BUILD_CMD=$(
       cd "${WORKSPACE_PATH}" || exit 0
-      tflint --init --loglevel="${TF_LOG_LEVEL}"  -c "${TERRAFORM_TFLINT_LINTER_RULES}" 2>&1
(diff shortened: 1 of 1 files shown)


lib/functions/detectFiles.sh
```diff
@@ -504,12 +504,17 @@ function RunAdditionalInstalls() {
...
+    if [ "${ACTIONS_RUNNER_DEBUG}" = "true" ]; then
+      TF_LOG_LEVEL="debug"
+    fi
     #########################
     # Run the build command #
     #########################
     BUILD_CMD=$(
       cd "${WORKSPACE_PATH}" || exit 0
-      tflint --init 2>&1
+      "tflint --init --loglevel=${TF_LOG_LEVEL}" 2>&1
```
t-botz, line 517: I feel like `tflint --init` will always be a no-op atm. 
`tflint --init` is used to install plugins from the config file and we are not providing a config file.
AWS plugin work because it is manually installed. `tflint --init` will only do something if it has something to download like :
plugin "foo" {
  enabled = true
  version = "0.1.0"
  source  = "github.com/org/tflint-ruleset-foo"
}

To make it work we need to point to the config file I believe:
```suggestion
      tflint --init --loglevel="${TF_LOG_LEVEL}"  -c "${TERRAFORM_TFLINT_LINTER_RULES}" 2>&1
```

PR #2960, opened 2022-05-25
Ensure TFLint plugin directory is correctly set

### LINKED ISSUE #2954: tflint failing in slim-v4.9.3 "Plugin `aws` not found in /github/home/.tflint.d/plugins"

t-botz:
Fixes #2954

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -376,6 +376,11 @@ COPY --from=base_image /home/r-library /home/r-library
 COPY --from=base_image /root/.tflint.d/ /root/.tflint.d/
 COPY --from=python_builder /venvs/ /venvs/
 
+##################################
+# Configure TFLint plugin folder #
+##################################
+ENV TFLINT_PLUGIN_DIR="/root/.tflint.d/plugins"
+
 ####################################################
 # Install Composer after all Libs have been copied #
 ####################################################


ferrarimarco:
It would be interesting to also update our test suite to catch this. @admiralAwkbar WDYT?

ferrarimarco:
And thanks for the PR @t-botz !

t-botz:
@ferrarimarco no worries, I have a pretty good idea on testing not only the bug but also making sure `tflint --init` work as expected.
I can work on that tomorrow, would you prefer me to update this PR or raise it separately?

ferrarimarco:
Updating this PR is ok, thanks!

admiralAwkbar (approved):

krowlandson:
Great to see this fix in code so quickly as I'm also hitting this with the `azurerm` plugin. Thank you!

@admiralAwkbar - do you have this scheduled for a release yet?

Also, @t-botz ... does the implementation of `tflint` work on the assumption that all plugins must be installed via the `.tflint.hcl` config, specifying `source` and `version` in the plugin configuration?

```hcl
plugin "azurerm" {
    enabled = true
    version = "0.16.0"
    source  = "github.com/terraform-linters/tflint-ruleset-azurerm"
}
```

Oddly this seems to work locally in release `v4.9.3` with no errors (as long as I don't specify `source` and `version`), but fails when running in a GitHub Action with or without these.

t-botz:
@krowlandson it works locally and not GitHub because how githhub change the home directory.
I would expect your problem to be fixed not by this PR but by #2959 

By not providing the `source` and `version you will be using the preinstalled from [tflint-bundle](https://github.com/terraform-linters/tflint-bundle)  while if you do, i expect `tflint` to download the plug-in.

PR #3250, opened 2022-08-15
Remove npm-groovy-lint custom handling as it now accepts files as arguments

nvuillam:
npm-groovy-lint now accepts a list of files as arguments (and provides better performances)

To custom handling adding -path and -files is now useless so has been removed


admiralAwkbar (approved):
very nice!

PR #3270, opened 2022-08-22
Add source variables from env file

### LINKED ISSUE #2851: Allow reading github action as local configuration.
Currently it's necessary to duplicate linter rules between action yaml and some local scripts. Having the linter be able to discover configuration settings from a supplied github action would reduce this type of duplication.

BryanQuigley:
This allows to stop requiring duplication of most env variables if you
use something like STRTA - for example:
[test script](https://github.com/azavea/echo-locator/blob/273f543ca734c813c16cda8f3750fcf35f6356a3/scripts/test#L41)
and [ci build](https://github.com/azavea/echo-locator/blob/273f543ca734c813c16cda8f3750fcf35f6356a3/.github/workflows/ci.yml#L43)

Instead you can just put the shared env variables in .github/super-linter.env and super-linter will pull from there by
default (can also override).

Fixes #2851

## Proposed Changes

1. Add check for and sourcing env file 

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ Example usage:
 
 ## Environment variables
 
-The super-linter allows you to pass the following `ENV` variables to be able to trigger different functionality.
+The super-linter allows you to pass the following `ENV` variables to be able to trigger different functionality. You can also specify environment variables in .github/super-linter.env which will be used by both a local super-linter run and from Github actions.
 
 _Note:_ All the `VALIDATE_[LANGUAGE]` variables behave in a very specific way:
 
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -21,6 +21,14 @@ IMAGE="${IMAGE:-standard}"                            # Version of the Super-lin
 LOG_FILE="${LOG_FILE:-super-linter.log}" # Default log file name (located in GITHUB_WORKSPACE folder)
 LOG_LEVEL="${LOG_LEVEL:-VERBOSE}"        # Default log level (VERBOSE, DEBUG, TRACE)
 
+##################################################################
+# Get Environment Variables from env file                        #
+##################################################################
+SUPER_LINTER_ENV_FILE="${SUPER_LINTER_ENV_FILE:-.github/super-linter.env}" # Default env file name
+if test -f "$SUPER_LINTER_ENV_FILE"; then
+  source SUPER_LINTER_ENV_FILE
+fi
+
 if [[ ${ACTIONS_RUNNER_DEBUG} == true ]]; then LOG_LEVEL="DEBUG"; fi
 # Boolean to see trace logs
 LOG_TRACE=$(if [[ ${LOG_LEVEL} == "TRACE" ]]; then echo "true"; fi)


lindluni (dismissed):

lib/linter.sh
```diff
@@ -21,6 +21,14 @@ IMAGE="${IMAGE:-standard}"                            # Version of the Super-lin
 LOG_FILE="${LOG_FILE:-super-linter.log}" # Default log file name (located in GITHUB_WORKSPACE folder)
 LOG_LEVEL="${LOG_LEVEL:-VERBOSE}"        # Default log level (VERBOSE, DEBUG, TRACE)
 
+##################################################################
+# Get Environment Variables from env file                        #
+##################################################################
+SUPER_LINTER_ENV_FILE="${SUPER_LINTER_ENV_FILE:-.github/super-linter.env}" # Default env file name
+if test -f "$SUPER_LINTER_ENV_FILE"; then
```
ferrarimarco, line 28: Instead of `test` use `[` and `]` for consistency with other `if`s

lib/linter.sh
```diff
@@ -21,6 +21,14 @@ IMAGE="${IMAGE:-standard}"                            # Version of the Super-lin
...
+  source SUPER_LINTER_ENV_FILE
```
ferrarimarco, line 29: This should be `source "$SUPER_LINTER_ENV_FILE"`

lib/linter.sh
ferrarimarco, line 28: What if the file exists and is not readable?

lib/linter.sh
ferrarimarco, line 29: Add relevant logging, both info and debug (add more explanation in this case).

ferrarimarco (changes requested):

README.md
```diff
@@ -265,7 +265,7 @@ Example usage:
 
 ## Environment variables
 
-The super-linter allows you to pass the following `ENV` variables to be able to trigger different functionality.
+The super-linter allows you to pass the following `ENV` variables to be able to trigger different functionality. You can also specify environment variables in .github/super-linter.env which will be used by both a local super-linter run and from Github actions.
```
ferrarimarco, line 268: Can you add an example?

lib/linter.sh
```diff
@@ -21,6 +21,14 @@ IMAGE="${IMAGE:-standard}"                            # Version of the Super-lin
 LOG_FILE="${LOG_FILE:-super-linter.log}" # Default log file name (located in GITHUB_WORKSPACE folder)
 LOG_LEVEL="${LOG_LEVEL:-VERBOSE}"        # Default log level (VERBOSE, DEBUG, TRACE)
 
+##################################################################
+# Get Environment Variables from env file                        #
+##################################################################
+SUPER_LINTER_ENV_FILE="${SUPER_LINTER_ENV_FILE:-.github/super-linter.env}" # Default env file name
```
ferrarimarco, line 27: The `SUPER_LINTER_ENV_FILE` environment variable needs to be added to the README in the env vars table.

ferrarimarco (changes requested):
Hi! Thanks for this PR. I've added a few comments to address.

ferrarimarco:
@admiralAwkbar Besides the (small) issues I spotted, I remember we had this discussion already and we didn't go ahead with the approach of having environment variables defined in a file that `super-linter` had to source. I don't remember the specific reasons behind that decision, but I see a couple now:

- There are two ways of configuring super-linter (env vars, env file) which may conflict (which way is the preferred one, if a user configures both? shall we warn the user in that case? If so, this requires some refactoring).
- This is easily accomplishable without super-linter offering any built-in support. Users can source that environment file as part of the command (action, from the shell...) that they use to run super-linter. ([docker run example](https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e---env---env-file)).

BryanQuigley:
@ferrarimarco that works too.  I thought I tried that but I guess I didn't have the syntax right.

I'd like to end this PR and make a new one that just documents how to do this. I've got it working here: https://github.com/azavea/echo-locator/pull/581

Feedback very welcome before I start a new PR.  Like, Is there a better way then `cat .github/super-linter.env >> $GITHUB_ENV`  to do this?

Thanks!
Bryan

BryanQuigley:
Closing this in favor of just a doc change: #3351

PR #3298, opened 2022-09-06
Update npm deps

lindluni:
Attempting to update NPM deps


ferrarimarco:
@lindluni I've cherry picked the asl-validator version fix and put it in #3367.

Also in #3367, I've configured ktlint to avoid checking the filename rule for our tests. We can't change the test file names in the current implementation of our test case logic because we rely on file names to be in the format `lintername_outcome_eventualnumber.extension`. Otherwise, tests will fail.

Dependabot is picking up some of those dependency updates, so I'm going to close this PR. Thanks!

PR #3357, opened 2022-09-26
Fix issues in the build workflow

ferrarimarco:
## Proposed Changes

1. Update GitHub Actions workflow with more accurate labels.
2. Fix the Trivy Scan job failure when the container image build fails.
3. Remove links to non-relevant documentation from README.
4. Don't build the container image when running the Trivy scan job because we run a `fs` vulnerability scan that doesn't need an image to be there.
5. Remove repository visualization workflow and README reference because it's broken (wrong permissions).
6. Move the alpine glibc package key in the repo because the host where it's stored is sometimes down and breaks our builds.
7. Instead of building the image again when publishing a release, pull the `latest` image and add release tags to it. This shortens build times when drafting releases, and avoids failures due to the non fixed dependencies that we have to use. Example: there are a few non-fixed dependencies that we download when building the super-linter container image (search for `latest` in `Dockerfile` if you want to see examples). If we rebuild the image when drafting a release, it might happen that we break the build because those non-fixed dependencies can point to releases that are different from when we built the `latest` image (from our `main` branch).

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
Amazing work!

Issue #3137, opened 2022-07-12
"Failed to call GitHub Status API" with blank errors.


Zee2:
## Describe the bug

Trying to setup `super-linter` and I can't get it to report statuses correctly, and I'm also getting these mysterious status API failures.

```
2022-07-11 23:53:51 [INFO]   ERROR! Failed to call GitHub Status API!
2022-07-11 23:53:51 [INFO]   ERROR:[]
2022-07-11 23:53:51 [INFO]   ERROR! Failed to call GitHub Status API!
2022-07-11 23:53:51 [INFO]   ERROR:[]
2022-07-11 23:53:52 [INFO]   ERROR! Failed to call GitHub Status API!
2022-07-11 23:53:52 [INFO]   ERROR:[]
2022-07-11 23:53:52 [INFO]   ERROR! Failed to call GitHub Status API!
2022-07-11 23:53:52 [INFO]   ERROR:[]
2022-07-11 23:53:52 [INFO]   ERROR! Failed to call GitHub Status API!
2022-07-11 23:53:52 [INFO]   ERROR:[]
```

I am passing in the token correctly with `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}`.

## Expected behavior

The status is set correctly and no Status API errors appear.

## Steps to Reproduce

My workflow file looks like:

```
name: Lint
on:
  push:
    branches: ["foobar"]
  pull_request:
    branches: ["foobar"]

jobs:
  lint:
    runs-on: ubuntu-latest

    steps:
    - name: Checkout
      uses: actions/checkout@v3
      with:
        # Full git history is needed to get a proper list of changed files within `super-linter`
        fetch-depth: 0
    
    - name: Lint
      uses: github/super-linter@v4
      env:
        VALIDATE_ALL_CODEBASE: false
        DEFAULT_BRANCH: foobar
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

## Additional context

Possibly related to #556?

Zee2:
Ah, same as #2641, too!

Zee2:
This is still an issue for us :(

Zee2:
Still an issue....

sid-am-ahd935:
Any workarounds for this??

ferrarimarco:
Hi! To troubleshoot this issue, we need to be able to reproduce it. Can you please provide an example of a source file that triggers this?

eitsupi:
[...]

PR #3367, opened 2022-09-28
Build process improvements

ferrarimarco:
## Proposed Changes

1. Don't install `actionlint` using the installation script because we already do that by copying it from its container image.
2. Install the `py3-pyflakes` package (+ relevant test) because it's needed by `actionlint`.
3. Added a missing test to verify that we install the `ca-certificates` package.
4. Fix a typo in the `.sqlfluff` file path in the InSpec test control.
5. Cherry-picked the version check for `asl-validator` from #3298
6. Configured ktlint to not expect test cases files to require PascalCase naming so that we don't have to change our test logic.
7. Update `asl-validator` to `3.0.8` by merging with #3310. This ensures that the fix we cherry-picked works and that we can update `asl-validator`. Otherwise, we're stuck in a chicken-and-egg problem. Dependabot will take care of closing #3310 after we merge this one.
8. Show errors when using curl to call the status API, otherwise we can't show anything if that call fails. Useful to investigate #3137.
9. Make the build more reproducible by explicitly setting versions for checkstyle, google-java-format, and ktlint. This should also make the build more robust by not requiring additional calls to the GitHub APIs to get the `latest` release download URL.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -28,3 +28,8 @@ spaces_around_brackets = false
 [{Makefile,go.mod,go.sum,*.go,.gitmodules}]
 indent_style = tab
 indent_size = 4
+
+[**/test/*.{kt,kts}]
+# Disable the filename rule because our test case logic requires that files used in test cases
+# follow a certain syntax which is not compatible with what ktlint currently mandates (PascalCase)
+disabled_rules = filename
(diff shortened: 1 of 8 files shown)


admiralAwkbar (approved):

admiralAwkbar:
@ferrarimarco Your always doing the great work!

ferrarimarco:
Thanks mate!

PR #3373, opened 2022-09-29
Remove non-working cauldron report

ferrarimarco:
## Proposed Changes

1. Remove the cauldron.io report because it's not working and I we don't have administrative access to that.

### Author/Contributor

### Reviewing Maintainer


admiralAwkbar (approved):
It was cool while it lasted :)

PR #3405, opened 2022-10-03
Update NPM Deps

lindluni:
Update NPM deps test


lindluni:
Just waiting on the `12.0.1` release of ts-standard that picks up my fixes to the their `--version` flag and then we should be able to finally update all of the ts-standard and eslint dependencies we've been holding back for months.

lindluni:
Need this upstream change to fix a regression in the `ts-standard` rewrite: https://github.com/standard/ts-standard/pull/269/files

PR #3500, opened 2022-10-27
templates: ignore version pinning from hadolint

gustavovalverde:
Fixes #

## Proposed Changes

It's very common not to pin the specific package inside the Dockerfiles, stopping a merge on this scenarios forces project to keep updating their Dockerfiles more than they need to.

Search for "pin version" in https://github.com/hadolint/hadolint to confirm the rules being disabled

### Author/Contributor

### Reviewing Maintainer


gustavovalverde:
Can we reopen this?

Hanse00:
@zkoppert and I had a brief discussion about this. As much as we believe you are correct Gustavo, that it is a common practice not to pin specific dependencies, it is also the less safe of the options. We believe this project should provide safe defaults, with the option to override them to better suit your specific scenario. We'd like to close this PR given that it modifies the defaults we ship to all of our users, assuming there are no objections from @lindluni.

PR #3502, opened 2022-10-28
JakubCzarlinski/add-clang-format-style-options

JakubCzarlinski:
Add functionality for clang-style options.

Usage has been demonstrated in the commits of this PR, but effectively reverted to keep the functionality constant.
See https://clang.llvm.org/docs/ClangFormatStyleOptions.html for reference.

## Proposed Changes

1. Enable the use of `.clang-format` in `.github/linters/`.

### Author/Contributor

### Reviewing Maintainer


.github/linters/.clang-format
```diff
@@ -0,0 +1,3 @@
+---
```
ferrarimarco, line 1: We don't need this file because it's identical to the default (`TEMPLATES/.clang-format`)

lib/linter.sh
```diff
@@ -892,7 +894,7 @@ LINTER_COMMANDS_ARRAY['ANSIBLE']="ansible-lint -c ${ANSIBLE_LINTER_RULES}"
 LINTER_COMMANDS_ARRAY['ARM']="Import-Module ${ARM_TTK_PSD1} ; \${config} = \$(Import-PowerShellDataFile -Path ${ARM_LINTER_RULES}) ; Test-AzTemplate @config -TemplatePath"
 LINTER_COMMANDS_ARRAY['BASH']="shellcheck --color --external-sources"
 LINTER_COMMANDS_ARRAY['BASH_EXEC']="bash-exec"
-LINTER_COMMANDS_ARRAY['CLANG_FORMAT']="clang-format --Werror --dry-run"
+LINTER_COMMANDS_ARRAY['CLANG_FORMAT']="clang-format --Werror --dry-run -style=file"
```
ferrarimarco, line 897: If I read the clang-format documentation correctly, you need to pass the path to the `.clang-format` file here.

Auto detection will not work because we're not sure that the `.clang-format` file that we provide is in one of the parent directories of the files to lint.

ferrarimarco (changes requested):
Thanks for this PR! A few things to refactor :)

TEMPLATES/.clang-format
```diff
@@ -0,0 +1,3 @@
+---
```
ferrarimarco, line 1: Can you also add .clang-format as a special case [here](https://github.com/github/super-linter/blob/9bcd24302089a9254ba77ec8b85079de7db10949/lib/functions/buildFileList.sh#L771), so it gets properly recognized as a YAML file?

In the current implementation, it's not recognized as a YAML file. Example [build](https://github.com/github/super-linter/actions/runs/3847772341/jobs/6554657319#step:10:2177) output:

```
2023-01-05 15:06:35 [DEBUG]   File:[/tmp/lint/.github/linters/.clang-format], File_type:[clang-format], Base_file:[.clang-format]
2023-01-05 15:06:35 [DEBUG]   Failed to get filetype for:[/tmp/lint/.github/linters/.clang-format]!
```

ferrarimarco (changes requested):

PR #3556, opened 2022-11-18
Use git check-ignore instead of a (incomplete) list of ignored files

### LINKED ISSUE #3555: git-ignored files are not always ignored
## Describe the bug

The IGNORE_GITIGNORED_FILES switch does not really work in all cases. If you ignore a folder, files in that folder are still linted. For local usage, this might include coverage reports, test logs, etc., massively increasing the noise and linting time.

## Expected behavior

All files ignored should be ignored by the linter if the option is set.

## Steps to Reproduce

```
#!/bin/bash

wd=$(mktemp -d)

pushd "$wd" || exit 1

git init

# setup super linter environment
cat > super-linter.env <<EOF
IGNORE_GITIGNORED_FILES=true
LOG_LEVEL=NOTICE
VALIDATE_PYTHON=true
EOF

# prepare a broken python file
mkdir foo/

cat > foo/error.py <<EOF
def syntax_err:
  pas
EOF

# git ignore the problematic file
echo foo/ > .gitignore

# run check, expectation is: "no error"
podman run -e RUN_LOCAL=true -e USE_FIND_ALGORITHM=true \
       --env-file "super-linter.env" -v ./:/tmp/lint:Z \
       -it github/super-linter:latest
```

choeger:
The list-based method breaks down in case of ignored folders: If .gitignore contains foo/, foo/bar.py is still checked.

Fixes #3555

## Proposed Changes

1. Do not try to list all ignored files
2. Use git cjeck-ignore instead on every found file

### Author/Contributor

### Reviewing Maintainer


lindluni:
😲

PR #3607, opened 2022-12-08
replace kubeval with kubeconform

### LINKED ISSUE #2460: Support kubeconform

kubeval is mostly outdated and rarely updated, it would be neat to support kubeconform

https://github.com/yannh/kubeconform

### LINKED ISSUE #3496: Replace kubeval with kubeconform
`kubeval` is no longer supported.

Replace `kubeval` with `kubeconform` per recommendation in `kubeval`'s [README](https://github.com/instrumenta/kubeval#kubeval). I have been using `kubeconform` in place of `kubeval` for quite some time and it works well.

There is also [`kube-linter`](https://github.com/stackrox/kube-linter), but it has slightly more opinionated linting rules than `kubeval`/`kubeconform`.

**Additional context**
This has been requested before: https://github.com/github/super-linter/issues/2460

chenrui333:
Closes #3496
Closes #2460

replacing kubeval with kubeconform as kubeval is no longer maintained.

- relates to https://github.com/instrumenta/kubeval/issues/268


Dockerfile
```diff
@@ -49,8 +49,8 @@ ARG KTLINT_VERSION='0.47.1'
 ARG PSSA_VERSION='latest'
 ARG PWSH_DIRECTORY='/usr/lib/microsoft/powershell'
 ARG PWSH_VERSION='latest'
-# Kubeval Version
-ARG KUBEVAL_VERSION='v0.16.1'
+# Kubeconform Version
+ARG KUBECONFORM_VERSION='v0.5.0'
```
ferrarimarco, line 53: Given that Kubeconform has a container image available, can you please refactor this PR to use that? By doing that, we can leverage Dependabot for updates. Thanks!

ferrarimarco (changes requested):

Dockerfile
chenrui333: is there any documentation specifying which is preferred? I personally would vote for binary though.

ferrarimarco: Basically, if a container image is available, we try using that so that Dependabot takes care of updating that. As you can see, in the current implementation, the Kubeconform image would need to be manually updated.

lindluni:
Added in #3800

PR #3609, opened 2022-12-11
Textlint should check every file

jvoigtlaender:
## Proposed Change

Textlint is not anomore applying only to `.md` files.

Instead, selection of files is left to `textlint` itself, potentially influenced by settings people make in `.textlintrc`.

I don't think additional documentation is needed, because the `textlint` documentation itself explains how to control which files are checked and how (https://github.com/textlint/textlint#supported-file-formats).

### Author/Contributor

### Reviewing Maintainer


ferrarimarco (changes requested):
This might need some more work because if you don't install the relevant textlint plugin, it will not lint them. I think it might be worth installing the plugins we need and explicitly add the files to the relevant lists as we're doing now.

For example, we could install the plugins mentioned [here](https://github.com/textlint/textlint#supported-file-formats) and add the files to the lists as needed.

jvoigtlaender:
Installing additional `textlint` plugins would probably be a good idea, but isn't that a separate issue? 

The documentation you linked to states that
> textlint supports Markdown and plain text by default

More specifically, https://github.com/github/super-linter/blob/aad87c569b8a3de467bf6504c06127822e2845aa/TEMPLATES/.textlintrc indicates that currently only the [`terminology`](https://github.com/sapegin/textlint-rule-terminology) rules set of `textlint` is used in `super-linter` at all. That rules set is in no way specific to Markdown. It would just as well apply to `.txt` files, `.sbv` files etc.

So, there seems to be no reason for the preferential treatment of `.md` files as of current. All "plain text" files (as per the above quotation) should be able to get the exact same treatment and benefit from it no less than `.md` files do.

ferrarimarco:
I think we shouldn't lint every file with Textlint. That would make the whole linting process slower because the way super-linter is designed at the moment (one linter execution invocation per file). I think we should rather add more file types, along with any missing textlint plugin. For example, we could have a new category (`NATURAL_LANGUAGE`) and all the relevant files there.

Example: `.md, .txt, .sbv`....

jvoigtlaender:
That sounds good.

But, maybe just for context:

My understanding is that from the user side, for plain text files, adding something like this: 
```
    "plugins": {
        "@textlint/text": {
            "extensions": [".sbv"]
        }
    }
```
to one's `.textlintrc` would be enough. That is, `super-linter` would not even have to somehow additionally install a plugin, because `textlint/text` is already builtin anyway/always. All that the user (of the `super-linter` GitHub action) would have to do is to indicate like above that `.sbv` files are text files. That's how I read https://github.com/textlint/textlint/tree/master/packages/%40textlint/textlint-plugin-text#usage and https://github.com/textlint/textlint/tree/master/packages/%40textlint/textlint-plugin-text#options

However, what keeps this from actually having the desired effect currently is that the `.sbv` files do not even end up in the list of files that `textlint` is called on. My attempt in this PR to address this was modelled on these lines for the `editorconfig-checker`: https://github.com/github/super-linter/blob/c26ca9fb2cce9f169fcb81077bb93017d7fe8fe5/lib/functions/buildFileList.sh#L312-L313 Actually, this same pattern appears to be followed for various linters: https://github.com/github/super-linter/blob/c26ca9fb2cce9f169fcb81077bb93017d7fe8fe5/lib/functions/buildFileList.sh#L318-L319 https://github.com/github/super-linter/blob/c26ca9fb2cce9f169fcb81077bb93017d7fe8fe5/lib/functions/buildFileList.sh#L324-L325

Do the efficiency downsides that you see for the suggested change, running `textlint` on too many files, not apply to the `editorconfig`, `jscpd`, `gitleaks` linters?

On the other hand, the apparent downside of 
> For example, we could have a new category (`NATURAL_LANGUAGE`) and all the relevant files there.

is that somebody would have to explicitly list all the file types that are to be considered as plain text. Who makes that list inside `super-linter`, and what can a user do if their case is not included?

ferrarimarco:
> is that somebody would have to explicitly list all the file types that are to be considered as plain text.

That's how super-linter works at the moment. You got this right.

> Who makes that list inside super-linter, and what can a user do if their case is not included?

The only way we have now is to open a PR or an issue requesting that.

jvoigtlaender:
> That's how super-linter works at the moment.

Except when it comes to `editorconfig`, `jscpd`, `gitleaks`, it seems. :smile:

But I'm not trying to detract you from the course you outlined. It's much appreciated!

I'm definitely fine doing pull requests for individual file types when needed. My pull request here attempted to avoid that, but if that's too quick-and-dirty, I of course defer to your judgment.

ferrarimarco:
You have a point!

Do you know what happens if you feed textlint with binary files? Also, linting large code bases with the linters you mentioned takes a lot of time.

@lindluni what's your take here?

lindluni:
I'm going to merge this and see if we get any comments from people who use the `latest` branch.

jvoigtlaender:
Cool, Thanks!

PR #2808, opened 2022-04-20
chore: 🤖 stylelint support less & prettier

hzz780:
Fixes #

## Proposed Changes

1. stylelint support postcss-less. 
2. stylelint support prettier.
3. ...

I found that stylelint don't support less & prettier yet.  It only support css standard, scss, scss standard.

Styleint have suggest prettier for css and postcss-less for less in the [GET START](https://stylelint.io/user-guide/get-started) of stylelint.

### Author/Contributor
- [ ] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [ ] Label as `breaking` if this is a large fundamental change
- [x] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #3646, opened 2022-12-20
Stylelint support prettier plugin

maxemann96:
Fixes an issue using stylelint with prettier plugin.
Can be seen as follow up to #2808. The original PR does not include the prettier plugin, which is needed to let stylelint function properly with prettier. See [stylelint prettier installation guide](https://github.com/prettier/stylelint-prettier#installation).

## Proposed Changes
Fixed an error while using stylelint with prettier plugin.

### Author/Contributor

### Reviewing Maintainer


maxemann96:
Rebased the PR.

PR #3689, opened 2022-12-23
Simplify automation workflows

lindluni:
This PR is a full rework of our CICD practices.

1. Splits the CI portions that run on Pull Request as a separate workflow from CD where we publish images on merge to main. This make it easier to understand what is happening when.
2. Cleaned up the names of each step to make them visible by making them shorter to prevent being cutoff in the UI, but also to make them more descriptive of what the step was doing.
3. Uses the larger-hosted runner which helps with the time it takes to build the image, but also with the number of times we encounter the anonymous IP rate limit throttles on GitHub due to using unauthenticated calls to GitHub.
4. Simplified the Matrix strategy to only need variations that are required, reducing the number of inputs.
5. Disabled the `PSMissingModuleManifestField` Powershell linting rule for our repo as it is erroneously assuming our PWSH linter templates are actual Powershell modules as we used poor naming schemes for the Powershell linter rules filename.


.github/linters/actionlint.yaml
```diff
@@ -0,0 +1,3 @@
+self-hosted-runner:
+  labels:
+    - ubuntu-20.04-8core-sl
```
ferrarimarco, line 3: Maybe switch to Ubuntu 22.04 if available?

If you go this route, check the rest of the jobs as well for other 20.04 instances.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,100 @@
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Login to GHCR
+        uses: docker/login-action@v2.1.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Start  ${{ matrix.images.environment }} Deployment
```
ferrarimarco, line 51: Remove extra space after `Start`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Build Image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: ./Dockerfile
+          build-args: |
+            BUILD_DATE=${{ env.BUILD_DATE }}
+            BUILD_REVISION=${{ github.sha }}
+            BUILD_VERSION=${{ github.sha }}
+          load: false
```
ferrarimarco, line 43: I don't know if this is intended, but by setting this to false you are forcing the steps below (`Run Super-Linter tests` and `Lint Entire Codebase`) to pull the image from the registry. If you want to reuse the image you just built instead, you need to split this in two steps (one that loads, one that pushes) because the two settings cannot be set to `true` at the same time, IIRC.

In the current implementation, we also test if we're able to pull the standard image from the registry at least. We don't pull the slim one though.

.github/workflows/release.yml
```diff
@@ -105,13 +105,13 @@ jobs:
       # Then, we use that reference to create new tags, so that the new tags point to the commit
       # the original tag was pointing to, and not to the original tag.
       # This notation is documented at https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emltrevgtemegemv0998em
-      - name: Update the major version and latest tags
+      - name: Update Major Version and Latest Git Tag
```
ferrarimarco, line 108: Is there a way to automatically bump the version in `action.yml` as well? I see that there are actions that do this, but I didn't try any.

.github/workflows/stack-linter.yml
```diff
@@ -17,6 +17,8 @@ name: Stack Linter
 #############################
```
ferrarimarco, line 17: Isn't this workflow redundant at this point?

ferrarimarco (changes requested):
A few small nits 😄

.github/workflows/stack-linter.yml
lindluni: Probably, my initial run through was just splitting up what we had, I'll drop this since it runs as part of the other workflow.

.github/workflows/release.yml
lindluni: Thats my top priority next. Having released last night and gone through that nightmare, we have to sort out a truly automated release process. I have some great plans, but I'm also waiting on GitHub to release some new features. We have some MAJOR changes coming to Actions maintenance and security this quarter that will probably render some of our workflows pointless afterwards.

.github/linters/actionlint.yaml
lindluni: Right now we only have 20.04 available, this was custom built for Super-Linter, I have asked to get the new larger runners for us, but they asked if we could wait for the new runners to be widely available. So once they go GA we will swap to the ubuntu-22.04-8core-sl runners.

.github/workflows/ci.yml
lindluni: Good catch, bad copy-paste on my part. Fixed.

Do you think we need to test the pull? Can certainly add that, but I would expect the push to fail in the first place if there was going to be an issue with the registry, or is there another scenario you had in mind that caused it to be introduced before.

.github/workflows/release.yml
ferrarimarco: Nice!

.github/workflows/ci.yml
ferrarimarco: SGTM!

ferrarimarco (approved):

ferrarimarco:
We might want to disable the `Stack linter` required check on the `main` branch protection rule.

ferrarimarco:
(I don't have the right privileges to do that)

PR #3728, opened 2022-12-30
Push scripts into standalone files and authenticate requests

lindluni:
Pushes inline scripts in the Dockerfile into standalone scripts and authenticates requests to GitHub using a the GITHUB_TOKEN to reduce build flakiness due to GitHub's abuse and rate limiting due to unauthenticated requests.

The GITHUB_TOKEN is injected via a docker secret. The `docker` directive has been added to the Makefile which validates the user has properly set the GITHUB_TOKEN environment variable on their local machine before building.

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -57,7 +57,7 @@ jobs:
           load: true
           push: false
           secrets: |
-            GITHUB_TOKEN=GITHUB_TOKEN${{ secrets.GITHUB_TOKEN }}
+            GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
           tags: |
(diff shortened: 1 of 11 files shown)


ferrarimarco (approved):
Quick question: what is the rationale behind moving stuff in scripts? Modularization maybe?

lindluni:
I had two rationale, one was by not directly embedding the GITHUB_TOKEN variables in the Dockerfile, the token isn't logged and derisks potentially leaking it if the Actions mask misses it for some reason. It also prevents the content of GITHUB_TOKEN from being leaked into the `docker history` command by mistake. I wanted to do everything I could to make sure the token wasn't accidentally leaked as part of the build or command history process, even if it was excessive.

And the second was to make it easier to run locally. Whenever I am debugging stuff I end up copying those scripts out of the dockerfile and into a standalone script that I can run in a container myself. Like this it just makes it easier to debug.

PR #3791, opened 2023-01-06
Don't ignore dotfiles when running eslint

### LINKED ISSUE #3574: Files beginning with . excluded from JSON linting
## Describe the bug

Files beginning with `.` are excluded from JSON linting. This also applies to JSON files in folders beginning with `.`.

## Expected behavior

Ignore settings passed through the variables should be respected, as well as potentially the ESLint settings. It appears that JSON linting is using settings from the tool used, that cannot be adjusted through Super Linter.

## Steps to Reproduce

1. Create a repo with a JSON file beginning with `.`, e.g. `.test.json`.
2. Scan the repo with Super Linter, including JSON linting.
3. You will see the warning
> /github/workspace/.test.json
> Warning:   0:0  warning  File ignored by default.  Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override

## Logs

```
/github/workspace/.test.json
Warning:   0:0  warning  File ignored by default.  Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override
```

ferrarimarco:
Fixes #3574

## Proposed Changes

1. Don't ignore dotfiles when running eslint, but ignore anything in the `node_modules` directory.
2. Added a couple of test cases for the previous point.
3. Remove the `.eslintignore` file from our repository to centralize the ignored files configuration in `.eslintrc.yml`.
4. Remove `.github/linters/.eslintrc.yml` because we don't override any property vs the default (`TEMPLATES/.eslintrc.yml`), so we can use that one directly.

### Author/Contributor

### Reviewing Maintainer


TEMPLATES/.eslintrc.yml
```diff
@@ -20,6 +20,10 @@ globals:
   Atomics: readonly
   SharedArrayBuffer: readonly
 
+ignorePatterns:
+  - "!.*"
```
muiriswoulfe, line 24: The convention in this file seems to be to use single rather than double quotes, so you might wish to align with that.

muiriswoulfe (approved):
Thanks for getting this out!

TEMPLATES/.eslintrc.yml
ferrarimarco: From what I can see, globs in this file use double quotes, so I guess no one cared to check :)

muiriswoulfe: I saw the single quotes and didn't look further. Didn't realise there was a convention like that in place.

ferrarimarco:
@muiriswoulfe No worries :)

As you can see from the [build logs](https://github.com/github/super-linter/actions/runs/3856587476/jobs/6572972045#step:9:9985), dotfiles are now picked up.

TEMPLATES/.eslintrc.yml
ferrarimarco: I think we should enable the [quoted-strings](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.quoted_strings) rule in this repo at one point :)

lindluni (approved):

PR #3812, opened 2023-01-11
Add terraform_fmt to inspec tests

lindluni:
While calling `terraform --version` or `terraform fmt --version` produce the same output, I'd like to be explicit and call `terraform fmt --version` to monitor for changed behavior using this test.

### DIFF
--- a/test/inspec/super-linter/controls/super_linter.rb
+++ b/test/inspec/super-linter/controls/super_linter.rb
@@ -149,6 +149,7 @@ control "super-linter-installed-commands" do
     { linter_name: "standard"},
     { linter_name: "stylelint"},
     { linter_name: "tekton-lint"},
+    { linter_name: "terraform_fmt", version_command: "terraform fmt --version"}
     { linter_name: "terragrunt"},
     { linter_name: "terrascan", version_option: "version"},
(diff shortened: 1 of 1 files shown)


ferrarimarco (approved):

lindluni:
Missed a comma 🤦🏻

ferrarimarco:
Perhaps I'm misunderstanding what you're trying to do, but `terraform fmt` doesn't seem to accept the `version` option. [source](https://developer.hashicorp.com/terraform/cli/commands/fmt).

ferrarimarco (changes requested):

lindluni:
> Perhaps I'm misunderstanding what you're trying to do, but `terraform fmt` doesn't seem to accept the `version` option. [source](https://developer.hashicorp.com/terraform/cli/commands/fmt).

Yea, I see that there, what's interesting is it does on my Mac on the latest version available in Homebrew, but it looks like its part of the global flag list rather than the subcommand list. I'm just going to remove the "fmt" part.

lindluni:
Have to dig into why it couldn't find the "terraform" command at all

ferrarimarco (approved):

PR #3905, opened 2023-02-08
Docker: Use image as input instead of hard-coded value

mihaigalos:
The current form of the `super-linter` contains a hard-coded docker image. We are using an internally hosted GitHub Enterprise and Actions.

These actions cannot connect to the internet to pull the image.

This PR allows users to specify the Docker image as an input and point it to their own container registry.

## Proposed Changes

1. Remove hard-coded Docker image.
2. Add image as an input parameter.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,6 +519,8 @@ echo "@generated" # @not-generated
 
 The **Docker** container that is built from this repository is located at [github/super-linter](https://hub.docker.com/r/github/super-linter)
 
+If you would like to use your own container registry instead, please set the `image` input attribute when calling the action.
+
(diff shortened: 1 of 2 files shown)


mihaigalos:
Seems some CI jobs are failing, I can see they're red on other PRs as well. Waiting for a fix here.

mihaigalos:
One could use a different container engine than `dockerd`, for example `cri-o` which supports mirror aliases, even if the image name is fully qualified.

That renders this PR superfluous. Closing.

PR #4029, opened 2023-03-28
Build for both AMD and ARM

### LINKED ISSUE #4028: Provide an arm64 docker image
The docker image is build for amd64.
When I try to run super-linter locally on my new mac (M1), docker fails to execute all the linters and fails.

Docker supports multi-architecture images and the same image can be build to excute natively on multiple processors.

Please add support for arm61 to enable running on M1/M2/… based macs

Here is the kind of error that running an arm64 image on amd64 produces:
```
------
2023-03-30 03:30:11 [ERROR]   Found errors in [textlint] linter!
2023-03-30 03:30:11 [ERROR]   Error code: 133. Command output:
------
assertion failed [result.value != EEXIST]: VmTracker attempted to allocate existing mapping
(ThreadContextVm.cpp:47 mmap)
```

alexnederlof:
Change CD so it build both AMD and ARM images. ARM images will make using the linter significantly faster on M1/M2 macs. 

It's currently so slow on a Mac that it's faster to push and run remotely. 

Fixes   #4028

### Author/Contributor

### Reviewing Maintainer


alexnederlof:
@lindluni  @zkoppert @nemchik  @Hanse00 @GaboFDC @ferrarimarco Does anyone have time for a quick one liner review on this one to increase the experience on M1 Mac significantly? 😄

lindluni (approved):

lindluni:
Thank you @alexnederlof and apologies for the delay, we had to work through removing a deprecated package in NPM that was breaking all of our workflows and tests

lindluni:
Hey @alexnederlof, that is the reason it ended up reverted. I'll try to find some time to take a deeper look this week to see if there is a supported driver that will allow us to build the image without any issues, unless you find something sooner.

Thanks for any help on this, and for raising this request in the first place.

BretFisher:
I can help. I teach these GitHub Actions. AMA 😉  

Full [example with lighting up everything for the docker build](https://github.com/BretFisher/github-actions-templates/blob/main/.github/workflows/docker-full.yaml).

Specifically, you need to run the other Docker Actions [qemu and buildx](https://github.com/BretFisher/github-actions-templates/blob/ba4bdd27f401830c489b0b199b376898959bf0e0/.github/workflows/docker-full.yaml#L26-L30) first. Enabling QEMU will allow the x86_64 public runners to build most arm64 things in Docker. Enabling Buildx will create a BuildKit environment that supports multi-arch and caching.

Adding these two steps anywhere before the build-push-action should solve it:

```yaml
      - name: Set up QEMU
        uses: docker/setup-qemu-action@v2

      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v2
```

After that, I recommend enabling [GitHub Cache for Docker's build](https://docs.docker.com/build/ci/github-actions/cache/#github-cache) layers by [adding these two lines to the build-push-action step](https://github.com/BretFisher/github-actions-templates/blob/ba4bdd27f401830c489b0b199b376898959bf0e0/.github/workflows/docker-full.yaml#L62-L63) and it should speed up future builds.

```yaml
          cache-from: type=gha
          cache-to: type=gha,mode=max
```

BretFisher:
LMK if you'd like me to create the PR @alexnederlof

alexnederlof:
That would be fantastic, thanks @BretFisher !

BretFisher:
@alexnederlof Did you try to build the Dockerfile on arm64? It's rare that a complex Dockerfile "just works" across platforms, and it looks like `main` branch fails on `install-ktlint.sh` because it trys to use a x86_64 library: `/lib/ld-musl-x86_64.so.1`

I'm going to try some things, but curious if others had success.

BretFisher:
Ah, OK, yea this is a much bigger PR from what I can tell (which is what I was saying in https://github.com/github/super-linter/discussions/1956#discussioncomment-2347889).  The current build assumes x86_64. An arm64 build shouldn't assume the docker build server or the Actions runner has QEMU/Rosetta enabled (to emulate x86_64 on arm64). *Every* tool used or built needs to be for arm64.

### Things that would need to be included in an arm64 PR:

- **Change the Dockerfile install style for many of the tools.** The Dockerfile pulls lots of binary releases via the easy "copy a file from another docker image" but many of those images don't have an arm64 variant. For example, the [alpine/terragrunt](https://hub.docker.com/r/alpine/terragrunt/tags) image, which has had a [request for over a year](https://github.com/alpine-docker/terragrunt/issues/21) to build an arm64 image. In this case of terragrunt, we'd need to move to curling [upstream binaries](https://github.com/gruntwork-io/terragrunt/releases/tag/v0.45.2), which **does** have arm64. This is an industry-wide issue that some of us have been working on, but it requires PRs (and often troubleshooting) across a wide range of repos. We'll likely find some tools don't even have a linux/arm64 build yet.
- **Support runtime architecture downloads in scripts.** The installers in `/scripts` are also assuming x86_64, with some being hardcoded and others not dynamically choosing the arch when downloading.
- **Update CI and CD for multi-platform.** Adding `docker/setup-buildx-action` and `platforms: linux/amd64,linux/arm64` to docker builds in both `ci.yml` and `cd.yml`

alexnederlof:
I was half way and then had to stop for other reasons. I guess I stopped before I hit the issues. Sorry to hear you ran into a bunch these issues, I was hoping it would be easy.

PR #4109, opened 2023-04-22
parametrize the install scripts

denisa:
As preparation for multi-architecture docker image with support for amd64 and arm64 architecture, this PR parametrizes every places that made an explicit reference to the x86/amd processor.

See https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

Fixes #

## Proposed Changes

1. use TARGETARCH instead of hardcoded  values
2. where necessary, convert the TARGETARCH value (amd64) to the desired hardcoded value
3. split the various install scripts so that each script handle one linter
4. reduce the `install-bash-exec` to simply copying the script, instead of running an installer to generate the file
5. inline `install-raku.sh` as it ended up being a one-liner

### Author/Contributor

### Reviewing Maintainer


denisa:
Question:
As I was going through the scripts, I saw that a few scripts install unrelated software.
Namely, `install-ktlint.sh` installs *ktlint* and *dart-sdk*, and `install-raku.sh` installs *raku*, *checkstyle*, *google-java-format* and *lua*.

Would it be ok to split them out, so that every tool gets its own install?

lindluni:
> Question: As I was going through the scripts, I saw that a few scripts install unrelated software. Namely, `install-ktlint.sh` installs _ktlint_ and _dart-sdk_, and `install-raku.sh` installs _raku_, _checkstyle_, _google-java-format_ and _lua_.
> 
> Would it be ok to split them out, so that every tool gets its own install?

Would love to see that, when those were originally split out into scripts (they were originally embedded in the Docker image) they were copied as is. Would love to see that cleaned up.

denisa:
I’ll add this refactor to the PR asap

denisa:
@lindluni Could I get a review on this?

Converting to a multi-arch image — my ultimate goal — is going to be long running effort and I’d rather have many small PS, to avoid drift.

lindluni (approved):

lindluni:
@denisa 

Thank you! And now that you've contributed your Pull Request should automatically trigger CI, you wont have to wait for approvals anymore

PR #4225, opened 2023-05-22
Removes dependency workflows

### LINKED ISSUE #2579: Drop custom script and workflow to update Python dependencies and use Dependabot
Currently, Python dependencies are handled by a [workflow](https://github.com/github/super-linter/blob/main/.github/workflows/deps-python.yml) and a [script](https://github.com/github/super-linter/blob/main/.github/scripts/update-python.sh).

Why don't we drop the custom workflow and script?

PS: I saw there are a script and a workflow to update NPM dependencies as well, but I didn't had a look to those yet.

/cc @lindluni who worked on this implementation IIRC
/cc @admiralAwkbar 

EDIT: updated after realizing that Dependabot handles all the requirements files in /dependencies/python. Example: https://github.com/github/super-linter/network/updates/314377296

Hanse00:
We remove the weekly GH Action workflows for dependency updates, given that this function is handled by Dependabot.

Fixes #2579

## Proposed Changes

1. Remove the Python and NPM dependency workflows

### Author/Contributor

### Reviewing Maintainer


zkoppert (approved):
I agree these are redundant and we can remove.

PR #4274, opened 2023-05-24
Fixes bare URL in Code of Conduct

Hanse00:
An update to markdownlint in #4158 has highlighted that the current file breaks one of the rules, namely: https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md

This commit resolves the issue by adding angle brackets around the URL.

## Proposed Changes

1. Add angle brackets around bare URL in code of conduct.

### Author/Contributor

### Reviewing Maintainer


zkoppert (approved):
🚀

PR #4320, opened 2023-05-26
Removes docker hub tag

Hanse00:
Per the output of https://github.com/super-linter/super-linter/actions/runs/5085678972/jobs/9139378583 it is not a valid option to comment out the tag in this context. Instead this PR removes it entirely for the time being. The Git history will always let us recover the tag later.

### Author/Contributor

### Reviewing Maintainer


zkoppert:
Looks like this line got removed somewhere else along the way. Closing as completed.

PR #4348, opened 2023-05-30
git check-ignore is run on every file

sabbene:
Fixes #

## Proposed Changes

1. By checking that `IGNORE_GITIGNORED_FILES` is set to true before running `git check-ignore "$FILE"` we can prevent the git check-ignore command from being run unnecessarily.   This has the secondary benefit a not causing the following error message to be printed once per linted file outside of a git project: `fatal not a git repository (or any of the parent directories) ...`

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/functions/buildFileList.sh
+++ b/lib/functions/buildFileList.sh
@@ -304,7 +304,7 @@ function BuildFileList() {
     ###################################################
     # Filter files if FILTER_REGEX_EXCLUDE is not set #
     ###################################################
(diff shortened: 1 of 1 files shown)


zkoppert (approved):
🚀

pemaxim:
`github/super-linter:latest` still does not have this fix.

ferrarimarco:
@pemaxim I think the fix landed in latest, so it might be something else. Can you open a new issue? Thanks!

sabbene:
I had to cut over from using the github release to using ghcr.io/super-linter/super-linter:latest to pick up these changes.

PR #4362, opened 2023-06-10
Use default `$GITHUB_SERVER_URL` for status URL creation and remove `$GITHUB_DOMAIN`

Shegox:
This pull request aims to improve the way we create the status URL by utilizing the default environment variable `$GITHUB_SERVER_URL` provided by GitHub Actions. This change will allow for seamless integration with GitHub Enterprise Server without the need to manually set the `GITHUB_DOMAIN`.

According to GitHub's documentation, the recommended syntax for constructing a workflow run URL is as follows:

```
$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
```

## Proposed Changes

1. Update the status URL creation method to always use the GitHub-provided `$GITHUB_SERVER_URL`.
2. Remove the `$GITHUB_DOMAIN` configuration environment variable as it will no longer be needed.

If retaining the `$GITHUB_DOMAIN` is preferred, an alternative approach similar to #2031 can be considered. Please provide feedback on the preferred implementation.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -4,7 +4,7 @@ author: 'GitHub'
 description: 'It is a simple combination of various linters, written in bash, to help validate your source code.'
 runs:
   using: 'docker'
-  image: 'docker://ghcr.io/super-linter/super-linter:v5.0.0'
+  image: 'docker://ghcr.io/super-linter/super-linter:v5'
 branding:
   icon: 'check-square'
   color: 'white'


zkoppert:
Hi @Shegox! Thanks for taking the time to submit this PR!

Since this change removes the `GITHUB_DOMAIN`, I am labelling this as a breaking change so that we can handle version number changes appropriately. That is not to say there is anything wrong with the change at all, just making sure users are informed they will need to update their existing configurations.

zkoppert:
@Shegox Have you been able to verify the functionality at all by testing in your local setup?

zkoppert:
Yes, I agree with going with the direction you mentioned ("same logic as for GITHUB_API_URL"). That would also be a benefit because this would no longer be a breaking change.

PR #4375, opened 2023-06-12
Update Kubernetes File Detection

ejether:
Fixes # https://github.com/super-linter/super-linter/issues/4350

## Proposed Changes

1. Add `match start` regex to `apiVersion` and `kind` lines in the `DetectKubernetesFile()`

This will prevent false positives where other forms of `yaml` files may have these markers in them but not at the line start, (like they would in a Kubernetes manifest), or if they are commented out.  We don't want to go about `kubeconform`ing the entire yaml file just because there are some embedded manifests.
  
This may change some incorrectly expected behavior.

### Author/Contributor

None Required

### Reviewing Maintainer


Hanse00 (approved):
Makes sense to me. Any input @zkoppert?

zkoppert (approved):
👍🏻 Thanks for submitting the PR! Appreciate it!

PR #4422, opened 2023-07-03
Improve GITHUB_DOMAIN/GITHUB_SERVER_URL default value handling

Shegox:
Follow up to #4362
Inspired by #2031

## Proposed Changes

1. If `GITHUB_DOMAIN` exists, use its value as GitHub domain
2. If `GITHUB_DOMAIN` does not exist, but `GITHUB_SERVER_URL` is already set, use its value as GitHub domain
3. If neither `GITHUB_DOMAIN` nor `GITHUB_SERVER_URL` is set, use `https://github.com` as GitHub domain

## Readiness Checklist

You can test the change with following configuration: 
- Update the docker image in the `action.yml` to `ghcr.io/shegox/super-linter:github-domains-2` in your testing fork
- Use it with `[your-fork]/github-super-linter@[your-branch]`

Testing with following settings in GitHub Actions:
1. GITHUB_DOMAIN=`https://github.enterprise.com`  ✅ 
2. GITHUB_DOMAIN=`github.enterprise.com`  ✅ 
3. GITHUB_DOMAIN unset ✅ 

### Author/Contributor

### Reviewing Maintainer

//cc @zkoppert

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -77,6 +77,15 @@ fi
 # Remove trailing slash if present
 GITHUB_API_URL="${GITHUB_API_URL%/}"
 
+# GitHub server url
+if [ -n "$GITHUB_DOMAIN" ]; then
+  GITHUB_SERVER_URL="${GITHUB_DOMAIN}"
+elif [ -z "$GITHUB_SERVER_URL" ]; then
+  GITHUB_SERVER_URL="https://github.com"
+fi
+# Extract domain name from URL
+GITHUB_SERVER_URL=$(echo "$GITHUB_SERVER_URL" | cut -d '/' -f 3)
+
 # Default Vars
 DEFAULT_RULES_LOCATION='/action/lib/.automation'          # Default rules files location
 LINTER_RULES_PATH="${LINTER_RULES_PATH:-.github/linters}" # Linter rules directory
@@ -654,7 +663,6 @@ CallStatusAPI() {
 
     debug "URL: ${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/statuses/${GITHUB_SHA}"
 
-    GITHUB_DOMAIN=$(echo "$GITHUB_DOMAIN" | cut -d '/' -f 3)
 
     ##############################################
     # Call the status API to create status check #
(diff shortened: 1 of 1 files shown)


lib/linter.sh
```diff
@@ -77,6 +77,15 @@ fi
...
 GITHUB_API_URL="${GITHUB_API_URL%/}"
 
+# GitHub server url
+if [ -n "$GITHUB_DOMAIN" ]; then
+  GITHUB_SERVER_URL="${GITHUB_DOMAIN}"
+elif [ -z "$GITHUB_SERVER_URL" ]; then
+  GITHUB_SERVER_URL="https://github.com"
+fi
+# Extract domain name from URL
+GITHUB_SERVER_URL=$(echo "$GITHUB_SERVER_URL" | cut -d '/' -f 3)
```
zkoppert, line 87: I don't think this is going to work with existing configurations. If I understand correctly, existing configurations `$GITHUB_DOMAIN` looks like `mycompany.github.com` so that then down on line [677](https://github.com/super-linter/super-linter/pull/4422/files#diff-0b89e6e14353371aacf69d4b9067d4481dd99c6994409b32e3b906c69d478c21R677), the `https://` is added in. 

Your pull request assumes that the user is inputting the `https://` and if they don't it chomps up some of the url. ie. `echo "zack.github.com/" | cut -d '/' -f 3` returns nothing because the cut has chopped off the who string. 

Instead, this code should either detect if the `https://` is present in the variable or as previous code did, it should assume it is not present. Does that make sense?

Shegox: thats correct, but this is the current behavior as well as I copied the logic from originally [line 657](https://github.com/super-linter/super-linter/pull/4422/files#diff-0b89e6e14353371aacf69d4b9067d4481dd99c6994409b32e3b906c69d478c21L657) to the top of the script. This code works with both URLs starting with `https://` (e.g. `https://zack.github.com`) as well as ones without **any** `/` (e.g. `zack.github.com`):
image

zkoppert: Thanks for clarifying. Looking at the code before and after again I think i missed that the first time.

zkoppert (approved):

PR #4441, opened 2023-07-03
Use calculated major version instead of hardcoded v5

### LINKED ISSUE #4419: Don't hardcode major version in release builds
The release workflow currently hardcodes that the release should be tagged as `v5` as well as the more granular minor and patch version. For now this is not a problem, but eventually we may have another major version, and having to remember to update this value might bite us.

https://github.com/super-linter/super-linter/blob/main/.github/workflows/release.yml#L87

We should find a way to ensure this value is inserted dynamically.

zkoppert:
fixes #4419

## Proposed Changes

1. Use calculated major version instead of hardcoded v5

### Author/Contributor

### Reviewing Maintainer


Hanse00 (approved):
Thanks for handling this!

zkoppert:
`.github/workflows/release.yml` failed because it has trailing whitespace. How dare I! 😆 Fixing that now so it (hopefully) passes the lint.

PR #4505, opened 2023-07-24
Update docker image used by slim version of the action

### LINKED ISSUE #4494: Is the slim image no longer being updated?
When transitioning over to using reusable workflows I noticed that the slim super-linter version I'd pinned didn't match the pulled docker image.

From my ```lint.yml``` file
```uses: super-linter/super-linter/slim@v5.1.1```

From my workflow logs
image

Note
v5.1.1 vs v5.0.0

Additionally the slim version of the action.yml is still referencing the old github/super-linter repo and not the newer super-linter/super-linter repo

grahame-student:
Fixes #4494

Update the docker image referenced in the slim metadata file to use the same version as the main version of the action.

## Proposed Changes

1. Uses the updated super-linter/super-linter repo
2. Bumps the slim image version to 5.2.0 to align with the full-fat version of the action

### Author/Contributor
- [ ] If documentation is needed for this change, has that been included in this pull request (N/A)

### Reviewing Maintainer


grahame-student:
Would it also be worth having a mechanism to either
* Keep the two versions aligned automatically?
* Detect when the two versions diverge?

zkoppert (approved):

zkoppert:
@grahame-student Yes, that would be really helpful. I would lean towards automation to keep the versions in sync.

PR #4563, opened 2023-08-15
Changing DEFAULT_BRANCH in README to main

aatmmr:
## Proposed Changes

1. Changes `DEFAULT_BRANCH` to `main`. `master` is still the default, however, the README does not align with the referenced `TEMPLATES/linter.yml` which is fixed with this PR.

### Author/Contributor

### Reviewing Maintainer


zkoppert (approved):
Thanks for fixing!

PR #1484, opened 2021-04-26
Move r

admiralAwkbar:
Pull `R` language into default locations...

PR #4579, opened 2023-08-21
fix: Move commands to correct place

Gustl22:
## Proposed Changes

During #1490 and #1484, the commands got moved to the wrong place.

I noticed this, when removing lua, as it's site was offline for while, and needed to build.

### Author/Contributor

### Reviewing Maintainer


Dockerfile
```diff
@@ -241,8 +241,8 @@ RUN --mount=type=secret,id=GITHUB_TOKEN /install-google-java-format.sh && rm -rf
 #################################
 # Install luacheck and luarocks #
 #################################
-COPY scripts/install-lua.sh /
-RUN --mount=type=secret,id=GITHUB_TOKEN /install-lua.sh && rm -rf /install-lua.sh
+#COPY scripts/install-lua.sh /
```
Gustl22, line 244: I'll revert this, once I tested it to work without lua

Gustl22:
Unfortunately Swift is not supporting Alpine Image and therefore swift-format cannot be build:
https://github.com/apple/swift-docker/issues/231

zkoppert (approved):
Thanks for fixing this up! 🎉

PR #4674, opened 2023-09-20
renovate-config-validator: validate shareable config preset

massongit:
## Proposed Changes

1. Add a environment variable `RENOVATE_SHAREABLE_CONFIG_PRESET_FILE` as a renovate validate shareable config preset for `renovate-config-validator`.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
Thanks for this PR!

`default.json` looks like a fairly common name. Is there anything we can do to also check for the path? Or some distinctive content inside the file?

massongit:
>`default.json` looks like a fairly common name. 

I agree.
I think that in many cases it is placed directly under the repository.

massongit:
How about specifying additional files like `default.json` that we want to validate using a environment variable?

ferrarimarco:
Apparently, there's also the chance to arbitrarily name files that contain presets: https://docs.renovatebot.com/config-presets/#github-hosted-presets. Maybe the environment variable approach is the one to follow here, although I would also consider how to pass multiple files there, if we don't find a way to select files based on contents, as we do for other linters.

As a side note, [renovate pre-commit hook](https://github.com/renovatebot/pre-commit-hooks/blob/main/.pre-commit-hooks.yaml) doesn't check presets.

massongit:
I add  a environment variable `RENOVATE_SHAREABLE_CONFIG_PRESET_FILE`.

lib/functions/buildFileList.sh
```diff
@@ -337,8 +337,9 @@ function BuildFileList() {
     fi
 
     # See https://docs.renovatebot.com/configuration-options/
-    if [[ "${BASE_FILE}" =~ renovate.+json5? ]] ||
-      [ "${BASE_FILE}" == ".renovaterc" ] || [[ "${BASE_FILE}" =~ .renovaterc.+json5? ]]; then
+    # See https://docs.renovatebot.com/config-presets/
+    if [[ "${BASE_FILE}" =~ renovate.json5? ]] || [ "${BASE_FILE}" == "${RENOVATE_SHAREABLE_CONFIG_PRESET_FILE}" ] ||
```
ferrarimarco, line 341: `RENOVATE_SHAREABLE_CONFIG_PRESET_FILE`: this variable probably needs to be renamed to `RENOVATE_SHAREABLE_CONFIG_PRESET_FILE_NAME`.

ferrarimarco:
Thanks for the updates to the PR.

I think this somewhat helps, although the change doesn't solve the issue in the first place because:

- What would happen if the user has a file named `default.json` in their repo that's not related to Renovate? Renovate validator is probably going to complain. Actually, we may implement a test case for this.
- How can a user validate multiple presets in one super-linter run? It seems that [Renovate supports this pattern](https://docs.renovatebot.com/config-presets/#github-hosted-presets).

lib/functions/buildFileList.sh
massongit: I renamed to it: https://github.com/super-linter/super-linter/pull/4674/commits/a58fee214a8d6d1fb1220a017875323bbd401a89

massongit:
>How can a user validate multiple presets in one super-linter run? It seems that [Renovate supports this pattern](https://docs.renovatebot.com/config-presets/#github-hosted-presets).

I fix to specify multiple additional files with a environment variable: https://github.com/super-linter/super-linter/pull/4674/commits/a58fee214a8d6d1fb1220a017875323bbd401a89

massongit:
>What would happen if the user has a file named `default.json` in their repo that's not related to Renovate? Renovate validator is probably going to complain. Actually, we may implement a test case for this.

I think that a non renovate-related file `default.json` will not validate if `RENOVATE_SHAREABLE_CONFIG_PRESET_FILE_NAMES` is not specified. Because the default value of this environment variable is empty.
I agree that it would be better to have a test for this case. However, I have no idea how to implement multiple tests with different given environment variables for one validator.

ferrarimarco (approved):

ferrarimarco:
@zkoppert can you please have a look at this as well, given that we're adding a new env variable? thanks!

zkoppert (approved):
Great! Thanks for the contribution! The out of the norm test file naming seems fine with me in this case.

PR #4694, opened 2023-09-29
Add exact version number to action.yml on release

### LINKED ISSUE #4650: version pinning needs to be exact?
Isn't that behaviour potentially unexpected? When I pin super-linter/slim to v5.2.2, it'll fetch the slim-v5 image, which is going to be some more recent version than v5.2.2 in the future, no?

_Originally posted by @bewuethr in https://github.com/super-linter/super-linter/issues/4564#issuecomment-1716147164_

zkoppert:
Fixes #4650

## Proposed Changes

1. Update action.yml and slim/action.yml with exact version numbers just before tagging a release.

### Author/Contributor

### Reviewing Maintainer


Hanse00:
Just to ensure I understand this change correct:

Each time we execute the release workflow (Thereby, building a new ghcr image), two automated git commits will be pushed to main, updating these two files: https://github.com/super-linter/super-linter/blob/main/action.yml, https://github.com/super-linter/super-linter/blob/main/slim/action.yml with the fully qualified image name of the newest container image.

The purpose of that, is to ensure that anyone using the GitHub Action in their own workflows, always has access to the latest super-linter container in their workflows.

Is that correct?

zkoppert:
@Hanse00 Yes that is correct

Hanse00 (approved):

zkoppert:
The workflow failure seems to be coming from [403 returns from the API](https://github.com/super-linter/super-linter/actions/runs/6410181832/job/17402919697?pr=4694#step:6:2945). I'll have to find some time to dig deeper into why this is happening.

zkoppert:
closing as duplicate of https://github.com/super-linter/super-linter/pull/4733

PR #4715, opened 2023-10-03
Fix image name in run-linter-locally.md

hamdrew:
The image name was missing `ghcr.io`, which caused the command to fail.


zkoppert (approved):
Thanks for the fix!

PR #4761, opened 2023-10-17
Revert exact version pinning

zkoppert:
Exact version pinning wont be available for super-linter until merge queue and gh cli merges are compatible. See https://github.com/cli/cli/issues/7213

## Proposed Changes

1. Revert release workflow code that attempts to change the version pinned in the action.yml files

### Author/Contributor

### Reviewing Maintainer


zkoppert:
reopening #4650

PR #4796, opened 2023-10-27
Simplify pip requirements files

### LINKED ISSUE #3727: Simplify dependency files

We currently have transitive dependencies defined in dependency files. I had a quick look at the pip requirements files and we may shorten them a bit. For example, the [ansible-lint requirements file](https://github.com/github/super-linter/blob/main/dependencies/python/ansible-lint.txt) lists transitive dependencies.

This leads to:

- (arguably) higher maintenance burden because we need to take care of more dependencies (more dependabot PRs, more reviews, more builds...)
- Duplicate effort because transitive dependencies should be already handled by the maintainers of each linter that we install.

1. Remove transitive dependencies from our dependency files.

ferrarimarco:
Fixes #3727

## Proposed Changes

1. Remove transitive dependencies of each linter from pip requirements files. We now rely on each linter to install its own dependencies. This will reduce the number of dependency updates to take care of, and reduce the number of failing builds due to dependency conflicts. It may also make the image somewhat smaller because some of these transitive dependencies were dropped by their upstreams, and not needed anymore.
2. Remove two dependency update scripts that we don't need anymore.
3. Fix a couple of typos in a comment.

### Author/Contributor

### Reviewing Maintainer


zkoppert (approved):
✨ Thanks for taking this on!! 🎉

PR #4821, opened 2023-10-31
Change WORKDIR in script, not relying on docker run -w by user

### LINKED ISSUE #4495: Log spammed with "fatal: not a git repository" when running locally
## Describe the bug

When running super-linter locally via docker, the log is full of messages about not being in a git repository.

## Expected behavior

These messages should not appear - and do not appear when running in a GitHub Action

## Steps to Reproduce

I am running on Docker 20.10.17 rootless. I see htis issue on both docker images d580fa3ea8af and 7c1a9b0b1b4e. 

I run:
```
docker run -it --rm \
        -e RUN_LOCAL=true \
        --env-file ".github/super-linter.env" \
        -v /my/code/path:/tmp/lint github/super-linter:v5
```

With my .github/super-linter.env
```
ACTIONS_RUNNER_DEBUG=true
IGNORE_GITIGNORED_FILES=true
DEFAULT_WORKSPACE=/tmp/lint
```

## Logs

```
2023-07-20 12:24:51 [DEBUG]   /tmp/lint/a/file is NOT a Kubernetes descriptor
2023-07-20 12:24:51 [DEBUG]
2023-07-20 12:24:51 [DEBUG]   File:[/tmp/lint/a/file], File_type:[py], Base_file:[file.py]fatal: not a git repository (or any of the parent directories): .git
2023-07-20 12:24:51 [DEBUG]
2023-07-20 12:24:51 [DEBUG]   File:[/tmp/lint/a/file], File_type:[/tmp/lint/a/file], Base_file:[ssh_config]
fatal: not a git repository (or any of the parent directories): .git
2023-07-20 12:24:51 [DEBUG]   Failed to get filetype for:[/tmp/lint/a/file]!
2023-07-20 12:24:51 [DEBUG]
2023-07-20 12:24:51 [DEBUG]   File:[/tmp/lint/a/file], File_type:[csv], Base_file:[template.drawio.csv]
fatal: not a git repository (or any of the parent directories): .git
2023-07-20 12:24:51 [DEBUG]   Failed to get filetype for:[/tmp/lint/a/file]!
2023-07-20 12:24:51 [DEBUG]
2023-07-20 12:24:51 [DEBUG]   File:[/tmp/lint/a/file], File_type:[txt], Base_file:[requirements.txt]
fatal: not a git repository (or any of the parent directories): .git
```

## Possible fixes

The Docker flag `--workdir /tmp/lint` resolves the issue, so that should either be added to the documentation, or fixed in the code to not require this.

kftsehk:
Fixes #4495

## Proposed Changes

pushd to `$DEFAULT_WORKSPACE` before resolving files

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -502,7 +502,7 @@ GetGitHubVars() {
 
     info "Linting all files in mapped directory:[${GITHUB_WORKSPACE}]"
 
-    pushd "$GITHUB_WORKSPACE" >/dev/null || exit 1
+    pushd "${GITHUB_WORKSPACE}" >/dev/null || exit 1
 
     # No need to touch or set the GITHUB_SHA
     # No need to touch or set the GITHUB_EVENT_PATH
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -502,7 +502,7 @@ GetGitHubVars() {
 
     info "Linting all files in mapped directory:[${GITHUB_WORKSPACE}]"
 
-    pushd "$GITHUB_WORKSPACE" >/dev/null || exit
+    pushd "$GITHUB_WORKSPACE" >/dev/null || exit 1
 
     # No need to touch or set the GITHUB_SHA
     # No need to touch or set the GITHUB_EVENT_PATH
--- a/lib/linter.sh
+++ b/lib/linter.sh
@@ -500,7 +500,9 @@ GetGitHubVars() {
       fatal "Provided volume is not a directory!"
     fi
 
-    info "Linting all files in mapped directory:[${DEFAULT_WORKSPACE}]"
+    info "Linting all files in mapped directory:[${GITHUB_WORKSPACE}]"
+
+    pushd "$GITHUB_WORKSPACE" >/dev/null || exit
 
     # No need to touch or set the GITHUB_SHA
     # No need to touch or set the GITHUB_EVENT_PATH


lib/linter.sh
```diff
@@ -500,7 +500,9 @@ GetGitHubVars() {
       fatal "Provided volume is not a directory!"
     fi
 
-    info "Linting all files in mapped directory:[${DEFAULT_WORKSPACE}]"
+    info "Linting all files in mapped directory:[${GITHUB_WORKSPACE}]"
+
+    pushd "$GITHUB_WORKSPACE" >/dev/null || exit
```
ferrarimarco, line 505: - Can you use `${GITHUB_WORKSPACE}` for consistency?
- Can you add a comment to explain what you're doing, and why not using a `cd`?
- Can you exit with an error if the directory doesn't exist?

ferrarimarco (changes requested):
Thanks for this PR! Mostly ok, we probably need some more documentation.

Thanks!

lib/linter.sh
ferrarimarco: Also, how would this change impact users? Shall they do something different?

kftsehk: Sorry that the typo of fixes #4459 instead of #4495 had confused you.

This fix applies to `RUN_LOCAL=true`

File resolver runs in `WORKDIR` defined in `Dockerfile`, it does not define one, and defaults to `/`. The reason of #4495 is the file resolver scan `$PWD = /` for code, which contains `/root /bin /lib` etc, all of which results in `fatal: not a git repository (or any of the parent directories): .git`, until it finds `/tmp/lint`

Now `docker run -w /tmp/lint` is required to suppress the above error message even if user is not aware of `DEFAULT_WORKSPACE` config

https://github.com/super-linter/super-linter/blob/c3ac3aa5d9c14fd80edc78eb3c4af8bea0849fa6/README.md?plain=1#L284


There is probably no reason to define `WORKDIR` in Dockerfile, as `DEFAULT_WORKSPACE` can be changed at runtime if user wish to mount elsewhere other than `/tmp/lint` via `docker run -v /host/filesystem/my/code:<somewhere-inside-container-not-tmp-lint>`

So the solution imo is to change to evaluated `$GITHUB_WORKSPACE` right after its evaluation

> * Can you use `${GITHUB_WORKSPACE}` for consistency?

is used in commit

> * Can you add a comment to explain what you're doing, and why not using a `cd`?

`pushd` to have a backup of docker's `WORKDIR` if ever needed, imo obvious enough to notice here that `info "Linting all files in mapped directory:[${GITHUB_WORKSPACE}]"` has to go into directory `$GITHUB_WORKSPACE`

> * Can you exit with an error if the directory doesn't exist?

https://github.com/super-linter/super-linter/blob/3d1ea8f5adeb62068bf8ecb318aa732541bb8de2/lib/linter.sh#L499-L501

Existence checked here, I guess it might fail for some strange reason like missing `+x` file permission

> Also, how would this change impact users? Shall they do something different?

No longer need to type docker flag `-w /tmp/lint`, they shall still use it this way but it won't do anything good or bad.

ferrarimarco: Thanks for the explanation!

I think I still see `$GITHUB_WORKSPACE` instead of `${GITHUB_WORKSPACE}`. :)

ferrarimarco (approved):
LGTM. @zkoppert and @Hanse00  probably want to have a look because it changes things a bit under the hood.

zkoppert (approved):
That's for explaining the changes. I looked through our building documentation and don't see anything that needs to change there to match this code change.

PR #4088, opened 2023-04-16
Batched & parallel support for cfn-lint, eslint, gitleaks

kftsehk:
Proposal to fix
- #4084  

Currently planned to provide implementation for
- [x] cfn-lint
- [x] eslint
- [x] gitleak

Use cases includes
- Faster whole repo linting on cloud (changed files can affect unchanged files)
- Local linting on commit / push without having developers to wait too much

## Proposed Changes

_Batched_ and _parallel_ linting for use case involving large number of files.
- _Batched_: Use `$LINTER_COMMAND FILE1 FILE2 FILE3 ....` instead of `$LINTER_COMMAND FILE` looping over all files
- _Parallel_: Files are divided into not-so-small chunks and multiple linter instances up to `nprocs` are run in parallel
As for most parallelization, this patch does not reproduce exact output of serial version

To enable this experimental worker, set env `EXPERIMENTAL_BATCH_WORKER` to `true`

In order to mimic serial behavior as closely as possible, the implementation
- show linter output as is, 
- does not break or interleave linter output in any way,
- provide the same error file count (lines outputting text such as `[ERROR]   ERRORS FOUND in JAVASCRIPT_ES:[123]`) as serial version when super-linter is completed

Current ERROR output on serial version cannot be recovered, since
- it is not possible retrieve error code of linter for individual file, batching implies only 1 return code is given for a batch of files.

### Performance comparison

A figure is provided only for eslint on a repo that I have on hand
- 8 core laptop
- approx. 450 .js file
- timing only on JAVASCRIPT_ES block

Test 1: 429 files contains linter error (not warning) artificially introduced (mainly `react/*` rules that was turned off for the project)
- original: `7 min 21 sec`
- new impl: `14 sec` (it is not necessary to parallel, still under 30s without parallel)

Test 2: No artificially introduced error (0 error, several warnings)
- original: `7 min 4 sec`
- new impl: `5 sec`

For the experimental implementation, majority of the performance issue is in bash-script implemented output parser to calculate number of error files, it is not too bad anyway.

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request
- [ ] Test case count is not yet recovered

### Reviewing Maintainer
- [x] Label as `breaking` if this is a large fundamental change
- [x] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

Issue #2920, opened 2022-05-20
Run linters in parallel.


tmc:
This project does not appear to run linters in parallel -- this causes the time it takes to run to in some cases be non-trivial.

PR #4852, opened 2023-11-12
Parallelize for all linters

kftsehk:
Follow up #4088

## Proposed Changes

 since parallelizing should not lead to linter incompatibility, except for a non-deterministic order of files being linted

1. Shall all linter be parallelized by a standard script?
2. Possible to make parallelization out of experimental if passing all the test cases?

### Author/Contributor
- [ ] Implementation
- [ ] Tests

### Reviewing Maintainer


kftsehk:
you are right, there is no code yet, still thinking of how to best make it compatible with current implementation during free time.

I added to the title to [Do not merge] WIP in case any other get confused.

ferrarimarco:
As part of this PR, can we please remove the experimental path added in #2920 ?

ferrarimarco:
@kftsehk In #5041, I've changed how Gitleaks, textlint, and Jscpd run. Now, they run on the whole workspace instead of running one by one. In following PRs, I'll probably refactor more, probably starting from the ones you implemented experimental batch support for, so we can compare the two approaches.

In that PR, I'm also removing experimental batch support, so we can see how we can work on it in this PR, eventually for all linters.

Thanks a lot for pushing this effort!

ferrarimarco:
To give you full context: this was prompted by Jscpd warning about duplicated code in the files that handled the experimental batch mode.

kftsehk:
The main concern with linting whole WORKDIR, e.g. `eslint .` is breaking the "error count" of super-linter, which might not be desirable.

If such breaking change is acceptable, a lot of linter can simply call `<linter-name> "$GITHUB_WORKDIR"`

ferrarimarco:
> The main concern with linting [the] whole WORKDIR, e.g. `eslint .` is breaking the "error count" of super-linter, which might not be desirable.

This is an interesting perspective. I wouldn't consider that change breaking though, because it's not part of the super-linter API. This can be improved in the future in the context of a new reporting feature if we have contributors willing to work on that particular reporting feature.

> If such breaking change is acceptable, a lot of linter can simply call `<linter-name> "$GITHUB_WORKDIR"`

#5041 adopts this approach for:

- jscpd, because looking for duplicates in files processing them one by one doesn't consider duplicates in other files, to it makes this linter way less effective, and it costs a lot in terms of overall performance.
- textlint, because it's based on plugins, and each plugin handles certain types of files. Also, textlint lets users specify custom extensions for a given file type in addition to the defaults. So, we let textlint handle this, instead of reimplementing the same logic. For example, with default plugins, textlint can lint `txt` and `md` files only, so there's no point in running it against all files one by one. If we/users install another textlint plugin, we don't need to modify any logic to handle an additional extension, nor custom extensions for any given plugin that users might have configured.
- Gitleaks, because we want to scan the whole workspace anyway.

This approach also has the added benefit of letting users rely on the file listing and ignoring logic of each linter, so their configuration becomes more portable if they decide to run the same linter by other means.

ferrarimarco:
Also, I think that the approach I describe in my previous comment is not in contrast with your suggestion to run linters in parallel. They can happily work together!

ferrarimarco:
After I rebased this PR, it got closed, sorry @kftsehk ! I'll see if I can reopen it.

ferrarimarco:
I don't have access to your `parallel` branch anymore because the PR got closed.

As soon as you can push at least one commit there, the system should let me reopen this PR. Sorry for the confusion!

For context: the PR where you initially introduced the experimental parallel feature is #4088. May be useful to inspect the original implementation.

Thanks a lot for working on this!

kftsehk:
sure, will update a bit later

PR #4889, opened 2023-11-22
Exit on errors when running Git

ferrarimarco:
Fixes #4716

## Proposed Changes

1. ~Don't conflate two git commands to pull changes and switch branches into a single subshell to avoid hiding command failures.~ Don't pull changes or switch branches because these are potentially disruptive operations on the local repository. It is now a responsibility of the user to do so, if they wish.
2. Check if the `DEFAULT_BRANCH` exists in the current repository before using it.
3. If the `DEFAULT_BRANCH` doesn't exist, try falling back to `origin/${DEFAULT_BRANCH}`.
4. If `origin/${DEFAULT_BRANCH}`, fail.
5. Exit with an error if diff commands return errors to avoid hiding command failures
6. Check if `GITHUB_SHA` exists in the current repository before using it.
7. Remove the undocumented `LOCAL_UPDATES` configuration flag.
8. Changed error output level from `info` to `error` on related errors.
9. Add debug output for `LOCAL_UPDATES` and some return codes.
10. Check if `GITHUB_SHA` exists before using it. If it doesn't exist, fail.
11. Forward stderr to stdout when pulling changes so we don't lose any output.
12. Reduced duplication by moving the git diff command to `DIFF_GIT_DEFAULT_BRANCH_CMD`, and reusing that variable.
13. Check the return code when updating `GITHUB_SHA`.
14. Add a check to inform the user if the current repository is shallow.
15. Remove some commented code.
16. Move `RUN_LOCAL` and `TEST_CASE_RUN` variable initialization up in the process so we can reuse these variables earlier.
17. Check if `GITHUB_WORKSPACE` is a Git repository before relying on that behavior.
18. Remove `DIFF_GIT` entirely because we don't need it anymore.
19. Move some debug messages when defining related variables to be less verbose.

### Author/Contributor

### Reviewing Maintainer


zkoppert (approved):
Looks great! 🚀

ferrarimarco:
There is another issue to fix: https://github.com/super-linter/super-linter/actions/runs/6961127801/job/18942089010#step:7:1753

ferrarimarco:
It turns out that there were several issues. It took a while to fix them.

Basically, there was an undocumented flag `LOCAL_UPDATES` (default empty) that disabled pulling latest changes from the repo (`git pull`) and switching to `DEFAULT_BRANCH` (`git checkout`). We only set it to `false` when running the `Test local action` job in CI. Also, the `LOCAL_UPDATES` flag wasn't considered when switching back to the original branch.

Given that the operations I mentioned above (pull, checkout) are potentially disruptive for (a local copy of) a Git repository, I propose removing this logic entirely. It is now a responsibility of the user to fetch the latest changes from the remote, and to checkout the right branch before running super-linter.

zkoppert (dismissed):
Seems good to go functionally. Thanks for making a a breaking change. 

We could further refactor the diff cod stuff. I don't think there is a purpose for that variable any longer if we are just setting its value to DIFF_GIT_DEFAULT_BRANCH_CMD. In other words let's replace any reference to DIFF_CMD with DIFF_GIT_DEFAULT_BRANCH_CMD instead.

zkoppert (approved):
✨

ferrarimarco:
@zkoppert thanks for the review! Do we want to cut a new release after merging minor updates, and before merging this one?

zkoppert:
Great question. Yes, lets create a release first, then merge this one, then release again.

ferrarimarco:
Btw, about points 3 and 4.

Do we want to make `origin` customizable with (yet another) variable that defaults to `origin`, or are we good having it hardcoded? It's a simple change, although this PR is already big enough...

zkoppert:
Hard coded is fine for now. We can revisit later if it presents an issue for folks.

tremblaysimon:
> Seems good to go functionally. Thanks for making a a breaking change.
> 
> We could further refactor the diff cod stuff. I don't think there is a purpose for that variable any longer if we are just setting its value to DIFF_GIT_DEFAULT_BRANCH_CMD. In other words let's replace any reference to DIFF_CMD with DIFF_GIT_DEFAULT_BRANCH_CMD instead.

What does it mean in other words is now you need to fetch all history for all branches and tags. So in the `actions/checkout` you must specify `fetch-depth: 0` parameter.

PR #4926, opened 2023-11-30
Validate container image labels

### LINKED ISSUE #4895: Add back container image labels tests and remove obsolete scripts
We are currently missing labels check. We used to have them. This got lost when reworking workflows.

Also, `.automation/cleanup-docker.sh` and `.automation/upload-docker.sh` are probably not needed anymore because we handle that with workflows now. Wanted to check before cleaning up.

```yaml
      - name: Run container image label test cases
        if: ${{ github.repository == 'github/super-linter' && github.ref != 'refs/heads/main' }}
        shell: bash
        run: .automation/validate-docker-labels.sh "${{ matrix.images.image-id }}"
```

Also, remove obsolete scripts.

@zkoppert WDYT?

ferrarimarco:
Fixes #4895

## Proposed Changes

1. Enable container image labels validation.
2. Fix a bug that caused the wrong revision to be set when building container images in the context of a pull request.
3. Remove obsolete scripts and docs.
4. Enable Docker buildx on GitHub Actions as we do locally.

### Author/Contributor

### Reviewing Maintainer


Makefile
```diff
@@ -51,6 +51,18 @@ ifeq ($(SUPER_LINTER_TEST_CONTAINER_URL),)
...
+ifeq ($(BUILD_DATE),)
+BUILD_DATE := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+endif
+
+ifeq ($(BUILD_REVISION),)
+BUILD_REVISION := $(shell git rev-parse HEAD)
+endif
+
+ifeq ($(BUILD_VERSION),)
+BUILD_VERSION := $(shell git rev-parse HEAD)
```
zkoppert, line 63: These had --short in them before. Just want to make sure that the change is desired.

Also, why do we have both? Seems like they are exactly the same thing...

ferrarimarco: This is intended, yes, so users can compare it with the value that we assign to the `org.opencontainers.image.revision` and `org.opencontainers.image.version` when building the image. We use the full hash in that case, so let's be consistent :)

Thanks for the in-depth review!

ferrarimarco: About having both: I think we were supposed to use the `org.opencontainers.image.version` label and `BUILD_VERSION` for the version. I thought about removing the `version` stuff, but I didn't because it can come in handy in the future if we refactor the build process to get that setting right. I'll create an issue to track this, so we don't forget.

zkoppert (dismissed):
🚢

PR #4933, opened 2023-12-04
Change directory when checking ignored files

ferrarimarco:
## Proposed Changes

1. Change the working directory to the workspace when checking files that should be ignored by Git

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/lib/functions/buildFileList.sh
+++ b/lib/functions/buildFileList.sh
@@ -304,7 +304,7 @@ function BuildFileList() {
     ###################################################
     # Filter files if FILTER_REGEX_EXCLUDE is not set #
     ###################################################
-    if [ "${IGNORE_GITIGNORED_FILES}" == "true" ] && git -C "${WORKSPACE_PATH}" check-ignore "$FILE"; then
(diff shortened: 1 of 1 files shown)


lib/functions/buildFileList.sh
```diff
@@ -304,7 +304,7 @@ function BuildFileList() {
     ###################################################
     # Filter files if FILTER_REGEX_EXCLUDE is not set #
     ###################################################
-    if [ "${IGNORE_GITIGNORED_FILES}" == "true" ] && git check-ignore "$FILE"; then
+    if [ "${IGNORE_GITIGNORED_FILES}" == "true" ] && git -C "${WORKSPACE_PATH}" check-ignore "$FILE"; then
```
zkoppert, line 307: Is `WORKSPACE_PATH` set when running this locally or will this only work when run as an action?

ferrarimarco: I think you're right! Refactored to `GITHUB_WORKSPACE`, as it should have been. We set that path in both cases. Leaving this unresolved so you can check if the change addresses this.

Thanks for reviewing!

PR #4934, opened 2023-12-04
Don't write colors and logs on disk if not necessary

ferrarimarco:
## Proposed Changes

1. Don't write logs on disk if the `CREATE_LOG_FILE` is not set to `true`. This resulted in about 50% increase in performance in our CI builds. For example, compare the results of this build against [this](https://github.com/super-linter/super-linter/actions/runs/7087431035/job/19287666006?pr=4944).
2. Don't write ANSI color markers in the log file. This makes the log file more readable.
3. Move the log file from the temporary directory to the final location instead of using `cat` to copy data.
4. Reduce some code duplication by defining the log format once.

Example log line in the log file before this change:

```shell
[0m2023-12-04 10:06:17 [0;34m[INFO][0m   Linting [JSCPD] files...[0m
```

After:

```shell
2023-12-04 14:20:46 [INFO]   Linting [JSCPD] files...
```

### Author/Contributor

### Reviewing Maintainer


zkoppert (approved):
🚀 🚀

PR #4953, opened 2023-12-06
Don't forcefully validate Git repos if not needed

### LINKED ISSUE #4952: Git repository validation should not be enforced when USE_FIND_ALGORITHM=true

### Current Behavior

PR #4889 (Exit on errors when running Git) was merged 2 days ago.

This code was added to lib/functions/validation.sh:

```
function ValidateLocalGitRepository() {
  debug "Check if ${GITHUB_WORKSPACE} is a Git repository"
  if ! git -C "${GITHUB_WORKSPACE}" rev-parse --git-dir; then
    fatal "${GITHUB_WORKSPACE} is not a Git repository."
  else
    debug "${GITHUB_WORKSPACE} is a Git repository"
  fi

  debug "Git branches: $(git -C "${GITHUB_WORKSPACE}" branch -a)"
}
```

Because that exits as fatal if the directory being linted isn't a git repo it completely breaks the ability to lint any directories which are not git repos.

### Expected Behavior

We've been using super-linter to lint directories which aren't git repos for quite a while now and this change broke that for us. This is a great linter and in my opinion it should work whether or not the directory is a git repo.

### Super-Linter version

```markdown
latest
```

### Relevant log output

```shell
[FATAL]   /tmp/lint is not a Git repository.
```

### Steps To Reproduce

1. Run super-linter against a directory that isn't a git repo.
2. Observe that it fails with the error "[FATAL]   /tmp/lint is not a Git repository."
3. Try the same with a previous version from before that PR, such as v5.7.1
4. Observe that it works fine with 5.7.1 and we're able to lint the contents of the directory.

### Anything else?

_No response_

ferrarimarco:
Fixes #4952

## Proposed Changes

1. Add a test case to run super-linter against a subdirectory after setting `USE_FIND_ALGORITHM=true`.
2. Move the configuration of `git safe.directory` after we set `GITHUB_WORKSPACE`, and not before because `GITHUB_WORKSPACE` might be initialized when running locally.
3. Validate variables before validating the Git repository, so that we get those variables initialized correctly before using them.
4. Validate the Git directory regardless of the value of `VALIDATE_ALL_CODEBASE` because we need a valid Git repository and references both when `VALIDATE_ALL_CODEBASE=true` and when `VALIDATE_ALL_CODEBASE=false`.
5. Validate the Git directory when `USE_FIND_ALGORITHM == false`, not when `USE_FIND_ALGORITHM != false`.
6. Initialize `GITHUB_SHA` when running locally and when `USE_FIND_ALGORITHM == false`.
7. Fail with a fatal error if we can't add the workspace to the list of safe Git directories.
8. Fix log level color markers for non-default cases.
9. Set `DEFAULT_BRANCH` when running tests.

### Author/Contributor

### Reviewing Maintainer

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -100,6 +100,7 @@ jobs:
           -e TEST_CASE_RUN=true \
           -e ANSIBLE_DIRECTORY=.automation/test/ansible \
           -e ACTIONS_RUNNER_DEBUG=true \
+          -e DEFAULT_BRANCH=main \
           -e RENOVATE_SHAREABLE_CONFIG_PRESET_FILE_NAMES="default.json,hoge.json" \
           -e ERROR_ON_MISSING_EXEC_BIT=true \
           -e TYPESCRIPT_STANDARD_TSCONFIG_FILE=".github/linters/tsconfig.json" \
@@ -112,6 +113,7 @@ jobs:
           -e RUN_LOCAL=true \
           -e OUTPUT_DETAILS=detailed \
           -e ACTIONS_RUNNER_DEBUG=true \
+          -e DEFAULT_BRANCH=main \
           -e RENOVATE_SHAREABLE_CONFIG_PRESET_FILE_NAMES="default.json,hoge.json" \
           -e ERROR_ON_MISSING_EXEC_BIT=true \
           -v "${GITHUB_WORKSPACE}:/tmp/lint" \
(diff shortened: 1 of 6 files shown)


ferrarimarco:
> When I look at the commit for this it only appears to include the code for "2. Add a workflow step to catch this issue."

Yep. As I explained in #4952, I'd like to see a failing workflow, that we'll fix by committing the change you suggested. :)

This helps us validate the fix, and may help catching this or similar issues in the future.

kftsehk:
Had another local run issue #4954 related to the same pr #4889 

So I added ci/cd tests for all non-default general super-linter behavior settings, in the hope to test local run more rigorously for blocking bugs, see PR #4955

PR #4958, opened 2023-12-07
Simplify multi status checks

ferrarimarco:
## Proposed Changes

1. Simplify the conditions to enable validation of the variables that we need to submit status checks. These checks and related logging run only if `MULTI_STATUS=true` and `RUN_LOCAL=false`.
2. Avoid emitting confusing message logs about status reports when `RUN_LOCAL=false`.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
We need #4960 before this one.

ferrarimarco:
@zkoppert going ahead with this one because it's a simple change so you can focus on reviewing other PRs :)

PR #4962, opened 2023-12-09
Simplify container image build

ferrarimarco:
## Proposed Changes

1. Define `GLIBC_VERSION` once as a global argument in `Dockerfile` to reduce duplication.
2. Update the slim image to `python:3.12.1-alpine3.19`. For some reason, dependabot was failing with a 403 when analyzing this dependency.
3. Install a specific `hadolint` version instead of `latest` to make the build more reproducible.
4. Remove the `base_image` build step entirely. This stage was largely redundant because we copied everything we installed to `slim` anyway. This saves up some build time, and makes the Dockerfile simpler.
5. Move static labels near the top of the `Dockerfile`.
6. Move dynamic labels near the bottom of the `Dockerfile` to avoid invalidating the cache when they change.
7. Move build arguments about linter versions to install where they are needed.
8. Install bash first so that we can enable additional error checks right away.
9. Move the longest `Dockerfile` build instructions earlier so we can better use the build cache.
10. Move the installation of OS packages from scripts to Dockerfile when a package is needed by more than one linter.
11. Move `terrascan` and `chktexrc` initialization from `install-ktlint.sh` to `Dockerfile` because they don't have anything to do with `ktlint`.
12. Don't install `glibc-compat` multiple times in different script, and consolidate the logic to do this in a `install-glibc.sh` script.
13. Fail on errors in subshells when installing dependencies using shell scripts (`install-xxxxx.sh`).
14. Use `phive.xml` to install dependencies instead of listing them when invoking phive.
15. Update `phive.xml` to install the same dependencies that we were installing explicitly from the command line. Note that `phive.xml` was already in `/dependencies`, but unused and outdated.
16. Install Composer using Phive instead of using an ad-hoc script.
17. Initialize TFLint plugins when building the Dockerfile so have the default ones in the image already.
18. Add missing OS package installation tests.
19. Add missing linter commands tests: `gitleaks`, `prettier`, `renovate-config-validator`, `textlint`, `ts-standard`.
20. Add missing PyPi packages tests. We weren't testing any.
21. Add missing NPM packages tests.
22. Add missing maintainers from the build failure issues assignee template.
23. Update CODEOWNERS according to the list of current maintainers.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
@zkoppert sorry if this turned to be quite big, but I kept finding new issues while fixing others :)

scripts/install-phive.sh
```diff
@@ -47,10 +16,7 @@ chmod +x phive.phar
...
 
-# Install PHARs
+# Install the PHARs listed in phive.xml
 phive --no-progress install \
-  --trust-gpg-keys 31C7E470E2138192,CF1A108D0E7AE720,8A03EA3B385DBAA1,12CE0F1D262429A5,5E6DDE998AB73B8E,51C67305FFC2E5C0 \
-  --target /usr/bin \
-  phpstan@^1.10.48 \
-  psalm@^5.17.0 \
-  phpcs@^3.8.0
+  --trust-gpg-keys 31C7E470E2138192,CF1A108D0E7AE720,8A03EA3B385DBAA1,12CE0F1D262429A5,5E6DDE998AB73B8E,51C67305FFC2E5C0,CBB3D576F2A0946F \
```
zkoppert, line 21: Any dynamic way to set these? How will we know when they need to be updated?

zkoppert (approved):

scripts/install-phive.sh
ferrarimarco: We don't unfortunately. We could trust everything, but that's probably not too secure :)

ferrarimarco:
@zkoppert thanks for your review! I fixed a few conflicts, so no major changes since you reviewed. Going ahead with this one :)

Issue #4956, opened 2023-12-07
Fail if users configured USE_FIND_ALGORITHM=true and VALIDATE_ALL_CODEBASE=false at the same time


ferrarimarco:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, super-linter allows setting `USE_FIND_ALGORITHM=true` and `VALIDATE_ALL_CODEBASE=false`.

In this case, super-linter silently ignores the `USE_FIND_ALGORITHM` setting. Users might be confused by this behavior.

### Expected Behavior

Given that this is an unsupported configuration, we should explicitly fail.

ferrarimarco:
> I think this should be a supported algorithm

When `VALIDATE_ALL_CODEBASE` is `false`, we need to diff against something, and we use Git for that. What shall we diff against without using Git?

> A situation is happens when using flag `IGNORE_GITIGNORED_FILES=true` there is an error "fatal: not a git repository: /<path/to/the/git/worktree>/worktrees/fix_java"

Currently, this is expected because we use `git check-ignore` for that https://github.com/super-linter/super-linter/blob/7a21f934b424c138cdcbeb2c070932f71f8a184b/lib/functions/buildFileList.sh#L230

To support this use case, we'd have to duplicate that logic, or find a tool that takes one or multiple `.gitignore` files, and tells if a given file is ignored according to those `.gitignore` files.

ferrarimarco:
And, in the current implementation, we should actually enable "git repository validation" if `IGNORE_GITIGNORED_FILES=true`, as we now do when `USE_FIND_ALGORITHM=false` (#4953)

ChaosEternal:
The cause is that the git repo is actually a git worktree, if mounted in docker container, it becomes an invalid git repo. 
A proper way to invoke super-linter on a git repo should be like this:

```
#!/bin/bash
GIT_TOP_LEVEL="$(git rev-parse --show-toplevel)"
GIT_COMMON="$(git rev-parse --git-common-dir)"
docker run \
  --env-file "${GIT_TOP_LEVEL}/config/lint/super-linter.env" \
  --env GITHUB_WORKSPACE="${GIT_TOP_LEVEL}" \
  --name "super-linter" \
  --rm \
  --volume "${GIT_TOP_LEVEL}:${GIT_TOP_LEVEL}" \
  --volume "${GIT_COMMON}:${GIT_COMMON}" \
[...]

PR #4965, opened 2023-12-11
Validate local git repo when ignoring files

ferrarimarco:
Related to #4956

## Proposed Changes

1. Check if the local Git repository is valid if we enable `IGNORE_GITIGNORED_FILES` because we rely on `git check-ignore` in that case.

### Author/Contributor

### Reviewing Maintainer


ferrarimarco:
@zkoppert going ahead with this simple change. :)

PR #4981, opened 2023-12-11
Update documentation

ferrarimarco:
## Proposed Changes

1. Update the documentation to better describe how to use super-linter.
2. Remove outdated information.

### Author/Contributor

### Reviewing Maintainer


zkoppert (commented):
Looks like if we are removing the `TEMPLATES/linter.yml` file it should also be removed from the test: https://github.com/super-linter/super-linter/blob/main/test/inspec/super-linter/controls/super_linter.rb#L405

ferrarimarco:
Yes, good catch!

PR #4982, opened 2023-12-11
Reduce duplication in CI and CD workflows

ferrarimarco:
## Proposed Changes

1. Define a `lint-codebase` Make target to run super-linter against the current codebase.
2. Define a `test-linters` Make target to run the super-linter test suite.
3. Add these new targets to the `test` target.
4. Remove the related build steps from the CI and CD workflows because we call `make test` there anyway.
5. Fix list numbers in README.
6. Support loading `GITHUB_TOKEN` from a file.

This is a step in the direction of reducing duplications in workflows.

### Author/Contributor

### Reviewing Maintainer


zkoppert:
This pull request looks good. I tried testing out the `make test` locally by following the [instructions here](https://github.com/super-linter/super-linter/blob/main/docs/run-linter-locally.md?plain=1#L89-L95) and got an error when checking the `image.created` tag. To be clear, I don't think this pull request has **caused** this issue but we should probably fix while we are in here. Does this pass for you @ferrarimarco or should we updated the `BUILD_DATE` here?
```
[ERROR] Invalid container image label: org.opencontainers.image.created: 2023-10-17T17:00:53Z. Expected: 2023-10-17T16:19:11Z
```

ferrarimarco:
Yes, I'll have a look at this. Thanks for the review!

ferrarimarco:
@zkoppert I think I fixed it. Can you please take another look? Thanks!

zkoppert (approved):
🚀 Great work! Thank you @ferrarimarco!

PR #4984, opened 2023-12-12
Lint Go modules

### LINKED ISSUE #3596: golangci-lint in super-linter produces false positives
## Describe the bug

Basically the same as closed as completed previous [issue #1599](https://github.com/github/super-linter/issues/1599)

It seems that this haven't fixed previous issue with https://github.com/github/super-linter/pull/3043

I just tried latest v4.9.7 both standard and slim images and they show errors for my code separated to multiple files for one package.

## Expected behavior

golang-ci scan do not produce false positive errors.

## Steps to Reproduce

1. Create two files in one folder: main.go and foo.go

```main.go
package main

var test int

func main() {
	foo()
}
```

```foo.go
package main

import "fmt"

func foo() {
	fmt.Println(test)
}
```

2. Run golangci-lint for each file separately without --fast option and get errors as expected (logs below)
3. Then add --fast option and all will be fine (logs below)
4. Copy [config used in super-linter](https://github.com/github/super-linter/blob/main/TEMPLATES/.golangci.yml) and add it to command line, the same way as it done in super-linter https://github.com/github/super-linter/blob/main/lib/linter.sh#L897 and it will produce error, even with --fast option (logs below)

I don't know how to fix this in correct way. Definitely all works fine if `.golangci.yml` is not passed.

## Logs

```log
# without fast option
$ golangci-lint run main.go 
main.go:6:2: undeclared name: `foo` (typecheck)
        foo()
        ^
$ golangci-lint run foo.go
foo.go:6:14: undeclared name: `test` (typecheck)
        fmt.Println(test)
                    ^

# with --fast
$ golangci-lint run --fast main.go
$ golangci-lint run --fast foo.go 

# with .golangci.yml
$ golangci-lint run --fast main.go  
main.go:6:2: undeclared name: `foo` (typecheck)
        foo()
        ^
$ golangci-lint run --fast foo.go  
foo.go:6:14: undeclared name: `test` (typecheck)
        fmt.Println(test)
                    ^
```

## Additional context

`golangci-lint` version 1.50.1 (also tested v1.49.0 with the same results)

I think it might be related to this issue https://github.com/golangci/golangci-lint/issues/1909

ferrarimarco:
Fixes #3596

## Proposed Changes

1. Implement a new mode for the Go linter to run against a directory. This is useful to lint an entire Go project, instead of linting individual files.
2. Introduce a new variable: `VALIDATE_GO_DIRECTORY`, defaults to `true`, to enable the this new mode.
3. Fail if there are Go modules to lint, and the user enabled both individual Go files linting and Go modules linting because this may cause false positives.

### Author/Contributor

### Reviewing Maintainer


anantdamle:
The variable option is not useful because of following two reasons.
(1) If the repo is a single project, its redundant
(2) If the repo is multi-language project, 
     (a) the code contributor can conveniently not enable it for their folder
     (b) There can be multiple go-lang modules

I propose a recursive search for a `go.mod` file and run it against all those folders.

ferrarimarco:
> The variable option is not useful because of following two reasons.
> (1) If the repo is a single project, its redundant

True. It's still worth having it to maintain backward compatibility with the current behavior.

> (2) If the repo is multi-language project, (a) the code contributor can conveniently not enable it for their folder (b) There can be multiple go-lang modules (/directories)

I don't get `(a)`, but `(b)` is true.

> > I propose a recursive search for a `go.mod` file and run it against all those folders.

That's a nice idea.

README.md
```diff
@@ -233,6 +233,7 @@ You can configure super-linter using the following environment variables:
 | **VALIDATE_GITHUB_ACTIONS**                     | `true`                          | Flag to enable or disable the linting process of the GitHub Actions.                                                                                                                                                 |
 | **VALIDATE_GITLEAKS**                           | `true`                          | Flag to enable or disable the linting process of the secrets.                                                                                                                                                        |
 | **VALIDATE_GO**                                 | `true`                          | Flag to enable or disable the linting process of the Golang language.                                                                                                                                                |
```
anantdamle, line 235: Maybe add documentation that, `VALIDATE_GO` flag checks the files individually,
also maybe some check so that there is an error if both VALIDATE_GO AND VALIDATE_GO_MODULE are set. `error = VALIDATE_GO && VALIDATE_GO_MODULE`

ferrarimarco: I added more details to the README, and a warning instead of an `error` or a `fatal` to be conservative. @zkoppert do we want to fail with a `fatal` instead? To me, it seems that @anantdamle has a point here.

zkoppert: Yeah a `fatal` seems appropriate for this type of configuration mismatch.

ferrarimarco: Done!

zkoppert (approved):
🎉

PR #5080, opened 2024-01-01
docs: fix README.md for slim action

chrisjohnson00:
## Proposed Changes

Fix to the README for how to run this action using the `slim` configuration.

## Readiness Checklist

### Pull request author tasks

- [X] I included all the needed documentation for this change.
- [X] I provided the necessary tests.
- [X] I squashed all the commits into a single commit.
- [X] I followed the [Conventional Commit v1.0.0 spec](https://www.conventionalcommits.org/en/v1.0.0/).
- [X] If this is a breaking change, write upgrade instructions in the [upgrade guide](../docs/upgrade-guide.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,8 +141,8 @@ For more information, see
 
 Super-Linter provides several variants:
 
-- `standard`: `super-linter/super-linter:[VERSION]`: includes all supported linters.
-- `slim`: `super-linter/super-linter:slim-[VERSION]`: includes all supported linters except:
+- `standard`: `super-linter/super-linter@[VERSION]`: includes all supported linters.
+- `slim`: `super-linter/super-linter/slim@[VERSION]`: includes all supported linters except:
 
   - `rust` linters
   - `dotenv` linters


chrisjohnson00:
Here is where I'm using it 

https://github.com/chrisjohnson00/plex-sad-search/blob/16b70b09cb4396bd6437c550b1d33b0a682ae96b/.github/workflows/ci.yml#L151

AFAIK, the standard usage reference is correct. However, I don't use it since the container pull takes so dang long. 

I will check (on phone now) later.

README.md
```diff
@@ -142,7 +142,7 @@ For more information, see
 Super-Linter provides several variants:
 
 - `standard`: `super-linter/super-linter:[VERSION]`: includes all supported linters.
```
ferrarimarco, line 144: Can you please refactor this one to: `super-linter/super-linter@[VERSION]`?

Thanks!

ferrarimarco (changes requested):
One small change and we should be good to go.

README.md
chrisjohnson00: oh yeah, good catch

ferrarimarco (approved):

ferrarimarco:
@chrisjohnson00 last thing to do: I've rebased this on the latest `main`, but we also need this to be a single commit. Can you please squash the two commits in a single one? Thanks a lot!

chrisjohnson00:
> @chrisjohnson00 last thing to do: I've rebased this on the latest `main`, but we also need this to be a single commit. Can you please squash the two commits in a single one? Thanks a lot!

Done!

PR #5100, opened 2024-01-06
Enable linting changed files with textlint

ferrarimarco:
# Proposed changes

textlint was expensive to run because we added every file in the list of files to lint to `FILE_ARRAY_NATURAL_LANGUAGE`. In #5041, we mitigated this issue but lost the ability to run textlint on changed files only. Given that textlint ignores files for which it doesn't have a plugin installed, and that we don't currently install additional plugins besides the default ones to lint markdown files and text files, we let textlint run on these files only, so we can have the feature to lint only changed files with this linter as well, back.

## Readiness checklist

### Pull request author tasks


zkoppert (approved):
Great cleanup!

PR #5123, opened 2024-01-10
Simplify worker and linterVersions

ferrarimarco:
# Proposed changes

- Remove the SKIP_FLAG variable and check for the length of the arrays of files to lint directly.
- Remove the LIST_FILES variable, and use the FILE_ARRAY variable directly.
- Remove the corner case for RENOVATE because renovate-config-validator supports passing the path to the file to lint using an argument as the default case does.
- Remove the corner case for ANSIBLE not having 'bad' tests because it has them now.
- Set TF_DATA_DIR to avoid any modification to any existing Terraform data directory that users might have in their workspace.
- Aggregate GO_MODULES and ANSIBLE corner cases because they are the same.
- Remove the corner case for ANSIBLE to add a trailing slash to TEST_CASE_FOLDER (similar reason as the previous point about ANSIBLE corner case).
- Simplify log messages by removing color markers because they are already handled in log.sh.
- Simplify linterVersions by removing redundant checks and functions.
- Avoid printing debug logs in the versions file.

## Readiness checklist

### Pull request author tasks


ferrarimarco:
@zkoppert @Hanse00 the diff on this one is large because removing `SKIP_FLAG` led to indentation changes. The changes should be straightforward. Thanks for reviewing!

zkoppert (approved):

PR #5240, opened 2024-02-06
Create mypy cache directory

ferrarimarco:
# Proposed changes

MyPy expects the cache directory to be there. We now remove it after each super-linter run to avoid leftovers, so we need to create it before running MyPy.

See https://github.com/python/mypy/issues/10768
and https://github.com/python/mypy/issues/10863

## Readiness checklist

### Pull request author tasks

### DIFF
--- a/lib/functions/detectFiles.sh
+++ b/lib/functions/detectFiles.sh
@@ -314,6 +314,13 @@ function RunAdditionalInstalls() {
     fi
   fi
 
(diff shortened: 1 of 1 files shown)


wolf-michael-bolle:
This bug only happens in certain scenarios when multiple Python files are linted but not in all cases that involve multiple Python files. In my case, it all revolved around a single module with utilities in an admittedly pointless subdirectory. I moved the file and gave it a less generic name which got it to work. Still, with super-linter 5.0.0, this did not happen.

ferrarimarco:
Yeah, this was probably caused by #5210

PR #5252, opened 2024-02-10
Rely on default pylint configuration

### LINKED ISSUE #5220: Outdated pylint config value for overgeneral-exceptions

### Current Behavior

```
2024-02-02 21:53:28 [INFO]   Command output for PYTHON_PYLINT:
------
pylint: Command line or configuration file:1: UserWarning: 'Exception' is not a proper value for the 'overgeneral-exceptions' option. Use fully qualified name (maybe 'builtins.Exception' ?) instead. This will cease to be checked at runtime when the configuration upgrader is released.
------
```

Getting the above error, this seems to be because of this config file entry in the docker image...

/action/lib/.automation/.python-lint: overgeneral-exceptions=Exception

It wants that to be defined as builtins.Exception instead. That will need to be corrected, otherwise pylint will error out like this each time it runs via the docker image.

### Expected Behavior

Should work without erroring.

### Super-Linter version

```markdown
I can't figure out how to get the version from inside the container and it doesn't seem to output
it when running. It's either 6.0.0 or 5.7.2. I pulled it down yesterday (02/01/2024) as the latest
docker image, so whatever version that would be.
```

### Relevant log output

```shell
2024-02-02 21:53:28 [INFO]   Command output for PYTHON_PYLINT:
------
pylint: Command line or configuration file:1: UserWarning: 'Exception' is not a proper value for the 'overgeneral-exceptions' option. Use fully qualified name (maybe 'builtins.Exception' ?) instead. This will cease to be checked at runtime when the configuration upgrader is released.
------
```

### Steps To Reproduce

Run the latest version of super-linter against some python code.

### Anything else?

_No response_

ferrarimarco:
# Proposed changes

Remove configuration values from pylint configuration file in case they match with their defaults.

Fix #5220.

## Readiness checklist

### Pull request author tasks


Hanse00:
On one hand, I see the point of deduplication. On the other, I can imagine the argument for creating this in the first place might have been: Relying on defaults also means they can change underneath you without noticing.

I don't have a strong leaning towards either. @zkoppert any input you want to provide?

ferrarimarco:
Thanks for the review!

In the case of Pylint, anything that users don't set, falls back on default values. The configuration file I removed configured values that matched with default ones, leading to duplication. That's why I removed most values.

If we keep this file, we would need to maintain it as the pylint configuration evolves. While working on this, I realized that some values that we set included obsolete values, while some others were moved to different sections. I think there's relatively little value in keeping a configuration file that duplicates the default values when the tool already provides a fallback mechanism to default value.

Hanse00 (approved):
Works for me, if Zack feels otherwise we can always revert.

PR #5288, opened 2024-02-15
Fix GitHub Actions debug logging

ferrarimarco:
# Proposed changes

Enable debug logging when one (or more) of the following conditions is
true:

- ACTIONS_RUNNER_DEBUG is set to true
- ACTIONS_STEPS_DEBUG is set to true
- RUNNER_DEBUG is set to 1

To test this change, I added:

- Three new make targets: one for each of the variables mentioned above.
- One new build step in the `CI` workflow to test the local action without enabling debug logging. By re-running the workflow with GitHub Actions debug mode enabled, it shows debug messages as expected. Example: https://github.com/super-linter/super-linter/actions/runs/7921109078/job/21627279501?pr=5288#step:10

Remove `ACTIONS_RUNNER_DEBUG` from the README so we point users to `LOG_LEVEL` as the preferred mechanism to configure the log level.

## Readiness checklist

### Pull request author tasks


Hanse00 (approved):
Makes sense to me. We still want to support those parameters internally since GH's official docs point to them, but that doesn't mean we need to expose them directly as the preferred debugging strategy.

Issue #5070, opened 2023-12-26
Support Arm64


denisa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When the image executes on an ARM64 architecture, docker warns that it is not a native image.

In my usage, the execution fails before completing the scan of my project

### Expected Behavior

Docker reports that the image supports (at a minimum) amd64 and arm64 architectures.

### Anything else?

Continuation of #1956

ferrarimarco:
Hi @denisa! Thanks for your research on this.

Some of the linters we install have multi-architecture images, so `buildx` should be able to grab the correct one for those, IIRC.

There are some linters that don't offer this at all tho, and some that require glibc.

It would be interesting to experiment with a Debian-based image, instead of an Alpine one to see if things would be easier in that case.

denisa:
my pleasure — I was waiting on the result of the discussion about collaborating with megalinter (#4591 ) to resume work on this ;-)

One big reason to experiment with an other linux version might be PowerShell, for which I only found ubuntu-based images (https://hub.docker.com/_/microsoft-powershell) with supports for both and and arm.

YaroShkvorets:
I can't even pull the image for arm64 on Mac M1.

```
❯ docker pull ghcr.io/super-linter/super-linter:latest
latest: Pulling from super-linter/super-linter
no matching manifest for linux/arm64/v8 in the manifest list entries
```

Works if I pass `--platform linux/amd64` flag though and linters seem to run fine.

ferrarimarco:
Hi @YaroShkvorets ! There's no arm64 image yet. I think you're likely running the amd64 one under emulation, so I would expect some performance degradation.

tchia04:
+1

cjlapao:
On the M2 it seems to be having an issue even with the rosetta enabled

2024-02-12 11:25:33 [INFO]   Stderr when building the file list:
------
/bin/sh: BuildFileArrays: not found
/bin/sh: BuildFileArrays: not found
/bin/sh: BuildFileArrays: not found
/bin/sh: BuildFileArrays: not found
/bin/sh: BuildFileArrays: not found

ryanmerolle:
would love this

ferrarimarco:
Hi folks! Thanks for expressing your interest in this feature.

Adding a build target would not be enough, unfortunately.

[...]

PR #5317, opened 2024-02-23
Install zlib from alpine packages

harryzcy:
# Proposed changes

Running `install-glibc.sh` failed in ARM64, because arch linux packages doesn't offer a build for `arm64`. i.e. `https://www.archlinux.org/packages/core/arm64/zlib/download` returns a 404.

Related to #5070

## Readiness checklist

### Pull request author tasks

### DIFF
--- a/scripts/install-glibc.sh
+++ b/scripts/install-glibc.sh
@@ -2,19 +2,6 @@
 
 set -euo pipefail
 
-case $TARGETARCH in
-amd64)
-  target=x86_64
-  ;;
-arm64)
-  target=arm64
-  ;;
-*)
-  echo "$TARGETARCH is not supported"
-  exit 1
-  ;;
-esac
-
 url=$(
   set -euo pipefail
   curl -s \
@@ -31,14 +18,9 @@ apk add --no-cache --force-overwrite "glibc-${GLIBC_VERSION}.apk"
 rm "glibc-${GLIBC_VERSION}.apk"
 
 # Install zlib
-mkdir /tmp/libz
-
 apk add --no-cache --virtual .glibc-build-deps \
   tar \
+  zlib \
   zstd
-
-curl --retry 5 --retry-delay 5 -sL https://www.archlinux.org/packages/core/${target}/zlib/download | tar -x --zstd -C /tmp/libz
-mv /tmp/libz/usr/lib/libz.so* /usr/glibc-compat/lib
-rm -rf /tmp/libz
 
 apk del --no-network --purge .glibc-build-deps


scripts/install-glibc.sh
```diff
@@ -33,12 +20,7 @@ rm "glibc-${GLIBC_VERSION}.apk"
 # Install zlib
 mkdir /tmp/libz
 
-apk add --no-cache --virtual .glibc-build-deps \
```
ferrarimarco, line 36: Can you leave the `.glibc-build-deps` identifier? We may need it in case we decide to remove build-time deps.

scripts/install-glibc.sh
```diff
@@ -33,12 +20,7 @@ rm "glibc-${GLIBC_VERSION}.apk"
...
   tar \
-  zstd
-
-curl --retry 5 --retry-delay 5 -sL https://www.archlinux.org/packages/core/${target}/zlib/download | tar -x --zstd -C /tmp/libz
-mv /tmp/libz/usr/lib/libz.so* /usr/glibc-compat/lib
-rm -rf /tmp/libz
-
-apk del --no-network --purge .glibc-build-deps
+  zstd \
+  zlib-dev
```
ferrarimarco, line 26: Is there a non `dev` version of this package that's hopefully smaller?

ferrarimarco (changes requested):
Thanks for this PR!

There are a couple of things to check.

Let's see what the CI says :)

scripts/install-glibc.sh
```diff
@@ -33,12 +20,7 @@ rm "glibc-${GLIBC_VERSION}.apk"
 # Install zlib
 mkdir /tmp/libz
```
ferrarimarco, line 21: We don't need this anymore, I suppose.

scripts/install-glibc.sh
```diff
@@ -33,12 +20,7 @@ rm "glibc-${GLIBC_VERSION}.apk"
...
+apk add --no-cache --virtual \
   tar \
-  zstd
-
-curl --retry 5 --retry-delay 5 -sL https://www.archlinux.org/packages/core/${target}/zlib/download | tar -x --zstd -C /tmp/libz
-mv /tmp/libz/usr/lib/libz.so* /usr/glibc-compat/lib
-rm -rf /tmp/libz
-
-apk del --no-network --purge .glibc-build-deps
+  zstd \
```
ferrarimarco, line 25: Can you please keep alphabetical ordering for these packages?

scripts/install-glibc.sh
```diff
@@ -2,19 +2,6 @@
 
 set -euo pipefail
 
-case $TARGETARCH in
```
ferrarimarco, line 5: Are you sure that this glibc distribution is compatible with arm64?

harryzcy: It installs with no errors on arm64

ferrarimarco: Does it run tho? What is the result of `make test-linters` ? :)

harryzcy: Thanks for reporting. From https://github.com/sgerrand/docker-glibc-builder, it looks like arm64 is not supported.. I'll try to find another solution

harryzcy: Is it ok to use the [`gcompat`](https://pkgs.alpinelinux.org/packages?name=gcompat&branch=edge&repo=&arch=&maintainer=) compatibility layer, instead of the actual glibc? Otherwise, I will try to rebuild one from [docker-glibc-builder](https://github.com/sgerrand/docker-glibc-builder)

ferrarimarco: Hi! We didn't try using `gcompat`, but it's worth trying, I suppose! If it works, that would make the build easier to maintain because we can just install that from the Alpine package repository. I would suggest trying this in another PR, and revert the this change (but keep the `zlib` changes), so we can go ahead and merge this.

WDYT?

harryzcy: > I would suggest trying this in another PR

Sent #5334 

> revert this particular change (but keep the rest of changes about `zlib`)

I didn't add the change here

ferrarimarco: I mean: can you please revert the change that checks for platforms? Thanks!

and thanks for the new PR!

ferrarimarco:
@harryzcy after looking at the linter failure (see below), I think we should remove the platform check `switch`, as you did initially :)

```
  In /github/workspace/scripts/install-glibc.sh line 10:
    target=arm64
    ^----^ SC2034 (warning): target appears unused. Verify use (or export if used externally).
```

harryzcy:
> @harryzcy after looking at the linter failure (see below), I think we should remove the platform check `switch`, as you did initially, you were right! :)
> 
> ```
>   In /github/workspace/scripts/install-glibc.sh line 10:
>     target=arm64
>     ^----^ SC2034 (warning): target appears unused. Verify use (or export if used externally).
> ```

@ferrarimarco Fixed.

ferrarimarco (approved):

PR #5320, opened 2024-02-24
Support arm64

### LINKED ISSUE #5070: Support Arm64

harryzcy:
# Proposed changes

Linter commented out are not supported in arm64 yet.

Fix #5070

## Readiness checklist

### Pull request author tasks


ferrarimarco:
Thanks for this PR!

@harryzcy instead of commenting things as a first step, I would start from making the `ci` and `cd` GitHub Actions workflow build the container image for the `arm64` platform as well, so you can get continuous feedback.

scripts/linterVersions.sh
```diff
@@ -77,15 +76,22 @@ LINTER_NAMES_ARRAY['TYPESCRIPT_STANDARD']="ts-standard"
...
-  LINTER_NAMES_ARRAY['POWERSHELL']="pwsh"
   LINTER_NAMES_ARRAY['RUST_2015']="rustfmt"
   LINTER_NAMES_ARRAY['RUST_2018']="rustfmt"
   LINTER_NAMES_ARRAY['RUST_2021']="rustfmt"
   LINTER_NAMES_ARRAY['RUST_CLIPPY']="clippy"
+
+  if [[ "${TARGETARCH}" == "amd64" ]]; then
+    LINTER_NAMES_ARRAY['ENV']="dotenv-linter"
+    LINTER_NAMES_ARRAY['POWERSHELL']="pwsh"
+  fi
```
ferrarimarco, line 94: These two linters are only available in the `standard` image. I'm wondering if we can start with the `slim` image only, instead of having to account for these cases here and in the Dockerfile.

Dockerfile
```diff
@@ -20,13 +20,13 @@ FROM koalaman/shellcheck:v0.9.0 as shellcheck
 FROM mstruebing/editorconfig-checker:2.7.2 as editorconfig-checker
 FROM mvdan/shfmt:v3.8.0 as shfmt
 FROM rhysd/actionlint:1.6.27 as actionlint
-FROM scalameta/scalafmt:v3.8.0 as scalafmt
+FROM scalameta/scalafmt:v3.8.0 as scalafmt-source
 FROM zricethezav/gitleaks:v8.18.2 as gitleaks
 FROM yoheimuta/protolint:0.47.6 as protolint
-FROM ghcr.io/clj-kondo/clj-kondo:2024.02.12-alpine as clj-kondo
+FROM ghcr.io/clj-kondo/clj-kondo:2024.02.13-SNAPSHOT as clj-kondo
```
ferrarimarco, line 26: I don't know if Dependabot is going to pick this up for updates, after we change it. Hopefully yes?

Do we know if clj-kondo folks are going to release this soon-ish?

dependencies/python/build-venvs.sh
```diff
@@ -24,6 +34,14 @@ mkdir -p /venvs
 ########################################
 pip install virtualenv
 
+if [[ "${TARGETARCH}" != "amd64" ]]; then
+  # Install Rust compiler (required by checkov on arm64) #
+  # remove this once https://github.com/bridgecrewio/checkov/pull/6045 is merged
+  apk add --no-cache curl
+  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+  export PATH=$PATH:$HOME/.cargo/bin
+fi
```
ferrarimarco, line 43: Let's see if that Checkov PR lands in, so we can skip this part.

dependencies/python/build-venvs.sh
```diff
@@ -8,11 +8,21 @@
...
+)
+
+if [[ "${TARGETARCH}" != "amd64" ]]; then
+  # libffi-dev is required for building wheel for cffi,
+  # until https://github.com/python-cffi/cffi/issues/69 is merged
+  packages+=(libffi-dev)
+fi
+
+apk add --no-cache --virtual .python-build-deps \
+  "${packages[@]}"
```
ferrarimarco, line 25: Let's see if that cffi PR lands, so we can skip this change.

scripts/linterVersions.sh
```diff
@@ -58,7 +58,6 @@ LINTER_NAMES_ARRAY['R']="R"
 LINTER_NAMES_ARRAY['RAKU']="raku"
 LINTER_NAMES_ARRAY['RENOVATE']="renovate-config-validator"
 LINTER_NAMES_ARRAY['RUBY']="rubocop"
-LINTER_NAMES_ARRAY['SCALAFMT']="scalafmt"
```
ferrarimarco, line 61: This one is going to be a bit more challenging. Is `scalafmt` not available on arm64?

scripts/install-glibc.sh
```diff
@@ -2,20 +2,7 @@
...
-    "https://api.github.com/repos/sgerrand/alpine-pkg-glibc/releases/tags/${GLIBC_VERSION}" |
-    jq --arg name "glibc-${GLIBC_VERSION}.apk" -r '.assets | .[] | select(.name | contains($name)) | .url'
-)
-curl --retry 5 --retry-delay 5 -sL -o "glibc-${GLIBC_VERSION}.apk" \
-  -H "Accept: application/octet-stream" \
-  -H "Authorization: Bearer $(cat /run/secrets/GITHUB_TOKEN)" \
-  "${url}"
-apk add --no-cache --force-overwrite "glibc-${GLIBC_VERSION}.apk"
-rm "glibc-${GLIBC_VERSION}.apk"
+apk add --no-cache gcompat
```
ferrarimarco, line 5: We can probably wait for #5334 to land

ferrarimarco (changes requested):

scripts/linterVersions.sh
harryzcy: No, not available yet

Dockerfile
harryzcy: `clj-kondo` has a release. Updated.

ferrarimarco:
@harryzcy thanks for the update!

ferrarimarco:
@harryzcy thanks for all this work.

I think can simplify things a bit.

For the linters that don't support `arm64` yet,  we can  use the `VALIDATE_XXXX` variables instead of conditionally altering the build. For example, we can add a validation function in `validate.sh` to check if `VALIDATE_XXXX` is `true` for the linters that don't support `arm64`, and fail if so. There are a few functions that do similar things in that file that you can use as examples, but I'll be happy to walk you through one to help you.

By doing this, we would achieve a few things:

- We don't have to maintain a conditional build
- We make this explicit to users, so they are not under the impression that files are being linted with those linters, when they are not.

WDYT?

ferrarimarco:
If we go this route, you can likely revert the changes to:

- The two "no-op" scripts for scalafmt and powershell, and related changes in the Dockerfile
- `linterVersions.sh` (I'm not sure about this yet)

harryzcy:
> we can add a validation function in validate.sh to check if VALIDATE_XXXX is true for the linters that don't support arm64, and fail if so

It makes sense to me

> * The two "no-op" scripts for scalafmt and powershell, and related changes in the Dockerfile

Ok for `scalafmt`, but `powershell` install script will fail. `powershell` doesn't have an arm64 docker image, but does have a arm64 binary published. We could probably get the version, and install from https://github.com/PowerShell/PowerShell/releases instead. I'll investigate.

> * `linterVersions.sh` (I'm not sure about this yet)

I believe they are necessary for build to pass. i.e. we can't run the executable and get its version.

harryzcy:
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed in 14 days if no further activity occurs. Thank you for your contributions.
> 
> If you think this pull request should stay open, please remove the `O: stale 🤖` label or comment on the pull request.
> 
> If you're a maintainer, you can stop the bot to mark this issue as stale in the future by adding the `O: backlog 🤖` label`.

bot remove stale

dependencies/python/build-venvs.sh
ferrarimarco: That PR landed, shall we remove this `if`?

harryzcy: Looking at the [logs](https://github.com/super-linter/super-linter/actions/runs/8699957369/job/23859392947), cffi still need to built from source until a release is landed

`#71 659.6   Building wheel for cffi (pyproject.toml): started`

scripts/linterVersions.sh
ferrarimarco: It seems that there's a [scalafmt-macos](https://github.com/scalameta/scalafmt/releases/download/v3.8.1/scalafmt-macos) executable available since a few releases? Maybe that's useful in this case?

scripts/linterVersions.sh
```diff
@@ -79,15 +78,22 @@ LINTER_NAMES_ARRAY['TYPESCRIPT_STANDARD']="ts-standard"
...
 LINTER_NAMES_ARRAY['YAML']="yamllint"
 
+if [[ "${TARGETARCH}" == "amd64" ]]; then
+  LINTER_NAMES_ARRAY['SCALAFMT']="scalafmt"
+fi
+
 if [[ "${IMAGE}" == "standard" ]]; then
   LINTER_NAMES_ARRAY['ARM']="arm-ttk"
   LINTER_NAMES_ARRAY['CSHARP']="dotnet"
-  LINTER_NAMES_ARRAY['ENV']="dotenv-linter"
```
ferrarimarco, line 85: It seems that dotenv-linter has an arm64 image? https://github.com/dotenv-linter/dotenv-linter/pull/533

scripts/linterVersions.sh
```diff
@@ -79,15 +78,22 @@ LINTER_NAMES_ARRAY['TYPESCRIPT_STANDARD']="ts-standard"
...
-  LINTER_NAMES_ARRAY['POWERSHELL']="pwsh"
```
ferrarimarco, line 86: arm64 request tracked in https://github.com/PowerShell/PowerShell-Docker/issues/795

lib/functions/validation.sh
```diff
@@ -25,6 +25,24 @@ function ValidateBooleanConfigurationVariables() {
   ValidateBooleanVariable "YAML_ERROR_ON_WARNING" "${YAML_ERROR_ON_WARNING}"
 }
 
+function ValidatePlatformArchitecture() {
```
ferrarimarco, line 28: I think we shouldn't validate this at runtime, but rather when we build the image. Also, we wouldn't want to silently disable linters in the case that they aren't available on a given platform.

I would revert this change and the related function call in `linter.sh`. We likely need to think a bit more about how to handle this.

Thanks!

lib/linter.sh
```diff
@@ -108,6 +108,7 @@ declare -l YAML_ERROR_ON_WARNING
 YAML_ERROR_ON_WARNING="${YAML_ERROR_ON_WARNING:-false}"
 
 ValidateBooleanConfigurationVariables
+ValidatePlatformArchitecture
```
ferrarimarco, line 111: Can you revert this change as well? See my previous comment.

dependencies/python/build-venvs.sh
```diff
@@ -8,11 +8,21 @@
...
+packages=(
+  gcc
+  linux-headers
+  musl-dev
   python3-dev
+)
+
+if [[ "${TARGETARCH}" != "amd64" ]]; then
+  # libffi-dev is required for building wheel for cffi,
+  # until https://github.com/python-cffi/cffi/issues/69 lands in a release
```
ferrarimarco, line 20: There's a pre-release available. Almost there :)

scripts/linterVersions.sh
harryzcy: Yes, it's actually included since https://github.com/dotenv-linter/dotenv-linter/pull/757

ferrarimarco: Nice! Thanks for the update!

scripts/linterVersions.sh
harryzcy: > It seems that there's a [scalafmt-macos](https://github.com/scalameta/scalafmt/releases/download/v3.8.1/scalafmt-macos) executable available since a few releases? Maybe that's useful in this case?

No, macos is different and I tested that it can't run on `linux/arm64`

scripts/linterVersions.sh
```diff
@@ -60,7 +60,6 @@ LINTER_NAMES_ARRAY['R']="R"
 LINTER_NAMES_ARRAY['RAKU']="raku"
 LINTER_NAMES_ARRAY['RENOVATE']="renovate-config-validator"
 LINTER_NAMES_ARRAY['RUBY']="rubocop"
-LINTER_NAMES_ARRAY['SCALAFMT']="scalafmt"
```
ferrarimarco, line 63: Can you revert this change? I think we'll need to find a way around this when we build the image instead of excluding this from the linters.

Example: build `scalafmt` when on `arm64`? IDK, we'll see later.

scripts/linterVersions.sh
```diff
@@ -79,6 +78,10 @@ LINTER_NAMES_ARRAY['TYPESCRIPT_STANDARD']="ts-standard"
 LINTER_NAMES_ARRAY['XML']="xmllint"
 LINTER_NAMES_ARRAY['YAML']="yamllint"
 
+if [[ "${TARGETARCH}" == "amd64" ]]; then
+  LINTER_NAMES_ARRAY['SCALAFMT']="scalafmt"
+fi
+
```
ferrarimarco, line 84: Part of the change revert mentioned above. Thanks!

dependencies/python/build-venvs.sh
```diff
@@ -8,11 +8,21 @@
...
+  packages+=(libffi-dev)
+fi
+
+apk add --no-cache --virtual .python-build-deps \
+  "${packages[@]}"
```
ferrarimarco, line 25: Considering that we might need this for other wheels, I think we can revert this change and just add `libffi-dev` to the list of packages without any check. It's not going to have any effect on the final image size because we run this script in a dedicated build stage.

So, we can unblock this part, and don't have to wait for a new `cffi` release.

dependencies/python/build-venvs.sh
ferrarimarco, line 43: Can we install it from the OS package repository, instead of fetching and running a script?

Dockerfile
```diff
@@ -65,11 +65,17 @@ RUN apk add --no-cache \
...
 COPY dependencies/python/ /stage
 WORKDIR /stage
 RUN ./build-venvs.sh && rm -rfv /stage
 
-FROM python:3.12.3-alpine3.19 as npm-builder
+# On arm64, `npm install` gives ECONNRESET errors:
+# ERR! network Client network socket disconnected before secure TLS connection was established
+# 
+# Since we only need node_modules, we can build it on an amd64 image instead
+FROM --platform=$BUILDPLATFORM python:3.12.3-alpine3.19 as npm-builder
```
ferrarimarco, line 78: What if there are platform-specific extensions? Can you please revert this change so we see the build failing, and we can investigate further? Thanks!

.github/workflows/ci.yml
```diff
@@ -92,6 +103,7 @@ jobs:
           cache-from: type=registry,ref=${{ env.CONTAINER_IMAGE_ID }}-buildcache
           outputs: type=docker,dest=/tmp/${{ env.CONTAINER_IMAGE_OUTPUT_IMAGE_NAME }}.tar
           push: false
+          platforms: ${{ matrix.images.platform }}
```
ferrarimarco, line 106: Given that we configure `QEMU`, can't we run this anyway?

(applies to similar `if`s as well.

Dockerfile
```diff
@@ -26,7 +26,7 @@ FROM rhysd/actionlint:1.7.1 as actionlint
 FROM scalameta/scalafmt:v3.8.1 as scalafmt
 FROM zricethezav/gitleaks:v8.18.3 as gitleaks
 FROM yoheimuta/protolint:0.50.2 as protolint
-FROM ghcr.io/clj-kondo/clj-kondo:2024.05.24-alpine as clj-kondo
+FROM ghcr.io/clj-kondo/clj-kondo:2024.05.24 as clj-kondo
```
ferrarimarco, line 29: Why this change?

harryzcy: There's no arm64 build for alpine image, only for glibc based binary.

Reference: see OS / Arch tab:
https://github.com/clj-kondo/clj-kondo/pkgs/container/clj-kondo/220477577?tag=2024.05.24-alpine
https://github.com/clj-kondo/clj-kondo/pkgs/container/clj-kondo/220477405?tag=2024.05.24

.github/workflows/ci.yml
harryzcy: Do you mean set both platforms here? Because it will run amd64 and arm64 in parallel instead of all together, maybe slightly less build time.

I can set both here if you prefer

dependencies/python/build-venvs.sh
harryzcy: done

.github/workflows/ci.yml
ferrarimarco: Yup

ferrarimarco:
@harryzcy https://github.com/npm/cli/issues/4652 seems relevant to the npm connection error.

dependencies/python/build-venvs.sh
harryzcy: done

scripts/linterVersions.sh
harryzcy: > Can you revert this change? I think we'll need to find a way around this when we build the image instead of excluding this from the linters.

This script is executed during build time, making sure that those executables exist. The build would fail if scalafmt is included here

harryzcy:
> @harryzcy [npm/cli#4652](https://github.com/npm/cli/issues/4652) seems relevant to the npm connection error.

Now we have `ubuntu-24.04-arm` github runners, and we can build natively without this problem.

scripts/linterVersions.sh
harryzcy: Looks like `scalafmt` has added arm64 support: https://github.com/scalameta/scalafmt/blob/360eabd1caec41f3d6febf4361f808b1dba4c7e2/.github/workflows/ci.yml#L151. Just not released yet

ferrarimarco:
Thanks for pushing this forward @harryzcy !

ferrarimarco:
Closing this in favor of #7021 because we refactored the workflow and the Dockerfile quite a bit since this PR has been created. We can continue working there. @harryzcy Thanks for pushing this forward, let's collaborate on #7021.

PR #5345, opened 2024-03-02
Find go.mod based on Go files

### LINKED ISSUE #5292: Search for `go.mod` whenever if found a Go file

### Current Behavior

Go modules lints are run only when `go.mod` is included, not when only some Go files are included.

This is strange when `VALIDATE_ALL_CODEBASE` is set to `true`. A Go module does exist but it's not identified because `go.mod` is not updated.

### Expected Behavior

Go modules lints should run when `VALIDATE_ALL_CODEBASE: false` and some Go files are updated.

### Anything else?

_No response_

harryzcy:
# Proposed changes

Close #5292

## Readiness checklist

### Pull request author tasks

- [ ] I included all the needed documentation for this change.

### DIFF
--- a/lib/functions/buildFileList.sh
+++ b/lib/functions/buildFileList.sh
@@ -320,21 +320,6 @@ BuildFileArrays() {
       fi
     done
 
-    if [ "${BASE_FILE}" == "go.mod" ]; then
-      debug "Found ${FILE}. Checking if individual Go file linting is enabled as well."
-      if [ "${VALIDATE_GO}" == "true" ]; then
-        debug "Checking if we are running tests. TEST_CASE_RUN: ${TEST_CASE_RUN}"
-        if [ "${TEST_CASE_RUN}" == "true" ]; then
-          debug "Skipping the failure due to individual Go files and Go modules linting being enabled at the same time because we're in test mode."
-        else
-          fatal "Set VALIDATE_GO to false to avoid false positives due to analyzing Go files in the ${FILE_DIR_NAME} directory individually instead of considering them in the context of a Go module."
-        fi
-      else
-        debug "Considering ${FILE_DIR_NAME} as a Go module."
-      fi
-      echo "${FILE_DIR_NAME}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO_MODULES"
-    fi
-
     if IsValidShellScript "${FILE}"; then
       echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-BASH"
       echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-BASH_EXEC"
@@ -370,8 +355,42 @@ BuildFileArrays() {
       echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-ENV"
     elif [ "${FILE_TYPE}" == "feature" ]; then
       echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GHERKIN"
-    elif [ "${FILE_TYPE}" == "go" ]; then
-      echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO"
+    elif [ "${FILE_TYPE}" == "go" ] ||
+      [[ "${BASE_FILE}" == "go.mod" ]]; then
+
+      # Check if we should lint a Go module
+      local GO_MOD_FILE_PATH=""
+      if [ "${BASE_FILE}" == "go.mod" ]; then
+        GO_MOD_FILE_PATH="${FILE}"
+        debug "${BASE_FILE} is a Go module file. Setting the go.mod file path to ${GO_MOD_FILE_PATH}"
+      else
+        echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO"
+        debug "${BASE_FILE} is a Go file. Trying to find out if it's part of a Go module"
+        local dir_name
+        dir_name="${FILE_DIR_NAME}"
+        while [ "${dir_name}" != "$(dirname "${GITHUB_WORKSPACE}")" ] && [ "${dir_name}" != "/" ]; do
+          local potential_go_mod_file_path="${dir_name}/go.mod"
+          if [ -f "${potential_go_mod_file_path}" ]; then
+            GO_MOD_FILE_PATH="${potential_go_mod_file_path}"
+            break
+          fi
+          dir_name=$(dirname "${dir_name}")
+        done
+      fi
+
+      if [[ -n "${GO_MOD_FILE_PATH:-}" ]]; then
+        debug "Considering ${FILE_DIR_NAME} as a Go module because it contains ${GO_MOD_FILE_PATH}"
+        local FILE_ARRAY_GO_MODULES_PATH="${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO_MODULES"
+
+        if [[ ! -e "${FILE_ARRAY_GO_MODULES_PATH}" ]] || ! grep -Fxq "${FILE_DIR_NAME}" "${FILE_ARRAY_GO_MODULES_PATH}"; then
+          echo "${FILE_DIR_NAME}" >>"${FILE_ARRAY_GO_MODULES_PATH}"
+          debug "Added ${GO_MOD_FILE_PATH} directory (${FILE_DIR_NAME}) to the list of Go modules to lint"
+        else
+          debug "Skip adding ${GO_MOD_FILE_PATH} directory (${FILE_DIR_NAME}) to the list of Go modules to lint because it's already in that list"
+        fi
+      else
+        debug "${FILE} is not considered to be part of a Go module"
+      fi
     # Use BASE_FILE here because FILE_TYPE is not reliable when there is no file extension
     elif [ "$FILE_TYPE" == "groovy" ] || [ "$FILE_TYPE" == "jenkinsfile" ] ||
       [ "$FILE_TYPE" == "gradle" ] || [ "$FILE_TYPE" == "nf" ] ||


lib/functions/buildFileList.sh
```diff
@@ -368,6 +368,16 @@ BuildFileArrays() {
       echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GHERKIN"
     elif [ "${FILE_TYPE}" == "go" ]; then
       echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO"
+
+      # find go.mod
+      dir_name="${FILE_DIR_NAME}"
+      while [ "${dir_name}" != "$(dirname ${GITHUB_WORKSPACE})" ] && [ "${dir_name}" != "/" ]; do
```
ferrarimarco, line 374: Can you please double-quote `${GITHUB_WORKSPACE}` ?

lib/functions/buildFileList.sh
```diff
@@ -368,6 +368,16 @@ BuildFileArrays() {
...
+        if [ -f "${dir_name}/go.mod" ]; then
+          echo "${dir_name}/go.mod" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO_MODULES"
```
ferrarimarco, line 376: What if we already added this `go.mod` file?

lib/functions/buildFileList.sh
```diff
@@ -368,6 +368,16 @@ BuildFileArrays() {
       echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GHERKIN"
     elif [ "${FILE_TYPE}" == "go" ]; then
       echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO"
+
+      # find go.mod
```
ferrarimarco, line 372: Can you add some `debug` statements so we can follow what's happening?

lib/functions/buildFileList.sh
```diff
@@ -368,6 +368,16 @@ BuildFileArrays() {
...
+      dir_name="${FILE_DIR_NAME}"
```
ferrarimarco, line 373: Can you add `local dir_name` before this assignment?

ferrarimarco:
Thanks for this PR! A few small things to check.

lib/functions/buildFileList.sh
harryzcy: I implemented a solution in 4992fa4, please review it. Thanks

ferrarimarco:
Thanks! A small thing to fix:

```sh
            if [ ! grep -Fxq \\"${go_mod_filename}\\" \\"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO_MODULES\\" ]; then
            ^-- SC1009 (info): The mentioned syntax error was in this if expression.
               ^-- SC1073 (error): Couldn't parse this test expression. Fix to allow more checks.
                                                  ^-- SC1072 (error): Expected test to end here (don't wrap commands in []/[[]]). Fix any mentioned problems and try again.
```

harryzcy:
> Thanks! A small thing to fix:
> 
> ```shell
>             if [ ! grep -Fxq \\"${go_mod_filename}\\" \\"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO_MODULES\\" ]; then
>             ^-- SC1009 (info): The mentioned syntax error was in this if expression.
>                ^-- SC1073 (error): Couldn't parse this test expression. Fix to allow more checks.
>                                                   ^-- SC1072 (error): Expected test to end here (don't wrap commands in []/[[]]). Fix any mentioned problems and try again.
> ```

I have fixed the test

ferrarimarco:
There's an issue, I think:

```
  2024-03-13 21:41:06 [FATAL]   Error when setting INDEX for GO_MODULES: 0
```

https://github.com/super-linter/super-linter/actions/runs/8271680556/job/22632678543?pr=5345

Can you kindly take a look? Thanks!

ferrarimarco:
> There's an issue, I think:
> 
> ```
>   2024-03-13 21:41:06 [FATAL]   Error when setting INDEX for GO_MODULES: 0
> ```
> 
> https://github.com/super-linter/super-linter/actions/runs/8271680556/job/22632678543?pr=5345
> 
> Can you kindly take a look? Thanks!

More logs after rebasing:

```
  2024-04-22 09:49:46 [DEBUG]   PARALLEL_COMMAND_OUTPUT for GO_MODULES (exit code: 255): cd /tmp/lint/test/linters/go_modules/go_modules_good || exit 255; golangci-lint run --allow-parallel-runners -c /tmp/lint/.github/linters/.golangci.yml && echo \"Linted: /tmp/lint/test/linters/go_modules/go_modules_good\"
  parallel: Error: Cannot change into non-executable dir /tmp/lint/test/linters/go_modules/go_modules_good/go.mod: File exists
```

This is due to how we configure the command to run for `GO_MODULES`: https://github.com/super-linter/super-linter/blob/56e675bd333f8af069b2d7ba67ade1f39fcd01c4/lib/functions/worker.sh#L123-L139

So the command becomes:

```sh
parallel ... --workdir {} ...
```

(`{}` means: the input item as it is, which is `/tmp/lint/test/linters/go_modules/go_modules_good/go.mod` because you set it to `echo "${go_mod_filename}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO_MODULES"`).

I'll send another comment with the necessary changes.

Thanks for all your work on this!

ferrarimarco (changes requested):

harryzcy:
@ferrarimarco Thanks for pointing out the issue! It's now updated

ferrarimarco:
Thanks!

We are on the right track!

I think we might still need some refactoring because in the current implementation users will not be able to disable lining of Go modules by using the `VALIDATE_GO_MODULES` variable. Also, we might report false positives because Go files might be linted one by one (`VALIDATE_GO` default mode), and also as a module.

Let me think a bit about this.

lib/functions/buildFileList.sh
```diff
@@ -370,22 +355,43 @@ BuildFileArrays() {
...
+        dir_name="${FILE_DIR_NAME}"
+        local potential_go_mod_file_path="${dir_name}/go.mod"
+        while [ "${dir_name}" != "$(dirname "${GITHUB_WORKSPACE}")" ] && [ "${dir_name}" != "/" ]; do
+          if [ -f "${potential_go_mod_file_path}" ]; then
+            GO_MOD_FILE_PATH="${potential_go_mod_file_path}"
+            break
           fi
-          break
+          dir_name=$(dirname "${dir_name}")
+        done
```
harryzcy, line 377: When walking up the directory, `potential_go_mod_file_path` here would not be changed? It should be updated within the loop for the parent `dir_name`

lib/functions/buildFileList.sh
```diff
@@ -370,22 +355,43 @@ BuildFileArrays() {
...
+        else
+          debug "Skip adding ${GO_MOD_FILE_PATH} directory (${FILE_DIR_NAME}) to the list of Go modules to lint because it's already in that list"
         fi
-        dir_name=$(dirname "${dir_name}")
-      done
+      else
+        debug "${FILE} is not considered to be part of a Go module"
+      fi
+
+      echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO"
```
harryzcy, line 394: Should line 394 be placed at line 391 instead? i.e. only lint individual Go files when it's not part of a Go module. I'm not sure here

harryzcy:
@ferrarimarco Thank you so much! There're two parts that I have doubts. If you agree, I'll make the edits and squash two commits.

lib/functions/buildFileList.sh
ferrarimarco: In this function, we just list and categorize files. Then, it's up to the user to eventually disable linting for individual Go files if they want to avoid false positives due to scanning individual files when there's a Go module. Before this change, we did checks in this regard, but it looked like a workaround and we had to account for a couple of corner cases about the test environment that complicated the logic when it wasn't necessary.

One check that we need anyway is that this should add `${FILE}` to `"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO"` only if it's a `.go` file, because `${FILE}` can be `go.mod`, and we don't want to add it to `file-array-GO`.

lib/functions/buildFileList.sh
ferrarimarco: This is correct. We should update `potential_go_mod_file_path` after updating `dir_name`. Good catch

ferrarimarco:
Thanks for the review @harryzcy ! Your proposal SGTM :)

lib/functions/buildFileList.sh
harryzcy: > One check that we need anyway is that this should add `${FILE}` to `"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO"` only if it's a `.go` file, because `${FILE}` can be `go.mod`, and we don't want to add it to `file-array-GO`.

Updated in https://github.com/super-linter/super-linter/pull/5345/commits/cb8ebd42c9805a9f8b97bfacad0238be87e2d014

lib/functions/buildFileList.sh
harryzcy: Fixed in https://github.com/super-linter/super-linter/pull/5345/commits/cb8ebd42c9805a9f8b97bfacad0238be87e2d014

harryzcy:
I just squashed two commits to become https://github.com/super-linter/super-linter/pull/5345/commits/4637a4ec70525882b1fdee2ccb977e2569d66180

lib/functions/buildFileList.sh
```diff
@@ -370,8 +355,45 @@ BuildFileArrays() {
...
+        else
+          debug "Skip adding ${GO_MOD_FILE_PATH} directory (${FILE_DIR_NAME}) to the list of Go modules to lint because it's already in that list"
+        fi
+      else
+        debug "${FILE} is not considered to be part of a Go module"
+      fi
+
+      if [[ "${BASE_FILE}" != "go.mod" ]]; then
+        echo "${FILE}" >>"${FILE_ARRAYS_DIRECTORY_PATH}/file-array-GO"
+      fi
```
ferrarimarco, line 396: Minor refactoring question. Do we want to move this echo in the else clause (L366) of the if if statement (L363)? That would perhaps make things clearer because we already have the relevant debug statements. WDYT?

ferrarimarco: @harryzcy

ferrarimarco: done

harryzcy: thanks

PR #5452, opened 2024-03-26
Set permissions in the README example

mohamedhamdyamer:
adding:
permissions: read-all

This is to avoid failure of super-linter that is triggered by `checkov`.

`checkov` flags that the top-level permissions are set to "write-all" (default value). Accordingly, the action will fail with an error.

# Proposed changes

Small addition to the yaml example as described above

## Readiness checklist

### Pull request author tasks

- [ ] I followed the [Conventional Commit v1.0.0 spec](https://www.conventionalcommits.org/en/v1.0.0/).
- [ ] If this pull request is about and existing issue,

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,8 @@ To run super-linter as a GitHub Action, you do the following:
       push: null
       pull_request: null
 
+    permissions: { }
+
     jobs:
       build:
         name: Lint


README.md
```diff
@@ -119,6 +119,8 @@ To run super-linter as a GitHub Action, you do the following:
       push: null
       pull_request: null
 
+    permissions: read-all # this is to avoid failure of super-linter that is triggered by checkov --> flagging that the top-level permissions are set to write-all (default value).
```
ferrarimarco, line 122: Can you remove the comment? It's an implementation detail :)

ferrarimarco (changes requested):
Thanks for this PR! A small thing to check.

README.md
mohamedhamdyamer: sure

mohamedhamdyamer: Done,

Could you check, please?

ferrarimarco (approved):

README.md
```diff
@@ -119,6 +119,8 @@ To run super-linter as a GitHub Action, you do the following:
       push: null
       pull_request: null
 
+    permissions: read-all
```
ferrarimarco, line 122: Can you please refactor this as `permissions: {}` instead?

We don't need the `read-all` permission.

Thanks!

ferrarimarco (changes requested):

README.md
mohamedhamdyamer: sure

mohamedhamdyamer: Done

ferrarimarco (approved):

PR #4925, opened 2023-11-30
Add support for Checkov

ferrarimarco:
Fixes #694

## Proposed Changes

- Add support to run [Checkov](https://github.com/bridgecrewio/checkov) against infrastructure as code descriptors that are in a given (configurable) directory. Defaults to lint the whole workspace.
- Fix a few findings that Checkov reported.
- Establish a baseline for our own codebase so we don't have to fix issues right away with this change potentially causing breaking changes.

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [x] Label as `breaking` if this is a large fundamental change
- [x] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #2821, opened 2022-04-23
Disable flake8 rule E203

massongit:
Fixes #2757

## Proposed Changes

I disable `flake8` rule `E203` because this rule conflicts with `black`.
FYI: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8

### Author/Contributor
- [x] If documentation is needed for this change, has that been included in this pull request

### Reviewing Maintainer
- [ ] Label as `breaking` if this is a large fundamental change
- [ ] Label as either `automation`, `bug`, `documentation`, `enhancement`, `infrastructure`, or `performance`

PR #5456, opened 2024-04-01
Add support for astral-sh/ruff

### LINKED ISSUE #5384: Add Ruff

### Current Behavior

[Ruff] can't be used for Python linting.

### Expected Behavior

[Ruff] can be used for Python linting.

### Anything else?

[Ruff]: https://github.com/astral-sh/ruff

Okabe-Junya:
Fix #5384 

# Proposed changes

- add support for [astral-sh/ruff](https://github.com/astral-sh/ruff)

## Readiness checklist

### Pull request author tasks


.github/linters/.ruff.toml
```diff
@@ -0,0 +1,5 @@
+# Same as .flake8
```
ferrarimarco, line 1: This file is not needed because it's the same as the one in TEMPLATES

TEMPLATES/.ruff.toml
```diff
@@ -0,0 +1,5 @@
+# Same as .flake8
+line-length = 120
+
+[lint]
+ignore = ["E203"]
```
ferrarimarco, line 5: Why disable this one?

ferrarimarco (changes requested):
Thanks for this PR! A couple of minor things to check. I'll also run CI workflows, and see what they say :)

TEMPLATES/.ruff.toml
Okabe-Junya: In super-linter, E203 made disabled in #2821. I think that when supporting ruff, it should not contradict other tools that are already supported, such as black and flake8.

This means that the rule was disabled in order to avoid errors by ruff check (on code formatted with black)

Okabe-Junya:
Thank you @ferrarimarco for reviewing!!

I fixed CI error and commented. PTAL again, thank you 💙

TEMPLATES/.ruff.toml
ferrarimarco: Makes sense.

README.md
```diff
@@ -230,6 +230,7 @@ You can configure super-linter using the following environment variables:
...
 | **PYTHON_MYPY_CONFIG_FILE**                     | `.mypy.ini`                     | Filename for [mypy configuration](https://mypy.readthedocs.io/en/stable/config_file.html) (ex: `.mypy.ini`, `setup.config`)                                                                                          |
 | **PYTHON_PYLINT_CONFIG_FILE**                   | `.python-lint`                  | Filename for [pylint configuration](https://pylint.pycqa.org/en/latest/user_guide/run.html?highlight=rcfile#command-line-options) (ex: `.python-lint`, `.pylintrc`)                                                  |
+| **PYTHON_RUFF_CONFIG_FILE**                     | `.ruff.toml`                    | Filename for [ruff configuration](https://docs.astral.sh/ruff/configuration/) (ex: `.ruff.toml`)                                                                                                                    |
```
ferrarimarco, line 233: Can you remove `(ex: `.ruff.toml`) `? Not needed :)

Okabe-Junya: deleted!!

lib/linter.sh
```diff
@@ -292,7 +294,8 @@ LANGUAGE_ARRAY=('ANSIBLE' 'ARM' 'BASH' 'BASH_EXEC' 'CHECKOV' 'CLANG_FORMAT'
   'KUBERNETES_KUBECONFORM' 'KOTLIN' 'LATEX' 'LUA' 'MARKDOWN'
   'NATURAL_LANGUAGE' 'OPENAPI' 'PERL' 'PHP_BUILTIN' 'PHP_PHPCS' 'PHP_PHPSTAN'
   'PHP_PSALM' 'POWERSHELL' 'PROTOBUF' 'PYTHON_BLACK' 'PYTHON_PYLINT'
-  'PYTHON_FLAKE8' 'PYTHON_ISORT' 'PYTHON_MYPY' 'R' 'RAKU' 'RENOVATE' 'RUBY' 'RUST_2015'
+  'PYTHON_FLAKE8' 'PYTHON_ISORT' 'PYTHON_MYPY' 'PYTHON_RUFF'
```
Okabe-Junya, line 297: I missed adding `'PYTHON_RUFF'` in this section, so now I believe **ALL** CI is passing ...

ferrarimarco (changes requested):
@Okabe-Junya we're almost there!

As a last change, can you please make the tests shorter? There's likely no need to include these big files. Maybe 3-4 lines each are enough.

Thanks!

Okabe-Junya:
Thank you @ferrarimarco

Your review is very sensible. However, other Python tools are (mostly) using the same test files. I think we should not prepare a completely different test file just for ruff now.

FYI: 

- black: https://github.com/super-linter/super-linter/tree/main/test/linters/python_black
- pylint: https://github.com/super-linter/super-linter/tree/main/test/linters/python_pylint
- flake8: https://github.com/super-linter/super-linter/tree/main/test/linters/python_flake8
- mypy: https://github.com/super-linter/super-linter/tree/main/test/linters/python_mypy

Meaning, if we are to change to test files of about 3-4 lines, shouldn't we do the same for the other files as well? And, I think this change is beyond the scope of this PR.

thank you 💙

ferrarimarco:
> Thank you @ferrarimarco
> 
> Your review is very sensible. However, other Python tools are (mostly) using the same test files. I think we should not prepare a completely different test file just for ruff now.
> 
> FYI:
> 
> * black: https://github.com/super-linter/super-linter/tree/main/test/linters/python_black
> * pylint: https://github.com/super-linter/super-linter/tree/main/test/linters/python_pylint
> * flake8: https://github.com/super-linter/super-linter/tree/main/test/linters/python_flake8
> * mypy: https://github.com/super-linter/super-linter/tree/main/test/linters/python_mypy
> 
> Meaning, if we are to change to test files of about 3-4 lines, shouldn't we do the same for the other files as well? And, I think this change is beyond the scope of this PR.
> 
> thank you 💙

Hi! There's no need for the test files to be aligned. In fact, we should reduce the size of the others as well, but this is out of scope for this PR, as you said. We can start with these ones, tho.

Thanks for your contributions!

Okabe-Junya:
Thanks!!

> but this is out of scope for this PR, as you said. We can start with these ones, tho.

I would like to work on updating the test files after this PR is merged :)

First, I will focus on merging this PR. I think that addressing the CI errors is what needs to be done. I will investigate the cause of the CI failures...

ferrarimarco:
Sorry, I mistyped. I meant: we can start with the ones in this PR. :)

thanks a lot!

Okabe-Junya (commented):
I have investigated similar PRs to resolve CI error. #4925 seems to be one of them.

Upon careful investment, I realized that I did not write code like `LINTER_NAMES_ARRAY['CHECKOV']="checkov"`. However, such code does not exist in `lib/linter.sh`.

When I used "grep" to check the code, I found that it had been moved to `scripts/linterVersions.sh` in #5197. This means that I need to make a simple modification to `scripts/linterVersions.sh`.

Okabe-Junya:
I believe now **ALL** CI will pass...

scripts/linterVersions.sh
```diff
@@ -54,6 +54,7 @@ LINTER_NAMES_ARRAY['PYTHON_PYLINT']="pylint"
 LINTER_NAMES_ARRAY['PYTHON_FLAKE8']="flake8"
 LINTER_NAMES_ARRAY['PYTHON_ISORT']="isort"
 LINTER_NAMES_ARRAY['PYTHON_MYPY']="mypy"
+LINTER_NAMES_ARRAY['PYTHON_RUFF']="ruff"
```
ferrarimarco, line 57: I think you may need to add the `ruff --version` special case below. The default command is `LINTER_NAME -v`.

test/linters/python_ruff/python_bad_1.py
```diff
@@ -0,0 +1,156 @@
+import json
```
ferrarimarco, line 1: As we mentioned, can you please reduce this file to be smaller? 3-4 lines should be enough. It's ok if it's not aligned to the test files we're using for other Python linters. Actually, we don't want that dependency. :)

Thanks!

test/linters/python_ruff/python_good_1.py
```diff
@@ -0,0 +1,201 @@
+import json
```
ferrarimarco, line 1: As we mentioned, can you please reduce this file to be smaller? 3-4 lines should be enough. It's ok if it's not aligned to the test files we're using for other Python linters. Actually, we don't want that dependency. :)

Thanks!

scripts/linterVersions.sh
Okabe-Junya: https://github.com/super-linter/super-linter/blob/dec3f48f49948d292f99953db549811f65e762c9/test/inspec/super-linter/controls/super_linter.rb#L186

I think it can be configured in this section, so I don't need to change..

test/linters/python_ruff/python_bad_1.py
Okabe-Junya: My apologies,  I misunderstood a bit. Ok, I'll fix it :)

Okabe-Junya: I fixed it!!

test/linters/python_ruff/python_good_1.py
Okabe-Junya: I fixed it!!

Okabe-Junya:
I have fixed to the test files, and all CI checks have passed in the forked repo!!

FYI: https://github.com/Okabe-Junya/super-linter/actions/runs/8568038203

Screenshot 2024-04-05 at 19 18 04

scripts/linterVersions.sh
ferrarimarco: The one in `super_linter.rb` is for running tests. The one I'm referring to is to generate the versions file. So, you need both :)

Thanks!

ferrarimarco: Ok, inspected `linterVersions.sh` once more. There's no need to add a special case here because the default is to use the `--version` switch.

ferrarimarco:
One small thing to check and we should be good to go! Thanks for all this work @Okabe-Junya !

Okabe-Junya:
Thank you for reviewing this many times. I have checked!

ferrarimarco (approved):

Okabe-Junya:
Thank you for your Considerate and kind support 💙 Looking forward to release!!

test/linters/python_ruff/python_good_1.py
Okabe-Junya: created #5501, thank you!!

PR #5631, opened 2024-05-09
Include Helm for checkov scans

andrewkreuzer:
# Proposed changes

Includes Helm for the running of checkov scans with the helm framework

## Readiness checklist

### Pull request author tasks

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ FROM alpine/terragrunt:1.8.4 as terragrunt
 FROM dotenvlinter/dotenv-linter:3.3.0 as dotenv-linter
 FROM ghcr.io/terraform-linters/tflint:v0.51.1 as tflint
 FROM ghcr.io/yannh/kubeconform:v0.6.6 as kubeconfrm
+FROM alpine/helm:3.14.4 as helm
 FROM golang:1.22.3-alpine as golang
 FROM golangci/golangci-lint:v1.59.0 as golangci-lint
 FROM goreleaser/goreleaser:v1.26.2 as goreleaser
@@ -258,6 +259,11 @@ COPY scripts/install-google-java-format.sh /
 COPY dependencies/google-java-format /google-java-format
 RUN --mount=type=secret,id=GITHUB_TOKEN /install-google-java-format.sh \
     && rm -rfv /install-google-java-format.sh /google-java-format
+
+################
+# Install Helm #
+################
+COPY --from=helm /usr/bin/helm /usr/bin/
 
 # Copy Node tools
 COPY --from=npm-builder /node_modules /node_modules
--- a/test/inspec/super-linter/controls/super_linter.rb
+++ b/test/inspec/super-linter/controls/super_linter.rb
@@ -160,6 +160,7 @@ control "super-linter-installed-commands" do
     { linter_name: "goreleaser"},
     { linter_name: "google-java-format", version_command: "java -jar /usr/bin/google-java-format --version"},
     { linter_name: "hadolint"},
+    { linter_name: "helm", version_option: "version"}, # not used as linter, needed for checkov's helm framework
     { linter_name: "htmlhint"},
     { linter_name: "isort"},
     { linter_name: "jscpd"},
@@ -510,7 +511,8 @@ control "super-linter-validate-files" do
     "/action/lib/.automation/.yaml-lint.yml",
     "/action/lib/.automation/phpcs.xml",
     "/action/lib/.automation/phpstan.neon",
-    "/action/lib/.automation/psalm.xml"
+    "/action/lib/.automation/psalm.xml",
+    "/usr/bin/helm", # needed for checkov's helm framework
   ]
 
   files.each do |item|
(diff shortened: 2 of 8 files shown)


ferrarimarco:
Hi @andrewkreuzer ! Thanks for this PR.

What problem are we solving here? :)

andrewkreuzer:
when running checkov using the "helm" framework I wasn't seeing scans being done for my helm charts. Running locally I was able to get the desired output:
```
helm scan results:
   Passed checks: 233, Failed checks: 0, Skipped checks: 0
```
but wasn't seeing this output with superlinter

Running checkov in the superlinter container `checkov --framework helm --var-file my-values.yaml --directory charts` I was able to get to this error: 
```
2024-05-09 14:17:51,287 [MainThread  ] [ERROR]  There are no runners to run. This can happen if you specify a file type and a framework that are not compatible (e.g., `--file xyz.yaml --framework terraform`), or if you specify a framework with missing dependencies (e.g., helm or kustomize, which require those tools to be on your system). Running with LOG_LEVEL=DEBUG may provide more information.
```

ferrarimarco:
I see, thanks for the details!

Dockerfile
```diff
@@ -259,6 +259,13 @@ COPY dependencies/google-java-format /google-java-format
 RUN --mount=type=secret,id=GITHUB_TOKEN /install-google-java-format.sh \
     && rm -rfv /install-google-java-format.sh /google-java-format
 
+################
+# Install Helm #
+################
+COPY scripts/install-helm.sh /
+RUN /install-helm.sh \
+    && rm -rfv /install-helm.sh
```
ferrarimarco, line 267: To have more controls over dependency updates, can you refactor this to grab helm from a container image?

Maybe this one https://hub.docker.com/r/alpine/helm

TLDR: Add a FROM statement near the others (see the top of this file), and copy the relevant binaries from that.

Also, that image appears to be a multi-platform image, so this will help when we'll eventually support arm64

scripts/install-helm.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
ferrarimarco, line 1: If we install helm from that container image, we don't need this script.

ferrarimarco (changes requested):
Thanks for this PR!

Besides this, we would need some test cases (good and bad ones) in this dir: https://github.com/super-linter/super-linter/tree/main/test/linters/checkov

Dockerfile
andrewkreuzer: ya can do, I was trying to find an image built by the helm team but couldn't find one, I can use that one though!

ferrarimarco:
One minor change: can you please make the test helm charts smaller? We don't need to have a fully working chart as a test case. What we need is the bare minimum to have tests pass :)

Thanks!

andrewkreuzer:
@ferrarimarco reduced the size of the test charts hopefully that is small enough

andrewkreuzer:
> @andrewkreuzer Thanks, the new test cases seems to be picked up. Nice!
> 
> Quick question: do the `good` test cases fail if you don't install helm? Just to confirm that we're covering this case as well.
> 
> Thanks!

no checkov will simple not run the helm checks if it can't find the helm binary

ferrarimarco:
SGTM, I just found a reference about this in the [Checkov docs](https://www.checkov.io/7.Scan%20Examples/Helm.html)

ferrarimarco (approved):

Dockerfile
```diff
@@ -259,6 +260,11 @@ COPY dependencies/google-java-format /google-java-format
 RUN --mount=type=secret,id=GITHUB_TOKEN /install-google-java-format.sh \
     && rm -rfv /install-google-java-format.sh /google-java-format
 
+################
+# Install Helm #
+################
+COPY --from=helm /usr/bin/helm /usr/bin/
```
ferrarimarco, line 266: Can you add an entry for `helm` in the

- [linters list](https://github.com/super-linter/super-linter/blob/88ea3923a7e1f89dd485d079f6eb5f5e8f937589/test/inspec/super-linter/controls/super_linter.rb#L134), so we exercise it at least once? Please also add a comment explaining that this is not a linter, but we need it for Checkov.
- [files list](https://github.com/super-linter/super-linter/blob/88ea3923a7e1f89dd485d079f6eb5f5e8f937589/test/inspec/super-linter/controls/super_linter.rb#L459), so we check for its existence.

ferrarimarco (changes requested):
Thanks a lot for all the work here.

We need just two more quick things about this in the test suite, and we should be good to go.

test/inspec/super-linter/controls/super_linter.rb
```diff
@@ -160,6 +160,7 @@
     { linter_name: "goreleaser"},
     { linter_name: "google-java-format", version_command: "java -jar /usr/bin/google-java-format --version"},
     { linter_name: "hadolint"},
+    { linter_name: "helm"}, # not used as linter, needed for checkov's helm framework
```
ferrarimarco, line 163: Please, refactor this to `{ linter_name: "helm", version_option: "version"}` to fix the InSpec error.

Thanks!

ferrarimarco (changes requested):

ferrarimarco (approved):

PR #5679, opened 2024-05-24
feat: allow using both prettier and standardjs

ferrarimarco:
# Proposed changes

- Allow using both Prettier and StandardJS at the same time by removing
  JAVASCRIPT_DEFAULT_STYLE and TYPESCRIPT_DEFAULT_STYLE
  configuration variables. They create a confusing corner case that
  Super-linter already handles with its core logic that's in place to
  enable and disable linters.
- Document VALIDATE_JAVASCRIPT_PRETTIER and VALIDATE_TYPESCRIPT_PRETTIER
  because they were missing from the README.
- Add missing TYPESCRIPT_PRETTIER tests. We didn't notice it was missing because we weren't running TYPESCRIPT_PRETTIER at all. 

## Readiness checklist

### Pull request author tasks


ferrarimarco:
@zkoppert can you please review this one? Thanks a lot!

zkoppert:
>Document VALIDATE_JAVASCRIPT_PRETTIER and VALIDATE_TYPESCRIPT_PRETTIER
because they were missing from the README.
Add missing TYPESCRIPT_PRETTIER tests. We didn't notice it was missing because we weren't running TYPESCRIPT_PRETTIER at all.

😅

test/linters/typescript_prettier/typescript_good_1.ts
```diff
@@ -0,0 +1,11 @@
+enum Test {
+  Hoo = "hoo",
+}
+
+const spiderman = (person: string): string => {
+  return `Hello, ${person}`;
+};
+
+const user = "Peter Parker";
```
zkoppert, line 9: ❤️

zkoppert (dismissed):
✨

test/linters/typescript_prettier/typescript_good_1.ts
ferrarimarco: :)

ferrarimarco:
sorry, forgot to fix the upgrade guide :)

zkoppert (approved):

ferrarimarco:
Hi!

You might have a point here, although the reason why we didn't mark this as a breaking change is that enabling this behavior has always been possible, but it wasn't documented.

The fact that they didn't run at the same time created a confusing corner case, which we classified as a bug, as you can also see from the fact that we were missing both documentation and a test case.

gavllew:
The thing that makes it a "confusing corner case" is that standard and prettier are formatters, not linters.  It makes sense to run lots of linters to pick up issues in a variety of file types.  IMO it never makes sense to run multiple formatters (for the same file type), as they are not going to agree on the correct style - if they did then one of them would be redundant.  The old config option made much more sense to me that what was introduced here.

Layer on top of that the fact it breaks existing configuration without a major version bump, and probably breaks out-of-the-box default configuration for JS/TS files, then my suggestion would be to roll back the change and re-consider how and when the changes should be re-introduced.

ferrarimarco:
Again, you have a point in suggesting that users might want to run only one formatter per file type.

The confusing corner case I'm referring to is that, before this change, to do that for Standard and Prettier, you would need to set the `JAVASCRIPT_DEFAULT_STYLE` and `TYPESCRIPT_DEFAULT_STYLE` variables, while all other tools (both linters and formatters) that Super-linter ships use `VALIDATE_xxxx` variables for that. After this change, you can use `VALIDATE_JAVASCRIPT_PRETTIER`, `VALIDATE_JAVASCRIPT_STANDARD`, `VALIDATE_TYPESCRIPT_PRETTIER`, and `VALIDATE_TYPESCRIPT_STANDARD` for the same purpose, bringing this in line with what you do with other linters and formatters.

The issue, in this case, is that the `VALIDATE_JAVASCRIPT_PRETTIER`, `VALIDATE_JAVASCRIPT_STANDARD`, `VALIDATE_TYPESCRIPT_PRETTIER`, and `VALIDATE_TYPESCRIPT_STANDARD` variables were breaking the behavior described in the [Configure Super-linter](https://github.com/super-linter/super-linter?tab=readme-ov-file#configure-super-linter) section of the readme, reported here for convenience:

> The VALIDATE_[LANGUAGE] variables work as follows:
> super-linter runs all supported linters by default.
> If you set any of the VALIDATE_[LANGUAGE] variables to true, super-linter defaults to leaving any unset variable to false (only validate those languages).
> If you set any of the VALIDATE_[LANGUAGE] variables to false, super-linter defaults to leaving any unset variable to true (only exclude those languages).
> If you set any of the VALIDATE_[LANGUAGE] variables to both true and false, super-linter fails reporting an error.

This is why this change was necessary.

The points you're raising warrant an addition to the upgrade guide, though. I'll provide it in a separate PR.

ferrarimarco:
Thanks for raising this @gavllew . I just sent https://github.com/super-linter/super-linter/pull/5948 to help with the configuration update.

PR #5983, opened 2024-08-06
fix: remove undocumented secondary extension corner cases

ferrarimarco:
# Proposed changes

  - Remove an undocumented corner case that might cause Super-linter to consider a configuration value for a configuration file that points to a non-existing YAML file as valid. Before this change, Super-linter falls back to a yml extension if the configuration value points to a non-existing file with a yaml extension, and vice-versa. Now, if the configuration file doesn't exist, Super-linter throws a fatal error.
  - Remove the LinterRulesLocation and move its logic in GetLinterRules because it's a simple if clause that we don't use anywhere else.
  - Avoid using eval when initializing configuration file variables.

## Readiness checklist

### Pull request author tasks

  I added the `Fix #ISSUE_NUMBER` or `Close #ISSUE_NUMBER` text to the description of the pull request.

- [x] Add the pull request to a milestone, eventually creating one, that matches with the version that release-please proposes.


ferrarimarco:
@zkoppert @Hanse00 Can you please have a look at this one? It's not essential, I think, but removes an untested and undocumented path of code that might hide configuration errors.

WDYT?

Thanks!

Hanse00:
I agree in principle with the idea that we should remove this. Although it's a convenience, it could cause some serious confusion if we just always assume that two different files, with different file extensions, are interchangeable. Either point us to a valid file, or don't.

That being said, I think this technically qualifies as a breaking change. If anyone out there is currently relying on this trick for their config to be loaded, then this update would suddenly mean it stops working. So I guess if we're following SemVer strictly, this needs to go in a major version?

ferrarimarco:
Good point. I'll mark this as a breaking change and move it to the 7.0.0 milestone.

Thanks!

Hanse00 (approved):

zkoppert (approved):
Agree with making it breaking change. Good call! 👍🏻

ferrarimarco:
Thanks for your reviews!

Sec32fun32 (approved):

PR #6242, opened 2024-10-08
fix: ensure setting /github/workdir as Git safe directory works always

klaernie:
When running super-linter on in a self hosted runner based on Action Runner Controller in K8s container mode the /github/home directory is not neccesarily mounted. But since we expect this directory to exist super-linter fails to set /github/workspace as a Git safe directory:

    2024-10-04 17:58:37 [INFO]   Successfully validated GITHUB_WORKSPACE: /github/workspace
    2024-10-04 17:58:37 [DEBUG]   Configuring Git safe directories
    2024-10-04 17:58:37 [DEBUG]   Set /github/workspace as a Git safe directory
    error: could not lock config file /github/home/.gitconfig: No such file or directory
    2024-10-04 17:58:37 [FATAL]   Cannot configure /github/workspace as a Git safe directory.

Making sure this directory exists in container image ensures that under no circumstances we can fail - if the action runner provides /github/home it will be mounted over, and if not we provide a directory where git can actually create the new .gitconfig.

## Readiness checklist

### Pull request author tasks

- [X] I included all the needed documentation for this change.
- [X] I provided the necessary tests.
- [X] I squashed all the commits into a single commit.
- [X] I followed the
- [X] I wrote the necessary upgrade instructions in the
      [upgrade guide](../docs/upgrade-guide.md).
- [X] If this pull request is about and existing issue, I added the
      the pull request.

      released version.
      `infrastructure`.


klaernie:
One caveat, which I don't know if you want to accept this: hadolint complains about two immediately adjacent `RUN` statements, however I took the authority to ignore this finding, since both statements do not have any thought in common. If you think these should be rather one statement feel free to tell me, and I'll amend the commit.

Theghost256 (approved):

ferrarimarco:
Thanks for this PR. I get the idea behind this fix.

What I don't know, is how the runner you mention configures volume mounts.

For example, when using a GitHub-hosted runners, it mounts:

- `-v "/home/runner/work/_temp/_github_home":"/github/home"`
- `-v "/home/runner/work/super-linter/super-linter":"/github/workspace"`

It also passes `-e "HOME"`, but we don't print it in the debug log, so I can only image that it's set to `/github/workspace`.

The command that's failing is:

```
git config --global --add safe.directory "${safe_directory}"
```

The `--global` option means:

```
--global
For writing options: write to global ~/.gitconfig file rather than the repository .git/config, write to $XDG_CONFIG_HOME/git/config file if this file exists and the ~/.gitconfig file doesn’t.
```

All this considered, I think that a more scalable solution might be to creating the `${HOME}` directory at runtime if it doesn't exist, rather than creating it when building the image. This way, we don't expect the `${HOME}` to always be `/github/home`. What do you think?

klaernie:
Figuring out how the volumes are mounted was not that easy - it's specified in https://github.com/actions/runner-container-hooks/blob/73655d4639a62f6e4b3d70b5878bc4367c0a436e/packages/k8s/src/k8s/utils.ts#L18

From having run actions both in GitHub.com and with self hosted runners in GitHub enterprise I've never seen a `$HOME` stray from `/github/home`. Sadly I have not yet found any documentation declaring it part of the interface.

You might be correct in that creating the home directory when we need it and it does not exist would work, but it requires us to at some point in the duration decide to create it - which means that there is always a time frame before that in which future changes could introduce a new requirement of the homedir. Creating the homedir as part of the container build ensures that it is present - always, and no matter if it is a volume mount.

klaernie:
Thinking and searching a bit more, I now considered the case that even users can override `$HOME`. Maybe the best would be to do both approaches simultaneously:
 - make sure that `/github/home` exists for 99% of the users
 - test the return code of `git config --global --add safe.directory "${safe_directory}"`, and if it failed check/create a missing homedir and repeat the command.

Digging more I found these:
- https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#file-systems
- https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#docker-container-filesystem

The document that `/github/home` is a reserved name and that it is created for docker based actions. However this is only documented for GitHub hosted runners, not for self-hosted runners. For these actions running without docker there is no guarantee for the home directory to have a stable name, this seems to both apply to runners setup manually and via [ARC](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller). But for docker-based actions both the manually setup runners and ARC's docker-in-docker mode adhere to what the documentation documents for GitHub-hosted runners. Only ARC's Kubernetes mode seems to deviate - but by still setting `$HOME` to `/github/home`, just not mounting a directory there, however that is the most beneficial mode to run a big image like super-linter (benefiting from the image cache of the K8s nodes, storing the image across runs, where the d-in-d needs to pull the image from the registry on every run).

klaernie:
I btw also asked in actions/runner-container-hooks#181 to clarify why this directory does not get mounted in the ARC K8s mode

ferrarimarco:
Thanks for all this research! Let's see where https://github.com/actions/runner-container-hooks/issues/181 leads us.

Creating the directory at build time might make sense as well because it will get overridden by the volume mount, eventually.

klaernie:
Thanks for the suggestion, it's a very good point to make that consistent. I applied it and rebased ontop of the current main branch. The tests still pass nicely

Dockerfile
```diff
@@ -435,6 +435,11 @@ ENV RENOVATE_X_IGNORE_RE2="true"
 ENV VERSION_FILE="/action/linterVersions.txt"
 RUN mkdir /action
 
+# create the homedir, so that in case it is not present (like on action-runner-controller based selfhosted runners)
+# we do not fail at setting /github/workspace as a safe git directory
+ENV HOME="/github/home"
+RUN mkdir -p /github/home
```
ferrarimarco, line 441: Can you reuse `${HOME}` here?

So refactor `/github/home` to `"${HOME}"`

Thanks!

klaernie: Done. Although there is one caveat: this only works while setting HOME during the build of a rebuilt image.

Dockerfile
```diff
@@ -434,6 +434,11 @@ ENV RENOVATE_X_IGNORE_RE2="true"
 ENV VERSION_FILE="/action/linterVersions.txt"
 RUN mkdir /action
 
+# create the homedir, so that in case it is not present (like on action-runner-controller based selfhosted runners)
+# we do not fail at setting /github/workspace as a safe git directory
+ENV HOME="/github/home"
+RUN mkdir -p "${HOME}"
```
ferrarimarco, line 440: We also need tests for these. Can you add them in `test/inspec/super-linter/controls/super_linter.rb`?

Basically, you need to:

- Add a check for `HOME` in the `super-linter-environment-variables` control
- Add a check for `/github/home` in the `super-linter-validate-directories` control

Thanks!

ferrarimarco: (note that I rebased this branch on top of the latest `main`)

klaernie: sure, done. I amended my commit with that, since I consider tests integral to the change.

ferrarimarco: Great!

ferrarimarco (approved):

ferrarimarco:
Thanks for working on this!

klaernie:
No problem! I think the latest commit hasn't run in the CI yet.

ferrarimarco:
Sure, I can have a look

ferrarimarco (approved):

PR #6808, opened 2025-05-25
Fix conditional push for lint fixes

saulshanabrook:
I found that if I didn't include an `always &&` in the conditional it wouldn't run when the previous failed for some reason. ref: https://github.com/egraphs-good/egglog-demo/pull/1

## Readiness checklist

### Pull request author tasks

- [ ] I included all the needed documentation for this change.
- [ ] I squashed all the commits into a single commit.
- [ ] I followed the
- [ ] If this pull request is about and existing issue, I added the
      the pull request.

      released version.
      `infrastructure`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,6 +547,7 @@ jobs:
         # - Pull requests
         # - Not on the default branch
         if: >
+          always() &&
           github.event_name == 'pull_request' &&
           github.ref_name != github.event.repository.default_branch
         uses: stefanzweifel/git-auto-commit-action@v5


ferrarimarco:
Yes, unless you add `always()`, as you did.

From what your reply, I suppose that the workflow you have in mind is that you still want a commit with fixes, even if there are linting errors that aren't automatically fixable?

saulshanabrook:
> From what your reply, I suppose that the workflow you have in mind is that you still want a commit with fixes, even if there are linting errors that aren't automatically fixable?

Yeah, if some can be fixed automatically I want to push them back and then if others can't be fixed automatically they would have to be fixed manually after.

Issue #7009, opened 2025-09-07
Local dev with worktree fails


danielorbach:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Are you using the latest Super-linter version available?

- [x] I am using the latest Super-linter version.
- [x] I can reproduce the issue running Super-linter using complete version identifier (example: vX.Y.Z), and not just with a partial one (example: vX)

- [x] I am using the super-linter/super-linter action or container image, and not the deprecated github/super-linter action or container image.

### Are you resonably sure that it's a Super-linter issue, and not an issue related to a tool that Super-linter runs?

- [x] I think that this is a Super-linter issue.

### Current Behavior

After mounting both the git worktree and the git common-dir, I observe the super-linter docker failing with an error in `lib/linter.sh` about an "unbound variable" (see the configuration and logs):

```shell
/action/lib/linter.sh: line 289: GITHUB_SHA: unbound variable
```

### Expected Behavior

After mounting both the git worktree and the git common-dir, I expect the super-linter docker to run successfully.

### Super-Linter version

```markdown
latest
```

### Super-linter configuration

```yaml
docker run --rm --platform=linux/amd64 \
  -e LOG_LEVEL=INFO \
  -e RUN_LOCAL=true \
  -e DEFAULT_BRANCH=main \
  -v "$(pwd)":/tmp/lint \
  -v "$(git rev-parse --path-format=absolute --git-common-dir)":"$(git rev-parse --path-format=absolute --git-common-dir)" \
  ghcr.io/super-linter/super-linter:latest
```

### Relevant log output

```shell
2025-09-07 20:36:20 [INFO]   --------------------------------------------------------------------------------

                              /@@#///////@@/(@//@%/(@.@(       @@
                          @@//////////////////////////////#*  @@@
                        @////@//(///////////@@@@@///@//@/@**//@@(
                      @///////@///////////////@@@@    (           @,
                     @/(&/@////////////////////                     @
                    @////////////////////////@@                      @
                  @%////////(//////////%/////&@            @@       *,@           ______________
             @@@@@/@/#/////(&//////////////////                       .@         /              \
        *@@@@@.    .%///(//@//////////////////&.   .@@,                 @%      / Don't mind me  \
      @@%           .&@&&/@.@//&/////(//////////    @@@@@@@@@         .. &@    / I'm just looking \
    @@%               @@@@@   @&/////////////////#   @/       V  @@/ ,@@@ @   <  for some trash... |
[...]

PR #6983, opened 2025-08-28
feat: support git worktrees

ferrarimarco:
Close #6944

## Readiness checklist

In order to have this pull request merged, complete the following tasks.

### Pull request author tasks

- [x] I checked that all workflows return a success.
- [x] I included all the needed documentation for this change.
- [x] I provided the necessary tests.
- [x] I squashed all the commits into a single commit.
- [x] I followed the
      [Conventional Commit v1.0.0 spec](https://www.conventionalcommits.org/en/v1.0.0/).
- [x] I wrote the necessary upgrade instructions in the
      [upgrade guide](https://github.com/super-linter/super-linter/blob/main/docs/upgrade-guide.md).
- [x] If this pull request is about and existing issue, I added the
      `Fix #ISSUE_NUMBER` or `Close #ISSUE_NUMBER` text to the description of
      the pull request.

### Super-linter maintainer tasks

- [x] Label as `breaking` if this change breaks compatibility with the previous
      released version.
- [x] Label as either: `automation`, `bug`, `documentation`, `enhancement`,
      `infrastructure`.
- [x] Add the pull request to a milestone, eventually creating one, that matches
      with the version that release-please proposes in the
      `preview-release-notes` CI job.

PR #7010, opened 2025-09-07
Fix GITHUB_SHA for both worktrees and regular repos

danielorbach:
Move GITHUB_SHA initialization outside the conditional block to ensure it's set for both git worktrees and regular repositories, fixing #7009 .

I am developing on Apple Silicon, so I am having some troubles running all the makefile targets. I was able to eventually build a local image (the `--platform=linux/amd64` flag helps for all cases except those calls to docker that use the `--check` flag) by modifying the Makefile a bit. Afterwards, I was able to run super-liner for my local worktree based on the locally built image.

## Readiness checklist

### Pull request author tasks

- [x] I followed the
- [ ] If this pull request is about and existing issue, I added the
      the pull request.

      released version.
      `infrastructure`.


ferrarimarco:
Thanks for this PR.

I suggest adding a test case to cover this scenario. You can find an example of a test case in #6983 (check the edits to `Makefile` and the `run_test_case_git_worktree` in `test/run-super-linter-tests.sh`, ignore the rest).

Thanks!

danielorbach:
I've had some trouble contributing on my ARM machine, but I think I was able to add the relevant test.

<details><summary>Details</summary>
<p>

After all the changes to the Makefile that I added to https://github.com/super-linter/super-linter/pull/6986, and installing a Bash version that is compatible with modern linux dev machines, I was able to run the tests on my machine. Yet, temporary directories are created in `/var` which is a link to `/private/var`. This confuses that paths that Git sees (because it uses absolute paths sometimes).

Though I was unable to get tests to pass, I was able to confirm that:
1. Testing _invalid_ worktree mounts fails with the correct error.
2. My test with _valid_ worktree mounts indeed mounts the git-common-dir when invoking super-linter.

</p>
</details>

danielorbach:
Also, I got a weird conflict in Makefile, so I pulled `upstream/main`. I hope I did not break anything. The conflict was in the long line that defines `test: ...` dependencies. @ferrarimarco verify me so I don't break the upstream 🥺

ferrarimarco:
@danielorbach thanks for this PR, I just applied minor refactoring to the test functions to avoid some duplication.

ferrarimarco (approved):

PR #7848, opened 2026-05-22
feat: update existing PR summary comment instead of creating new ones

joshjohanning:
When `ENABLE_GITHUB_PULL_REQUEST_SUMMARY_COMMENT` is enabled (it is by default?), super-linter now searches for an existing summary comment (identified by an HTML marker) and updates it in place, rather than creating a new comment on every push. This reduces comment noise / spamminess on pull requests with multiple subsequent pushes.

The implementation:
  identify super-linter summary comments
- Lists existing PR comments and finds any with the marker
- Updates the existing comment via PATCH if found, or creates a new one
- Refactors CallGitHubApi to accept an optional HTTP method parameter (backwards-compatible, defaults to POST)

Closes: super-linter/super-linter#7458

I verified that this new logic works in
- https://github.com/joshjohanning-org/super-linter-test/pull/1#issuecomment-4520097610

image

## Readiness checklist

### Pull request author tasks

- [x] I checked that all workflows return a success.
- [x] I followed the
- [x] I wrote the necessary upgrade instructions in the
- [x] If this pull request is about and existing issue, I added the
      the pull request.

      released version.
      `infrastructure`.


lib/functions/output.sh
```diff
@@ -215,6 +220,91 @@ CreateGitHubIssueComment() {
   fi
 }
 
+SUPER_LINTER_SUMMARY_COMMENT_MARKER="<!-- super-linter-summary-comment -->"
```
ferrarimarco: Can you move this to `lib/globals/output.sh` ?

lib/functions/output.sh
```diff
@@ -215,6 +220,91 @@ CreateGitHubIssueComment() {
...
+    return 1
+  fi
+
+  local NEXT_URL="${GITHUB_ISSUE_COMMENTS_URL}?per_page=100"
+  while [[ -n "${NEXT_URL}" ]]; do
+    local RESPONSE_HEADERS
+    RESPONSE_HEADERS="$(mktemp)"
+
+    local LIST_COMMENTS_OUT
+    if ! LIST_COMMENTS_OUT=$(
```
ferrarimarco: Can we reuse the `CallGitHubApi` function here?

lib/functions/output.sh
```diff
@@ -215,6 +220,91 @@ CreateGitHubIssueComment() {
...
+        --show-error \
+        --silent \
+        --dump-header "${RESPONSE_HEADERS}" \
+        --url "${NEXT_URL}" \
+        -H "accept: application/vnd.github+json" \
+        -H "authorization: Bearer ${GITHUB_TOKEN}" \
+        -H "X-GitHub-Api-Version: 2022-11-28"
+    ); then
+      rm -f "${RESPONSE_HEADERS}"
+      warn "Failed to list comments for issue #${GITHUB_ISSUE_NUMBER}: ${LIST_COMMENTS_OUT}"
```
ferrarimarco: We likely want an `error` here.

lib/functions/output.sh
```diff
@@ -215,6 +220,91 @@ CreateGitHubIssueComment() {
...
+      return 1
+    fi
+
+    local EXISTING_COMMENT_ID
+    if ! EXISTING_COMMENT_ID=$(echo "${LIST_COMMENTS_OUT}" | jq -r --arg marker "${SUPER_LINTER_SUMMARY_COMMENT_MARKER}" '[.[] | select((.body // "") | startswith($marker))] | last | .id // empty'); then
+      rm -f "${RESPONSE_HEADERS}"
+      warn "Error while parsing comments response"
```
ferrarimarco: We likely want an `error` here.

lib/functions/output.sh
```diff
@@ -215,6 +220,91 @@ CreateGitHubIssueComment() {
...
+
+  if [[ -z "${GITHUB_TOKEN:-}" ]]; then
+    warn "Provide a GitHub token to call the GitHub API: ${GITHUB_ISSUE_COMMENTS_URL}"
+    return 1
+  fi
+
+  local NEXT_URL="${GITHUB_ISSUE_COMMENTS_URL}?per_page=100"
+  while [[ -n "${NEXT_URL}" ]]; do
+    local RESPONSE_HEADERS
+    RESPONSE_HEADERS="$(mktemp)"
```
ferrarimarco: Is there room to move to in-memory processing, instead of saving output to a temp file? that would simplify cleanup and speed things up a bit. You might have to tweak `jq` invocations as well.

lib/functions/output.sh
```diff
@@ -227,8 +317,32 @@ CreateGitHubPullRequestSummaryComment() {
...
   fi
 
-  if ! CreateGitHubIssueComment "${SUMMARY_COMMENT_BODY}" "${GITHUB_PULL_REQUEST_NUMBER}"; then
-    warn "Error while posting pull request summary"
-    return 1
+  # Prepend the marker so we can find this comment later
+  SUMMARY_COMMENT_BODY="${SUPER_LINTER_SUMMARY_COMMENT_MARKER}
+${SUMMARY_COMMENT_BODY}"
+
+  # Check if there's an existing summary comment to update
```
ferrarimarco: Can you please refactor this to be optional (default `true`)? So users that prefer one comment for each run can still have that.

You need to introduce a new configuration variable for that. Define it in `lib/globals/output.sh` and describe it in the README configuration variables table. Maybe `UPDATE_EXISTING_GITHUB_PULL_REQUEST_SUMMARY_COMMENT`

ferrarimarco (changes requested):
Thanks a lot for this PR! Some minor things to address 👍

lib/functions/output.sh
joshjohanning: Done — moved to `lib/globals/output.sh`.

lib/functions/output.sh
joshjohanning: Done — refactored to use `CallGitHubApi` with a new optional `INCLUDE_RESPONSE_HEADERS` param. When true it passes `--include` to curl, so the caller gets headers + body in `CALL_GITHUB_API_OUT` without a temp file.

lib/functions/output.sh
joshjohanning: Fixed.

lib/functions/output.sh
joshjohanning: Fixed.

lib/functions/output.sh
joshjohanning: Done — switched to `--include` so headers and body come back together in `CALL_GITHUB_API_OUT`. Body extraction and Link header parsing use awk to handle multiple header blocks (e.g. from a redirect).

lib/functions/output.sh
joshjohanning: Done — added `UPDATE_EXISTING_GITHUB_PULL_REQUEST_SUMMARY_COMMENT` (defaults to `true`), defined in `lib/globals/output.sh` and documented in the README.

joshjohanning:
Comments addressed @ferrarimarco! Thanks for the review

lib/functions/output.sh
```diff
@@ -134,28 +134,39 @@ CallGitHubApi() {
...
+
+  if [[ "${INCLUDE_RESPONSE_HEADERS}" == "true" ]]; then
+    CURL_ARGS+=(--include)
+  fi
+
+  if [[ -n "${PAYLOAD}" ]]; then
+    CURL_ARGS+=(-d "${PAYLOAD}")
+  fi
+
+  CALL_GITHUB_API_OUT=""
```
ferrarimarco: You can remove this initialization. We don't want to leak this outside the function.

Add an `echo ${CALL_GITHUB_API_OUT}` after the next `if`, so we emit the output only on success.

This is needed to avoid leaking state across functions, and makes it easier to write tests.

lib/functions/output.sh
```diff
@@ -215,6 +226,86 @@ CreateGitHubIssueComment() {
...
+  debug "Listing comments for issue #${GITHUB_ISSUE_NUMBER} to find existing summary comment"
+
+  if [[ -z "${GITHUB_TOKEN:-}" ]]; then
+    warn "Provide a GitHub token to call the GitHub API: ${GITHUB_ISSUE_COMMENTS_URL}"
+    return 1
+  fi
+
+  local NEXT_URL="${GITHUB_ISSUE_COMMENTS_URL}?per_page=100"
+  while [[ -n "${NEXT_URL}" ]]; do
+    if ! CallGitHubApi "${NEXT_URL}" "${GITHUB_TOKEN}" "" "GET" "true"; then
```
ferrarimarco: Please, refactor this if as `if ! CALL_GITHUB_API_OUT="$(CallGitHubApi....)"` so you can avoid leaking `CALL_GITHUB_API_OUT` outside `CallGitHubApi`.

We adopted this pattern across the code base. Thanks!

lib/functions/output.sh
```diff
@@ -215,6 +226,86 @@ CreateGitHubIssueComment() {
...
+      error "Failed to list comments for issue #${GITHUB_ISSUE_NUMBER}: ${CALL_GITHUB_API_OUT}"
+      return 1
+    fi
+
+    # Extract the response body from the last HTTP response block.
+    # curl --include --location can produce multiple header blocks (one per redirect);
+    # reset on each new HTTP/ status line so we only parse the final response body.
+    local RESPONSE_BODY
+    RESPONSE_BODY=$(printf '%s' "${CALL_GITHUB_API_OUT}" | awk '
```
ferrarimarco: Can you please wrap this into an `if ! RESPONSE_BODY...` to check if it errors out? Otherwise it might silently swallow errors. Thanks!

lib/functions/output.sh
```diff
@@ -215,6 +226,86 @@ CreateGitHubIssueComment() {
...
+    fi
+
+    if [[ -n "${EXISTING_COMMENT_ID}" ]]; then
+      debug "Found existing summary comment with ID: ${EXISTING_COMMENT_ID}"
+      SUPER_LINTER_EXISTING_SUMMARY_COMMENT_ID="${EXISTING_COMMENT_ID}"
+      return 0
+    fi
+
+    # Parse the Link header from the last response block for the next page URL
+    NEXT_URL=$(printf '%s' "${CALL_GITHUB_API_OUT}" | awk '
```
ferrarimarco: Can you please wrap this into an `if ! NEXT_URL...` to check if it errors out? Otherwise it might silently swallow errors. Thanks!

lib/functions/output.sh
ferrarimarco: Please add `set -o pipefail` to this subshell

lib/functions/output.sh
```diff
@@ -215,6 +226,86 @@ CreateGitHubIssueComment() {
...
+    local RESPONSE_BODY
+    RESPONSE_BODY=$(printf '%s' "${CALL_GITHUB_API_OUT}" | awk '
+      /^HTTP\//{in_headers=1; body=""; next}
+      in_headers && /^\r?$/{in_headers=0; next}
+      !in_headers{body = body ? body "\n" $0 : $0}
+      END{print body}
+    ')
+
+    local EXISTING_COMMENT_ID
+    if ! EXISTING_COMMENT_ID=$(printf '%s' "${RESPONSE_BODY}" | jq -r --arg marker "${SUPER_LINTER_SUMMARY_COMMENT_MARKER}" '[.[] | select((.body // "") | startswith($marker))] | last | .id // empty'); then
```
ferrarimarco: Please add `set -o pipefail` to this subshell

lib/functions/output.sh
ferrarimarco: Please add `set -o pipefail` to this subshell

lib/functions/output.sh
ferrarimarco: Please redirect stderr to stdout in this subshell so we get any eventual error message. Thanks!

lib/functions/output.sh
ferrarimarco: Please redirect stderr to stdout in this subshell so we get any eventual error message. Thanks!

lib/functions/output.sh
ferrarimarco: Please redirect stderr to stdout in this subshell so we get any eventual error message. Thanks!

lib/functions/output.sh
```diff
@@ -215,6 +226,86 @@ CreateGitHubIssueComment() {
...
+  return 0
+}
+
+# Ref: https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#update-an-issue-comment
+UpdateGitHubIssueComment() {
+  local COMMENT_PAYLOAD="${1}" && shift
+  local COMMENT_ID="${1}" && shift
+
+  local UPDATE_ISSUE_COMMENT_API_PAYLOAD
+  if ! UPDATE_ISSUE_COMMENT_API_PAYLOAD="$(jq --null-input --arg body "${COMMENT_PAYLOAD}" '{body: $body}')"; then
```
ferrarimarco: Please redirect stderr to stdout in this subshell so we get any eventual error message. Thanks!

lib/functions/output.sh
```diff
@@ -215,6 +226,86 @@ CreateGitHubIssueComment() {
...
+    warn "Error while loading the contents of COMMENT_PAYLOAD to UPDATE_ISSUE_COMMENT_API_PAYLOAD"
```
ferrarimarco: Add `: ${UPDATE_ISSUE_COMMENT_API_PAYLOAD}`

lib/functions/output.sh
```diff
@@ -227,8 +318,39 @@ CreateGitHubPullRequestSummaryComment() {
...
   fi
 
-  if ! CreateGitHubIssueComment "${SUMMARY_COMMENT_BODY}" "${GITHUB_PULL_REQUEST_NUMBER}"; then
-    warn "Error while posting pull request summary"
-    return 1
+  # Prepend the marker so we can find this comment later
+  SUMMARY_COMMENT_BODY="${SUPER_LINTER_SUMMARY_COMMENT_MARKER}
+${SUMMARY_COMMENT_BODY}"
+
+  if [[ "${UPDATE_EXISTING_GITHUB_PULL_REQUEST_SUMMARY_COMMENT}" == "true" ]]; then
```
ferrarimarco, line 339: Can you please add a check for this "boolean" variable to

https://github.com/super-linter/super-linter/blob/95cfd43c5b2d9243f81129d094c3a69e2b8e7dab/lib/functions/validation.sh#L3

thanks!

lib/functions/output.sh
```diff
@@ -215,6 +226,86 @@ CreateGitHubIssueComment() {
   fi
 }
 
+# Ref: https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#list-issue-comments
+# Sets the global SUPER_LINTER_EXISTING_SUMMARY_COMMENT_ID variable; does not use stdout
+# so that log output from this function doesn't contaminate callers using command substitution.
+SUPER_LINTER_EXISTING_SUMMARY_COMMENT_ID=""
```
ferrarimarco: Please refactor the `FindExistingSummaryComment` function to `echo` the `SUPER_LINTER_EXISTING_SUMMARY_COMMENT_ID` instead of using a global variable, so we avoid leaking state outside functions.

ferrarimarco (changes requested):
Thanks a lot for your contribution! We really appreciate the effort!

This is almost good to go. There's some minor polishing to address to make this implementation a bit more robust.

You already did a lot, so please let me know if you need any help here.

An LLM should be able to address these comments fairly quickly, anyway 🤖  😄

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done.

lib/functions/output.sh
joshjohanning: Done, added in lib/functions/validation.sh.

joshjohanning:
No worries @ferrarimarco my pleasure to assist here.

Pushed another commit (0259fe22) addressing the remaining feedback:

- `CallGitHubApi` now uses a local variable and echoes output on success
- All callers updated to use `if ! VAR="$(CallGitHubApi ...)"` pattern
- `FindExistingSummaryComment` now echoes the ID instead of using a global
- `RESPONSE_BODY` and `NEXT_URL` wrapped in `if !` for error detection
- Added `set -o pipefail` and `2>&1` to all piped subshells
- Added : `${UPDATE_ISSUE_COMMENT_API_PAYLOAD}` reference
- Added `UPDATE_EXISTING_GITHUB_PULL_REQUEST_SUMMARY_COMMENT` to `ValidateBooleanConfigurationVariables` in `validation.sh`

Let me know if I missed anything :D

lib/functions/output.sh
ferrarimarco: Looks like this still needs to be addressed? Thanks!

As in:

```
 warn "Error while loading the contents of COMMENT_PAYLOAD to UPDATE_ISSUE_COMMENT_API_PAYLOAD: ${UPDATE_ISSUE_COMMENT_API_PAYLOAD}"
```

lib/functions/output.sh
```diff
@@ -215,6 +227,102 @@ CreateGitHubIssueComment() {
...
+UpdateGitHubIssueComment() {
+  local COMMENT_PAYLOAD="${1}" && shift
+  local COMMENT_ID="${1}" && shift
+
+  local UPDATE_ISSUE_COMMENT_API_PAYLOAD
+  if ! UPDATE_ISSUE_COMMENT_API_PAYLOAD="$(jq --null-input --arg body "${COMMENT_PAYLOAD}" '{body: $body}' 2>&1)"; then
+    warn "Error while loading the contents of COMMENT_PAYLOAD to UPDATE_ISSUE_COMMENT_API_PAYLOAD"
+    return 1
+  fi
+  : "${UPDATE_ISSUE_COMMENT_API_PAYLOAD}"
```
ferrarimarco: This should be moved where my previous comment is :)

ferrarimarco (changes requested):
Thanks! I think we're there, there's only a typo to fix.

Thanks!

ferrarimarco:
I fixed the linting issues and the typo.

There's an error tho:

```
2026-06-08 20:28:08 [WARN]   Failed to call GitHub API (https://api.github.com/repos/super-linter/super-linter/issues/comments/2026-06-08 20:28:08 [DEBUG]   Listing comments for issue #7848 to find existing summary comment): curl: (3) bad range in URL position 93:
https://api.github.com/repos/super-linter/super-linter/issues/comments/2026-06-08 20:28:08 [DEBUG]   Listing comments for issue #7848 to find existing summary comment
                                                                                            ^
2026-06-08 20:28:08 [WARN]   Failed to update GitHub issue comment
2026-06-08 20:28:08 [WARN]   Error while updating pull request summary comment
2026-06-08 20:28:08 [WARN]   Error while posting GitHub pull request comment.
```

ferrarimarco:
Ok, now we just have a permission issue:

```
2026-06-09 12:03:21 [WARN]   Failed to call GitHub API (https://api.github.com/repos/super-linter/super-linter/issues/7848/comments): curl: (22) The requested URL returned error: 403
```

This is most likely due to the workflow not having comment permission because it comes from a fork:

```
GITHUB_TOKEN Permissions
  Contents: read
  Issues: read
  Metadata: read
  Packages: read
  PullRequests: read
  Statuses: read
```

joshjohanning:
Thanks @ferrarimarco for pushing this over the finish line! I appreciate the effort 🙇  ✨

ferrarimarco:
Thanks for your contribution!

joshjohanning:
@ferrarimarco just noticed- were you planning on revving the action to `8.6.1` or `8.7.0` for us to be able to use? 😄

ferrarimarco:
V8.7.0 😀

joshjohanning:
@ferrarimarco  I mean, are you going to cut the release or are you waiting for a few other things to batch up? No worries either way just nudging in case you forgot to do this here :D

ferrarimarco:
The only thing we're waiting for Is a dependency update for checkout to fix a transient dependency vulnerability

joshjohanning:
Sounds good! Thank you

yermulnik:
@joshjohanning @ferrarimarco Yay, just noticed this wonderful improvement in the latest version and already have enjoyed `super-linter` updating the same comment in dependabot's PR to bump `super-linter` to latest 🥳 Just wanted to express appreciation ❤️

joshjohanning:
Woot woot! Awesome @yermulnik 🙇

Issue #396, opened 2020-07-07
HTMLHint v1.0.0 upgrade


zkoppert:
For this issue we want to make the necessary changes needed to support HTMLHint v1.0.0 once it is available.

> Hi, I'm a co-maintainer of HTMLHint
> I want to let you know that `.htmlhintrc` will not work with `v1.0.0` anymore, because we are currently rewrite the complete configuration structure https://github.com/htmlhint/HTMLHint/issues/279

_Originally posted by @Shinigami92 in https://github.com/github/super-linter/pull/373#issuecomment-654679978_

PR #7884, opened 2026-06-10
fix(go_modules): scan module sub-packages with golangci-lint

### LINKED ISSUE #8044: GO_MODULES fails when a module root contains no Go files

- [x] I have searched the existing issues

### Are you using the latest Super-linter version available?

- [x] I am using the latest Super-linter version.
- [x] I can reproduce the issue running Super-linter using a complete version identifier, and not just a partial one.
- [x] I am using the super-linter/super-linter action or container image, and not the deprecated github/super-linter action or container image.

### Are you reasonably sure that it's a Super-linter issue, and not an issue related to a tool that Super-linter runs?

- [x] I think that this is a Super-linter issue.

### Current Behavior

`GO_MODULES` fails for a valid Go module when the module root contains `go.mod`, but Go packages are located only in subdirectories.

For this layout:

```text
.
├── go.mod
└── cmd
    └── example
        └── main.go
```

Super-Linter creates two `GO_MODULES` items:

```text
/tmp/lint/cmd/example
/tmp/lint
```

It then runs `golangci-lint` against both directories. The command for the module root fails because the root is not a Go package:

```text
golangci-lint run --allow-parallel-runners -c /action/lib/.automation/.golangci.yml /tmp/lint
```

Running the same bundled `golangci-lint 2.12.2` directly succeeds:

```text
golangci-lint run --allow-parallel-runners
golangci-lint run --allow-parallel-runners ./...
```

### Expected Behavior

`GO_MODULES` should lint all packages in the module without requiring the directory containing `go.mod` to also contain Go source files.

### Super-Linter version

```markdown
v8.7.0
```

### Super-linter configuration

```yaml
RUN_LOCAL: "true"
USE_FIND_ALGORITHM: "true"
DEFAULT_WORKSPACE: /tmp/lint
VALIDATE_GO: "false"
VALIDATE_GO_MODULES: "true"
MULTI_STATUS: "false"
ENABLE_GITHUB_ACTIONS_STEP_SUMMARY: "false"
ENABLE_GITHUB_PULL_REQUEST_SUMMARY_COMMENT: "false"
SUPPRESS_POSSUM: "true"
LOG_LEVEL: DEBUG
```

### Relevant log output

```shell
Super-linter
- Image Revision: 4ce20838b8ab83717e78138c5b3a1407148e0918
- Image Version: 8.7.0

[GO_MODULES] golangci-lint: 2.12.2

There are 2 items to lint for GO_MODULES: /tmp/lint/cmd/example /tmp/lint

Command: golangci-lint run --allow-parallel-runners -c /action/lib/.automation/.golangci.yml /tmp/lint/cmd/example
Exitval: 0
Stdout: 0 issues.

Command: golangci-lint run --allow-parallel-runners -c /action/lib/.automation/.golangci.yml /tmp/lint
Exitval: 5
Stderr: level=error msg="Running error: context loading failed: failed to load packages: failed to load packages: package example.com/super-linter-go-modules-repro: no go files to analyze"

Errors found in GO_MODULES
Super-linter detected linting errors
```

### Steps To Reproduce

1. Create this directory structure:

```text
.
├── go.mod
└── cmd
    └── example
        └── main.go
```

2. Add `go.mod`:

```go
module example.com/super-linter-go-modules-repro

go 1.25
```

3. Add `cmd/example/main.go`:

```go
package main

func main() {}
```

4. Run Super-Linter locally:

```shell
docker run --rm \
  -e RUN_LOCAL=true \
  -e USE_FIND_ALGORITHM=true \
  -e DEFAULT_WORKSPACE=/tmp/lint \
  -e VALIDATE_GO=false \
  -e VALIDATE_GO_MODULES=true \
  -e MULTI_STATUS=false \
  -e ENABLE_GITHUB_ACTIONS_STEP_SUMMARY=false \
  -e ENABLE_GITHUB_PULL_REQUEST_SUMMARY_COMMENT=false \
  -e SUPPRESS_POSSUM=true \
  -e LOG_LEVEL=DEBUG \
  -v "$PWD:/tmp/lint" \
  ghcr.io/super-linter/super-linter:slim-v8.7.0
```

5. Observe that the `GO_MODULES` check fails with `no go files to analyze`.

### Anything else?

The same failure occurs in a public GitHub Actions run: https://github.com/Netcracker/qubership-monitoring-operator/actions/runs/30622680143/job/91130693120

The file-list builder appears to add the directory of every Go file to `GO_MODULES`, rather than the directory containing the discovered `go.mod`:

https://github.com/super-linter/super-linter/blob/4ce20838b8ab83717e78138c5b3a1407148e0918/lib/functions/buildFileList.sh#L607-L635

This is related to, but different from, #6294. That issue concerns individual-file linting through `GO`; this failure occurs with `GO` disabled and `GO_MODULES` enabled.

vlsi:
## Why

When a PR changes only `go.mod` (Renovate bumps, `toolchain` directive additions, and similar), the `GO_MODULES` linter fails with:

```
package github.com/example/my-operator: no go files to analyze
```

Super-linter passes the module root directory to golangci-lint as a bare positional argument. Many Go projects — especially Kubernetes operators — keep all source code in sub-directories (`cmd/`, `api/`, `controllers/`, `pkg/`) and have no `.go` files directly in the module root, so golangci-lint finds nothing to analyse and fails.

## What

Append `/...` to the directory path passed to golangci-lint. The `/...` wildcard is idiomatic Go and works with absolute paths too (`/abs/path/...`): it tells golangci-lint (and the underlying `go list`) to scan the given directory and all sub-packages recursively, matching the behaviour a developer gets from `golangci-lint run ./...` locally.

Implementation note: the `/...` suffix is added as an explicit GNU Parallel replacement string (`{}/...`). This both injects the path and stops GNU Parallel from appending its default `{}`.

The `go_modules_good` test fixture now keeps its source under `cmd/`, leaving only `go.mod` and `go.sum` in the module root, so the test suite reproduces the failing case.

## How to verify

Create a Go module with `go.mod` in its root and all source files in a `cmd/` sub-directory, then run the `GO_MODULES` linter against a PR that changes only `go.mod`.

- Before this fix: `no go files to analyze`.
- After this fix: golangci-lint runs against all packages successfully.

Reproduced on super-linter `v8.6.0` with `Netcracker/qubership-monitoring-operator` (for example PR #396).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Fix #8044


ferrarimarco:
Hi, thanks for this PR.

I am not sure that this would actually fix the issue you're facing without breaking other use cases, such as linting only the files that changed (the entire Go module that has any change, in this case) when `VALIDATE_ALL_CODEBASE` is `false`.

The logic to find the path of the Go modules to lint checks if either:

- A `go.mod` file has been modified, get the path of the directory containing the modified `go.mod` file.
- A `.go` file has been modified, and there's a `go.mod` file in the same directory as the modified `.go` file, or in any directory that's a parent of the directory where the modified `.go` file.

https://github.com/super-linter/super-linter/blob/6c432fd7d4a3cba53f275da5a54f882bacdeee8b/lib/functions/buildFileList.sh#L609-L641

In both cases, the end result is getting a path to a directory containing a `go.mod` file.

Can you please share more details about the directory structure of your Go project?

(I'd suggest opening an issue to investigate the problem first.)

ferrarimarco:
(removing the `bug` label until we confirm it's a problem with Super-linter.)

vlsi:
Thanks for the quick look. Here is the directory layout and a self-contained reproducer.

**Project layout** — a standard kubebuilder/operator module: `go.mod` in the root, all sources under sub-directories, no `.go` files in the module root:

```
operator/
├── go.mod
├── api/
├── cmd/
│   └── main.go
├── controllers/
└── pkg/
```

**Minimal reproducer**

`go.mod`:

```
module example.com/operator

go 1.24
```

`cmd/main.go`:

```go
package main

import "fmt"

func main() {
	fmt.Println("hello from cmd")
}
```

**Result with `golangci-lint v2.12.2`** (the version pinned in the Dockerfile), reproducing how Super-linter invokes it — the working directory is the module root and the path is passed as a positional argument:

```console
$ golangci-lint run /path/to/repro
level=error msg="Running error: context loading failed: ... package example.com/operator: no go files to analyze"
$ echo $?
5

$ golangci-lint run /path/to/repro/...
0 issues.
$ echo $?
0
```

**Root cause.** `golangci-lint run <dir>` is non-recursive: it loads only the package located directly in `<dir>`. When the path is the module root and the root holds no `.go` files, there is nothing to load, so golangci-lint exits `5` with `no go files to analyze`. The `/...` suffix makes the load recursive, matching what `golangci-lint run ./...` does locally — the command developers actually run.

**On the `VALIDATE_ALL_CODEBASE=false` concern.** This change does not narrow that path. golangci-lint already lints a whole package (every `.go` file in the directory), not a single changed file, so there is no per-file granularity to lose. Appending `/...` to the module-root path lints the whole module that changed — which matches the intent you described, linting the entire Go module that has any change. For a path that points at a sub-package (a `.go` file changed deep in the tree), `/...` only widens it from that one package to its nested packages; it never narrows the set.

Happy to open a separate issue to track this if you prefer.

jabenedicic:
Independently hit this on `v8.6.0` with a standard Go service layout (`go.mod` in root, all code under `cmd/`, `internal/`, …), and dug into the `VALIDATE_ALL_CODEBASE=false` concern since that's the one that matters. Sharing a full reproduction through the real super-linter container — I think the change is not just safe for incremental linting, it *fixes* a silent gap there.

**The mechanism is a typo, and recursion is the original intent.** `worker.sh` already sets `LINTER_WORKING_DIRECTORY="{}"` for `GO_MODULES` to run golangci-lint *from inside* the module, and the no-op guard just below is clearly meant to include it (like `ANSIBLE`/`RUST_CLIPPY`) so no path is appended — a bare `golangci-lint run` from the module dir, which is already recursive. But that guard checks `FILE_TYPE == "GO_MODULE"` (singular) while the type is `GO_MODULES` (plural), so it never fires, GNU Parallel appends the bare module dir, and `golangci-lint run <dir>` collapses to the **root package only**. This PR's `{}/...` restores the intended recursive behaviour explicitly (and more robustly than relying on the default target).

**On `VALIDATE_ALL_CODEBASE=false` — it doesn't widen the module set; it fixes a false green.** Module selection happens in `buildFileList.sh`: it writes `FILE_DIR_NAME` (the changed file's directory, or the module root for a `go.mod` change) into the GO_MODULES list. `{}/...` is appended to *that already-selected directory* in `worker.sh`, so it can only add nested sub-packages of it — never siblings, the parent, or another module. Which directories get linted is decided entirely before `worker.sh` and is unchanged by this PR.

Reproduced through the real `slim-v8.6.0` container (only `worker.sh` swapped for the patched one), incremental mode, a PR that changes **only `moduleB/go.mod`**, with a pre-existing `govet` bug in `moduleB/sub/`:

| | Result |
|---|---|
| **Stock** | exit **0** — "Successfully linted GO_MODULES" → false green; `moduleB/sub` never linted |
| **This PR** | exit **1** — flags `moduleB/sub/sub.go: printf … (govet)` |

Scope was identical in both runs (only `moduleB` linted; the unchanged `moduleA` was not). So a Renovate/dependency `go.mod` bump — the classic `VALIDATE_ALL_CODEBASE=false` case — currently passes even when the module it touches has real issues, and this PR is what makes it lint the module that changed. (Under `VALIDATE_ALL_CODEBASE=true` each package dir is added separately, which is why full-codebase runs mostly escape it — this is specifically the incremental path.)

**`/...` respects module boundaries.** A parent module with a nested module (its own `go.mod`) under it: `golangci-lint run <parent>/...` does not descend into the nested module (Go's `...` stops at nested `go.mod`), so there's no double-linting or cross-module leak — the nested module is linted on its own entry when its files change.

Happy to attach the self-contained `repro.sh` (builds the fixture, patches `worker.sh`, runs stock vs patched), and to extend the `go_modules_good` fixture with a second package outside `cmd/` so the test proves multi-directory recursion. And glad to open a tracking issue if you'd prefer that as the anchor.

jabenedicic:
@ferrarimarco Any chance of getting a re-assessment and re-review of this with the expanded information?

ferrarimarco:
Hey folks. It's on my radar, I'll have a look :)

IldarMinaev:
I opened #8044 to track this as requested. It includes a standalone reproducer and the public CI failure from `Netcracker/qubership-monitoring-operator`.

This also matches `golangci/golangci-lint#6092`: passing a directory that contains `go.mod` but no Go files is expected to fail, and golangci-lint should be run from inside the module. Super-Linter already sets `LINTER_WORKING_DIRECTORY` to the selected module, but then appends the same directory as a bare positional target. This PR fixes that mismatch by using a recursive package pattern. The per-module workaround discussed in `golangci/golangci-lint#828` uses the same `<module>/...` form.

ferrarimarco:
Thanks!

ferrarimarco:
Sorry for the delay in reviewing this. I'll get to it asap

test/linters/go_modules/go_modules_good_root/golang_good_01.go
ferrarimarco, line 1: Can you please refactor this change to have the following tree:

```
test/linters/go_modules
├── go_modules_bad
│   ├── golang_bad_01.go
│   ├── go.mod
│   └── go.sum
├── go_modules_good_root
│   ├── golang_good_01.go
│   ├── go.mod
│   └── go.sum
└── go_modules_good_subdir
    ├── cmd
    │   └── golang_good_01.go
    ├── go.mod
    └── go.sum

5 directories, 9 files
```

You can:

1. Leave the `go_modules_bad` directory as is.
2. Rename `go_modules_good` to `go_modules_good_root`
3. Create `go_modules_good_subdir` so that it has the `.go` file in the `cmd` subdir.

This way, we test both `.go` files in at the root of the module, and `.go` files only in subdirs.

Thanks!

ferrarimarco (changes requested):
Thanks for this contribution.

I was able to validate the fix, and I just have a minor review comment so that we don't leave test cases out.

vlsi:
Done — I refactored the fixtures into the exact tree you proposed:

```
test/linters/go_modules
├── go_modules_bad          # unchanged
├── go_modules_good_root    # .go file in the module root
│   ├── golang_good_01.go
│   ├── go.mod
│   └── go.sum
└── go_modules_good_subdir  # .go file only under cmd/
    ├── cmd
    │   └── golang_good_01.go
    ├── go.mod
    └── go.sum
```

`go_modules_good` became `go_modules_good_root` as a pure rename (no content change), and `go_modules_good_subdir` is the new one, so we now cover both a module with `.go` files at its root and a module whose sources live only in sub-directories. The latter reproduces the failing case: `golangci-lint run <module-root>` reports `no go files to analyze`, while `<module-root>/...` scans the `cmd/` package. Thanks for the review.

ferrarimarco (approved):

PR #8025, opened 2026-07-25
feat: add azure bicep linter

### LINKED ISSUE #1610: Add Bicep as a supported linter type
The Bicep v0.4 release introduced linting support. It would be great if the github / super-linter could be updated to include .bicep files as a supported linter type.

stektron:
<h2>Summary</h2>

-  Adds a linter for Azure Bicep
- Close https://github.com/super-linter/super-linter/issues/1610

## Readiness checklist

### Pull request author tasks

- [x] I followed the
- [x] If this pull request is about and existing issue, I added the
      the pull request.

      released version.
      `infrastructure`.


stektron:
I just realized that I likely didn't need to add an entry lib/functions/detectFiles.sh since bicep has unique file extension.  Let me know if that should be stripped out to keep things tidy.

ferrarimarco:
Thanks for this PR, I'll review and get back to you 🙂

TEMPLATES/bicepconfig.json
```diff
@@ -0,0 +1,128 @@
+{
```
ferrarimarco, line 2: Can we please keep a configuration that's as close as possible to the default? If the tool accepts an empty config file, that would be great so we don't have to maintain our own config.

Makefile
```diff
@@ -351,6 +351,7 @@ lint-subset-files-enable-expensive-io-checks: ## Lint a small subset of files in
 		-e FILTER_REGEX_EXCLUDE=".*(/test/linters/|CHANGELOG.md|/test/data/test-repository-contents/).*" \
 		-e VALIDATE_ALL_CODEBASE=true \
 		-e VALIDATE_ARM=true \
+		-e VALIDATE_BICEP=true \
```
ferrarimarco: Can you please revert this change? This test case is not focused on Bicep (or Arm) specifically.

lib/functions/detectFiles.sh
```diff
@@ -91,6 +91,28 @@ DetectARMFile() {
   fi
 }
 
+DetectBicepFile() {
```
ferrarimarco: Yeah, as you mentioned, we don't need this function because we just check for the extension.

lib/functions/detectFiles.sh
```diff
@@ -212,6 +234,7 @@ export -f DetectDependabot
 export -f DetectGitHubActions
 export -f DetectARMFile
 export -f DetectAWSStatesFIle
+export -f DetectBicepFile
```
ferrarimarco: And we can revert this as well

scripts/install-bicep-cli.sh
```diff
@@ -0,0 +1,46 @@
...
+set -euo pipefail
+
+# Install Bicep CLI
+# Requires curl and sudo
+# Reference https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter
+
+INSTALL_DIR="/usr/local/bin"
+EXECUTABLE_PATH="${INSTALL_DIR}/bicep"
+
+info() {
```
ferrarimarco: No need to redefine these logging functions here. Just plain echoes are fine, for simplicity

lib/globals/linterRules.sh
```diff
@@ -13,6 +13,8 @@ ANSIBLE_FILE_NAME="${ANSIBLE_CONFIG_FILE:-.ansible-lint.yml}"
 ARM_FILE_NAME=".arm-ttk.psd1"
 BASH_FILE_NAME="${BASH_FILE_NAME:-".shellcheckrc"}"
 BASH_SEVERITY="${BASH_SEVERITY:-""}"
+# shellcheck disable=SC2034  # Variable is referenced indirectly
+BICEP_FILE_NAME=".bicep.json"
```
ferrarimarco: Can you add the variable here and in the readme to allow for customizing the config file name?

lib/functions/linterCommands.sh
```diff
@@ -108,6 +108,7 @@ if [ "${BASH_EXEC_IGNORE_LIBRARIES:-}" == 'true' ]; then
   debug "Enabling bash-exec option to ignore shell library files."
   LINTER_COMMANDS_ARRAY_BASH_EXEC+=('true')
 fi
+LINTER_COMMANDS_ARRAY_BICEP=(bicep lint)
```
ferrarimarco, line 111: How does Bicep load the configuration file?

README.md
```diff
@@ -71,6 +71,7 @@ Super-linter supports the following tools:
...
 | **AWS CloudFormation templates**      | [AWS CloudFormation Linter (cfn-lint)](https://github.com/aws-cloudformation/cfn-lint), [Checkov](https://www.checkov.io/), [Trivy](https://trivy.dev/)                                                                   | See YAML formatters                                                                                                |
 | **Azure Resource Manager (ARM)**      | [Azure Resource Manager Template Toolkit (arm-ttk)](https://github.com/azure/arm-ttk), [Checkov](https://www.checkov.io/), [Trivy](https://trivy.dev/)                                                                    | See JSON formatters                                                                                                |
+| **Azure Bicep**                       | [bicep lint](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter)                                                                                                                                 | [bicep format](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-cli?tabs=bicep-cli/)     |
```
ferrarimarco: This line implies that we're running bicep format as well, but the PR only implements running bicep lint. I'd suggest implementing support for running the formatter too.

ferrarimarco (changes requested):
Thanks for the PR!

README.md
stektron: This should be doable.  The same bicep cli that does linting can also format. I'll need to do more testing though
https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-cli?tabs=bicep-cli#format

lib/functions/linterCommands.sh
stektron: The Bicep configuration file must be in the same directory or a parent directory of your Bicep files, otherwise it falls back to the default linter rules.
https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-config#understand-the-file-resolution-process

So I'm a bit unsure what the best path forward is.  Should we just have a note in the main readme.md stating the above?

stektron: Working to add the ability to specify a custom config file location here:
https://github.com/Azure/bicep/pull/20149

Will convert this to draft and update when I get some clarity from the azure bicep maintainers.

ferrarimarco: Thanks! Also, there are workaround we can use, especially if the bicep linter and formatter automatically discovers config files by traversing directories.

PR #8094, opened 2026-09-01
fix(codespell): skip root package lockfiles

### LINKED ISSUE #8082: codespell should not check `package-lock.json` ?

- [x] I have searched the existing issues

### Are you using the latest Super-linter version available?

- [x] I am using the latest Super-linter version.
- [x] I can reproduce the issue running Super-linter using complete version identifier (example: vX.Y.Z), and not just with a partial one (example: vX)

- [x] I am using the super-linter/super-linter action or container image, and not the deprecated github/super-linter action or container image.

### Are you reasonably sure that it's a Super-linter issue, and not an issue related to a tool that Super-linter runs?

- [x] I think that this is a Super-linter issue.

### Current Behavior

SPELL_CODESPELL is checking `package-lock.json`. See screenshot:

Image

`package-lock.json` is a generated file which the user has no control over. I don't think it should ever report issues. I thought this issue might have been out of scope for Super Linter and the user may need to block this themselves, but I researched a bit and saw the PR where it was merged does appear to have a default config which skips lock file?
https://github.com/super-linter/super-linter/pull/7357/changes#diff-becf3e6526b23a645c199933acfb8c13acc58269e87aa607da7f0417ce6e49c3R6

(I have my linter configs set to be in project root and don't have a `.codespellrc‎` though. could that be the reason?

### Expected Behavior

`package-lock.json` is a generated file which the user has no control over. I don't think it should ever report issues.

### Super-Linter version

```markdown
8.7.0
```

### Super-linter configuration

```yaml
---
name: Super Linter

on:
  pull_request:
    branches-ignore:
      - "dependabot/**"

permissions:
  contents: read

jobs:
  build:
    permissions:
      contents: read
      issues: write
      packages: read
      pull-requests: write
      statuses: write

    name: Lint Code Base
    runs-on: ubuntu-latest

    steps:
      - name: Checkout Code
        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
        with:
          fetch-depth: 0
          persist-credentials: false

      - name: Install dependencies
        run: npm ci

      - name: Lint Code Base
        uses: super-linter/super-linter/slim@4ce20838b8ab83717e78138c5b3a1407148e0918 # v8.7.0
        env:
          CSS_FILE_NAME: stylelint.config.mjs
          DEFAULT_BRANCH: main
          FILTER_REGEX_EXCLUDE: "/docs/"
          IGNORE_GITIGNORED_FILES: true
          LINTER_RULES_PATH: /
          LOG_LEVEL: NOTICE
          MARKDOWN_CONFIG_FILE: .markdownlint.json
          SUPPRESS_POSSUM: true
          VALIDATE_ALL_CODEBASE: false
          VALIDATE_BIOME_FORMAT: false
          VALIDATE_BIOME_LINT: false
          VALIDATE_CHECKOV: false
          VALIDATE_EDITORCONFIG: false
          VALIDATE_GIT_COMMITLINT: false
          VALIDATE_GITLEAKS: false
          VALIDATE_JSCPD: false
          VALIDATE_MARKDOWN_PRETTIER: false
          VALIDATE_NATURAL_LANGUAGE: false
          VALIDATE_XML: false
```

### Relevant log output

```shell
explained above
```

### Steps To Reproduce

explained above 

### Anything else?

_No response_

RRXXZZYY:
Fixes #8082

## What

- Add an exact `package-lock.json` skip pattern while retaining the existing nested-path pattern in the default and repository codespell configurations.
- Add a regression in `test/lib/linterRulesTest.sh` that exercises both root and nested lockfiles.
- Register the regression's temporary directory with the repository cleanup trap, as requested in review.

## Why

When Super-linter passes a root lockfile to codespell as `package-lock.json`, the existing `*/package-lock.json` pattern does not match because the argument has no directory separator. Nested lockfiles are already covered, so both forms are needed.

## Verification

Fresh on the branch rebased onto current `main` (`2066911`):

- repository-pinned codespell 2.4.3: root and nested lockfiles exit 0; a non-lockfile control containing the same misspelling exits 65
- `bash -n test/lib/linterRulesTest.sh`: passed
- shfmt 3.13.1 on `test/lib/linterRulesTest.sh`: passed
- ShellCheck 0.11.0 on `test/lib/linterRulesTest.sh`: passed
- `git diff --check`: passed

## Validation boundary

Docker is not available in the local Windows environment. The complete `linterRulesTest.sh`/`make test` path depends on the project container and `/action/lib/.automation`, so the full container suite remains for PR CI. I have not marked that suite as locally passed.

## Readiness checklist

- [ ] I checked that all workflows return a success. (Waiting for PR CI.)
- [x] No user-facing documentation change is needed for this default-configuration correction.
- [x] I provided the necessary regression test.
- [x] I squashed the change into a single commit.
- [x] I followed the Conventional Commits v1.0.0 specification.
- [x] No upgrade-guide entry is needed because this is a backwards-compatible bug fix.
- [x] The pull request links and closes the existing issue.

## AI-assisted disclosure

I used an AI coding assistant to help inspect the configuration path, draft the focused patch and test, and review the final diff. I reproduced the bug with the pinned codespell version, ran every check listed above, addressed the maintainer's review comments, and reviewed the submitted changes and public text before updating this pull request.

### DIFF
--- a/.github/linters/.codespellrc
+++ b/.github/linters/.codespellrc
@@ -3,4 +3,4 @@
 ; https://github.com/codespell-project/codespell/issues/3854 is resolved
 ; check-filenames =
 check-hidden =
-skip = */package-lock.json,*.log,super-linter-output,.git
+skip = package-lock.json,*/package-lock.json,*.log,super-linter-output,.git
--- a/TEMPLATES/.codespellrc
+++ b/TEMPLATES/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
 check-hidden =
-skip = */package-lock.json,*/go.mod,*/go.sum,*.log,.git
+skip = package-lock.json,*/package-lock.json,*/go.mod,*/go.sum,*.log,.git
--- a/test/lib/linterRulesTest.sh
+++ b/test/lib/linterRulesTest.sh
@@ -139,6 +139,30 @@ function LinterRulesVariablesExportTest() {
   notice "${FUNCTION_NAME} PASS"
 }
 
+function CodespellDefaultConfigurationSkipsPackageLockFilesTest() {
+  local FUNCTION_NAME
+  FUNCTION_NAME="${FUNCNAME[0]}"
+  info "${FUNCTION_NAME} start"
+
+  local TEST_DIRECTORY
+  TEST_DIRECTORY="$(mktemp -d)"
+  initialize_temp_directory_cleanup_traps "${TEST_DIRECTORY}"
+  mkdir -p "${TEST_DIRECTORY}/nested"
+  printf '{"description":"t%sh"}\n' "e" >"${TEST_DIRECTORY}/package-lock.json"
+  printf '{"description":"t%sh"}\n' "e" >"${TEST_DIRECTORY}/nested/package-lock.json"
+
+  local CODESPELL_DEFAULT_CONFIGURATION
+  CODESPELL_DEFAULT_CONFIGURATION="$(realpath "TEMPLATES/.codespellrc")"
+  if ! (
+    cd "${TEST_DIRECTORY}"
+    codespell --config "${CODESPELL_DEFAULT_CONFIGURATION}" package-lock.json nested/package-lock.json
+  ); then
+    fatal "The default codespell configuration should skip package-lock.json files at the workspace root and in nested directories"
+  fi
+
+  notice "${FUNCTION_NAME} PASS"
+}
+
 LoadDefaultRulesTest() {
   local FUNCTION_NAME
   FUNCTION_NAME="${FUNCNAME[0]}"
@@ -160,4 +184,5 @@ GetLinterRulesTest
 GetLinterRulesEmptyDotRulesPathTest
 GetLinterRulesEmptyRootRulesPathTest
 LinterRulesVariablesExportTest
+CodespellDefaultConfigurationSkipsPackageLockFilesTest
 LoadDefaultRulesTest


test/lib/linterCommandsTest.sh
```diff
@@ -86,6 +86,29 @@ function LinterCommandPresenceTest() {
   notice "${FUNCTION_NAME} PASS"
 }
 
+function CodespellDefaultConfigurationSkipsPackageLockFilesTest() {
+  local FUNCTION_NAME
+  FUNCTION_NAME="${FUNCNAME[0]}"
+  info "${FUNCTION_NAME} start"
+
+  local TEST_DIRECTORY
+  TEST_DIRECTORY="$(mktemp -d)"
```
ferrarimarco: After creating the temp dir, call the function to initialize cleanup traps:

```
initialize_temp_directory_cleanup_traps "${TEST_DIRECTORY}"
```

test/lib/linterCommandsTest.sh
```diff
@@ -86,6 +86,29 @@ function LinterCommandPresenceTest() {
   notice "${FUNCTION_NAME} PASS"
 }
 
+function CodespellDefaultConfigurationSkipsPackageLockFilesTest() {
```
ferrarimarco: (nit)

I would move this test to `test/lib/linterRulesTest.sh`, because we're actually testing a rules file. WDYT?

ferrarimarco (changes requested):
Hi @RRXXZZYY !

Thanks for this PR. Just two small things to address.

test/lib/linterCommandsTest.sh
RRXXZZYY: Addressed in 60e365f. The temp directory now registers `initialize_temp_directory_cleanup_traps` immediately after `mktemp -d`.

test/lib/linterCommandsTest.sh
RRXXZZYY: Agreed. I moved the regression to `test/lib/linterRulesTest.sh` in 60e365f so it lives with the rules-file tests.

RRXXZZYY:
Addressed both review items in 60e365f.

Validation on the updated files:
- focused Codespell regression passed with the repository-pinned `codespell==2.4.3`
- `bash -n` passed
- `shfmt v3.13.1 -d` passed
- ShellCheck v0.11.0 passed
- `git diff --check` passed

I could not complete the entire `linterRulesTest.sh` outside the project container because the local environment does not provide `/action/lib/.automation`; the focused regression runs before that environment-only failure and passed. The existing npm/pip/Trivy audit failures are on unchanged dependency lockfiles and are also affecting concurrent main merge-queue runs, so I left those unrelated dependency updates out of this focused PR.

ferrarimarco:
Thanks! I'll have another look :)

ferrarimarco (approved):

